PR #2, opened 2022-06-04
Add Using Prisma Option

iDarkLightning:
Adds an option to the CLI to use Prisma in the created template.

The created example uses SQLite for the database and shows an example query and mutation using the Prisma client.


nexxeln:
I'm so embarrassed by that typo in the readme lol. Thank you for the PR.

iDarkLightning:
haha no problem :)

PR #6, opened 2022-06-15
TypeScript correction happens after choice now

barrybtw:
-- TypeScript correction now happens right after choosing JavaScript instead of after all the prompts.


nexxeln:
Looks good thank you for this!

Issue #11, opened 2022-06-17
Add default project name


anthonylaflamme:
Enhancement

Add a default project name on prompt so I can just enter and proceed to next step

anthonylaflamme:
This can also resolve empty project name prompt that resolve in an issue : 
<img width="454" alt="Screen Shot 2022-06-17 at 3 48 55 PM" src="https://user-images.githubusercontent.com/4626172/174392238-53c43813-1856-46d5-b067-b117ea7cba6d.png">

mallsjr:
I believe I have a fix for this. What would be a good default project name?

nexxeln:
> I believe I have a fix for this. What would be a good default project name?

t3-app would be a nice default name imo.

nexxeln:
implemented in #19

PR #15, opened 2022-06-18
implemented better type safety for adding future prompts

juliusmarminge:
Put all the prompts into an array with some type safety so that future prompts can be added and get some warnings.

The packages are now enables as a multi-select so adding more packages is easier and doesn't require a new prompt to be added.

Also put in some response messages such as when leaving name blank (Closing #11 in the process). See a demo [here (CleanShot link)](https://cleanshot-cloud-fra.accelerator.net/media/38032/Ze78t4RhWyHtcN6249n1JfUB211oD9zaQotGhnWQ.mp4).

**There is a limitation however. If selecting Next-Auth, but not Prisma, it will go ahead and create a project with both of them due to there only being one Next-Auth template available. I can look into perhaps making it so that every individual package is installed separately, and configs copied over accordingly, so that there won't have to be a template-folder for each combination.**

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -66,7 +66,7 @@ const promts: PromptObject[] = [
 (async () => {
   console.log(chalk.red("Welcome to the create-t3-app !"));
 
-  const { name, packages } = await prompts(promts);
+  const { name, packages } = await prompts(promts) as { name: string, packages: string[]};
 
   // TODO: It should probably be createProject's responsiblity to interpret the `packages` array
   const usingPrisma = packages.some((p: string) => p === "prisma");
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
-import chalk from "chalk";
 import prompts, { type PromptObject } from "prompts";
+import { logger } from "./helpers/logger";
 
 import createProject from "./helpers/create";
 
@@ -14,7 +14,7 @@ const promts: PromptObject[] = [
     message: "What will your project be called?",
     format: (name: string) => {
       if (name === "") {
-        console.log(chalk.yellow(`Using default name: ${DEFAULT_PROJECT_NAME}`));
+        logger.warn(`Using default name: ${DEFAULT_PROJECT_NAME}`);
         return DEFAULT_PROJECT_NAME;
       }
       return name.trim();
@@ -37,9 +37,9 @@ const promts: PromptObject[] = [
     ],
     format: (language: string) => {
       if (language === "javascript") {
-        console.log(chalk.red("Wrong answer, using TypeScript instead..."));
+        logger.error("Wrong answer, using TypeScript instead...");
       } else {
-        console.log(chalk.green("Good choice! Using TypeScript!"));
+        logger.success("Good choice! Using TypeScript!");
       }
       return;
     }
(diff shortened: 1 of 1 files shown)


juliusmarminge:
My bad, the limitation is that Prisma has to be selected in order for NextAuth to be installed. Selecting NextAuth but not Prisma will cause the base `template` to be pulled.

nexxeln:
Thank you for migrating all the console.logs to the logger, that was long overdue.

Yes for next-auth, prisma is required, I think the existing select prompt was fine. But if more people like this multi select prompt, we can have an option for `prisma` and `next-auth + prisma`, which will work with the current templates.

What do you think?

juliusmarminge:
Sure, that would work. However, there would still be the issue with expansion when/if adding more packages from the init.tips stack (I guess there is really only Zustand left right now...)

Theo mentioned an enhancements with passing a whole configuration and installing each package separately ([on stream](https://youtu.be/-v42G_RbKvw?t=7056)) which would require each package to be individually selected and that was my plan to tackle next if that's something you'd like (it's your project after all)

nexxeln:
Yeah I would love that. What do you have in mind for making that work?

juliusmarminge:
I have never done anything like it before so perhaps its more difficult than I imagine, but my idea was just having a base Next.js app as the template-folder. Then have separate folders like addons/tailwind, addons/trpc, addons/prisma, addons/next-auth and have installation scripts for all of them that puts those files where they should.

The diffucult part I can see is how to change the _app.tsx accordingly. I can take a first look at it tonight and see if its possible.

Hämta Outlook för iOS<https://aka.ms/o0ukef>
________________________________
Från: Shoubhit Dash ***@***.***>
Skickat: Sunday, June 19, 2022 5:31:33 AM
Till: nexxeln/create-t3-app ***@***.***>
Kopia: Julius Marminge ***@***.***>; Author ***@***.***>
Ämne: Re: [nexxeln/create-t3-app] implemented better type safety for adding future prompts (PR #15)


Yeah I would love that. What do you have in mind for making that work?

nexxeln:
Yeah I was looking into this too, we could use something like [jscodeshift](https://github.com/facebook/jscodeshift), BlitzJS uses this for their [*recipes*](https://github.com/blitz-js/blitz/tree/canary/recipes), but I'm not sure how that will work here.

juliusmarminge:
Cool, I have some time available tonight and early next week to look into this. What do you want me to do about this pr? Go back to the individual selects instead of the multi one?

Hämta Outlook för iOS<https://aka.ms/o0ukef>
________________________________
Från: Shoubhit Dash ***@***.***>
Skickat: Sunday, June 19, 2022 10:23:25 AM
Till: nexxeln/create-t3-app ***@***.***>
Kopia: Julius Marminge ***@***.***>; Author ***@***.***>
Ämne: Re: [nexxeln/create-t3-app] implemented better type safety for adding future prompts (PR #15)


Yeah I was looking into this too, we could use something like jscodeshift<https://github.com/facebook/jscodeshift>, BlitzJS uses this for their recipes<https://github.com/blitz-js/blitz/tree/canary/recipes>, but I'm not sure how that will work here.

nexxeln:
Yeah I'd like that, after that I'll merge it. After that we can talk about how to use jscodeshift on discord maybe, and work together on it.

nexxeln:
This looks good thank you so much!

PR #16, opened 2022-06-18
Issue #11 add default project name

mallsjr:
Added check for empty project name if so use a default project name of t3-app


nexxeln:
Thank you! Looks good, but I think if I merge this now it will clash with #15 . So I'll close this, it's already fixed in that PR. Thanks though!

PR #14, opened 2022-06-18
Add scaffolding directory to log output

BenDundon:
closes #13 

Perhaps warrants more implementation of scaffolding into directories other than CWD, but this PR will fulfill the enhancement request in the issue.

PR #17, opened 2022-06-19
Added the ability to scaffold outside CWD

BenDundon:
Mentioned in #14 that this might be a useful feature. If you disagree or don't like it as part of the workflow, feel free to close this PR, just thought I'd follow-up.

### DIFF
--- a/src/helpers/create.ts
+++ b/src/helpers/create.ts
@@ -10,6 +10,7 @@ const execa = promisify(exec);
 
 const createProject = async (
   projectName: string,
+  projectDir: string,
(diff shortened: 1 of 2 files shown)


juliusmarminge:
It is already possible to scaffold outside cwd by providing the path as project-name.

for example, entering `apps/t3-app` would create `t3-app` inside of `apps`.
CleanShot 2022-06-19 at 22 57 04

BenDundon:
🤦🏻
How'd I miss that? Ok, I'll close this PR then.

PR #19, opened 2022-06-20
Individual installs

juliusmarminge:
Would love to get some discussions going for this approach before maybe discovering codemods etc...

I have restructured the template layout. Now, there is a base Next.js application under [template/base](https://github.com/juliusmarminge/create-t3-app/tree/individual-installs/template/base). This contains nothing extra.

Then there are addons for `tailwind`, `trpc`, `prisma` and `next-auth` under [template/addons/*](https://github.com/juliusmarminge/create-t3-app/tree/individual-installs/template/addons). These contains necessary boilerplate for that package.

What is not so good is when two packages need to modify the same file. This part probably needs some more work and thought to get right and be more accessible for contributors. It's hacked together right now.

The packages and its boilerplate are installed from the installers under [src/installers/*](https://github.com/juliusmarminge/create-t3-app/tree/individual-installs/src/installers). I also refactored the [src/helpers/create](https://github.com/juliusmarminge/create-t3-app/blob/individual-installs/src/helpers/create.ts) since it was getting a bit long for a single function, feel free to modify this to your liking.

What do you think about the overall approach? All packages can be installed individually so if you are just building a backend service like the [joltik polls app was on stream](https://github.com/TheoBr/joltik) you dont have to include tailwind and trpc for example.


juliusmarminge:
Including a screenshot of the CLI with these steps: 
CleanShot 2022-06-20 at 19 03 54

I have tested creating some example apps that you can browse [here](https://github.com/juliusmarminge/t3-starters)

nexxeln:
Everything looks good to me! Thank you so much!

PR #28, opened 2022-06-22
Add inferQueryResponse helper for tRPC

vitorwlima:
Added a helper to infer types from queries in tRPC. It is a really important one that always needs to be googled because the syntax is really difficult. I believe it is really helpful to have this in the template.

Example:
Definition: https://github.com/TheoBr/roundest-mon/blob/e2834b689941df2c61b5acb9816f675437c92df9/src/pages/api/trpc/%5Btrpc%5D.ts#L11
Implementation: https://github.com/TheoBr/roundest-mon/blob/827d89463e1fc6edb7bbb5eb814fde67f6c36381/src/pages/index.tsx#L88

### DIFF
--- a/template/addons/trpc/api-handler.ts
+++ b/template/addons/trpc/api-handler.ts
@@ -1,6 +1,7 @@
 // src/pages/api/trpc/[trpc].ts
+import { inferProcedureOutput } from "@trpc/server";
 import { createNextApiHandler } from "@trpc/server/adapters/next";
-import { appRouter } from "../../../server/router";
+import { AppRouter, appRouter } from "../../../server/router";
 import { createContext } from "../../../server/router/context";
 
 // export API handler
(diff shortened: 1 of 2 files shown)


vitorwlima:
image

nexxeln:
This looks cool! I think it's definitely less used though. I personally have never used it, including it by default feels wrong to me because someone might never use it and its just unused code at the end. Does that make sense?

Would love to hear other people's thoughts on this as well.

c-ehrlich:
I have used this before, saw it in Theo's video and found it useful immediately. In a perfect world this would be something that gets added to tRPC in some way, instead of having to put it in the starter where a keeping the codebase small is desireable.

For the time being while this is not the case, I'm not sure if it's better to include or not.

juliusmarminge:
Maybe just leaving a commented link to where you can find them?
https://trpc.io/docs/infer-types#inference-helpers

p.s. I think this is a util that I usually put in my utils/trpc.ts

nexxeln:
> Maybe just leaving a commented link to where you can find them? https://trpc.io/docs/infer-types#inference-helpers
> 
> p.s. I think this is a util that I usually put in my utils/trpc.ts CleanShot 2022-06-22 at 10 21 25

I like this. @vitorwlima I think this is the way to go.

vitorwlima:
I agree, this way it helps but it never leaves unused code. I'll make the change to the PR.

PR #29, opened 2022-06-22
feat: check if dir is empty

asheeeshh:
screenshots were added in previous pr so not adding again.


nexxeln:
Have you tested it?

asheeeshh:
yes, the same as earlier

nexxeln:
Cool thank you again!

Issue #22, opened 2022-06-21
Erroneously selects yarn just because I have it installed


TheoBr:
I have Yarn installed for mostly react native stuff, for some reason create-t3-app always picks it. Would love to be able to pick between yarn, npm and pnpm, but would be cool with it auto-selecting which I "ran it with" as well

nexxeln:
Yeah looks like https://github.com/nexxeln/create-t3-app/blob/main/src/helpers/get-pkg-manager.ts is very buggy, I'll rewrite this.

nexxeln:
Fixed in #30

PR #30, opened 2022-06-22
fix pacakage manager logic and add prettier option

asheeeshh:
~~added a new option to use prettier with the project with default config -~~

fixed the bug in package manager logic which was causing it to pick random package manager. temporary fix for #22


nexxeln:
please remove prettier :pray:

asheeeshh:
> please remove prettier 🙏

i though we discussed about it 🤔

nexxeln:
i thought so too 🤔

nexxeln:
thanks!

Issue #39, opened 2022-06-23
Adding project tooling to standardize contributions


MWhite-22:
Tools:

- Changesets 
  - Managing version numbers and changelogs. If the package is published on NPM, I think this is a good idea to include 

- Prettier
  - Formatting for all contributions to avoid 1000 line "trailing comma" git history
  
- Eslint
  - Code linting to conform to a general standard. This is going to be very opinionated so we can go back and forth on this one for a bit in PR review

- Husky
  - Trigger scripts on git actions

- Lint-Staged
  - Rather than linting the entire project on each commit, only lint the staged folders

I think adding the above is a good step to "open-sourcing" the repo. We can also start to get into some CI/CD actions with each PR triggering a lint action before going to review.

MWhite-22:
Should have a PR ready shortly

MWhite-22:
Merged in #40

Issue #25, opened 2022-06-21
Prettier?


TheoBr:
🤔

nexxeln:
Yeah we could have a config but I think that may be too opinionated? I personally just use the default that comes with the vscode extension. If someone wants prettier they can install/configure it themselves imo. What do you think?

booleans-oss:
Make it optional ? When generating a project, it prompts whether to add a prettier config file and packages or not.

TheoBr:
Thought a bit about this and I don't think it meaningfully differs from my stance on "add Jotai/Zustand"

You can just NPM install if you want prettier. No need for us to include. I think I let my desire for prettier get ahead of me 😂

nexxeln:
I think that most people have prettier installed in their editor and if they have a preferred config then can do that on their own. Closing this now after discussing on discord.

PR #40, opened 2022-06-23
Adding in dev tooling to the contribution process

MWhite-22:
PR for issue #39

Happy to walk through individual commits with more detail


package.json
```diff
@@ -2,33 +2,30 @@
   "name": "create-t3-app",
   "version": "4.0.5",
   "description": "Create web application with the t3 stack",
+  "author": "Shoubhit Dash <youremail@here.com>",
```
MWhite-22, line 5: @nexxeln Be sure to add your email in here as I believe GitHub actually uses the information

staaph: shoubhit2005@gmail.com

.gitignore
```diff
@@ -1,2 +1,46 @@
...
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# MISC
+.idea
+.turbo/
+.cache/
+.next/
+.nuxt/
```
staaph, line 38: remove nuxt here to keep it clean

MWhite-22: Its just a generic .gitignore that covers 90% of use cases. If it doesn't hurt anything I would rather keep it complete and easily extensible to other projects.

package.json
MWhite-22: Done

staaph:
@nexxeln 
move prettier to editorconfig ?

MWhite-22:
> @nexxeln move prettier to editorconfig ? https://editorconfig.org/

VSCode needs a plugin to work and I dont think this is a valid replacement for prettier (which works with eslint as well to prevent overriding rules re: styling)

asheeeshh:
1. prettier -- it has been decided earlier that we don't need prettier either in the repository itself or in the generated project.
2. the same goes for eslint ig. @nexxeln your thoughts?
3. i personally don't know if we really need husky, let's see what others think about it

MWhite-22:
1) The problem with not having a repo level prettier is that, as people contribute to the project, we will inevitably get a bloated git history of double quotes turning into single quotes, and vice versa, as well as other minor stylistic changes. If this project is expected to live on and grow as new technologies are integrated, setting a standard now, even if that standard changes, is a good idea IMO.

MWhite-22:
> 1. prettier -- it has been decided earlier that we don't need prettier either in the repository itself or in the generated project. see #25 and #30

Both of these issues are about having prettier in the template built by create-t3, not having prettier IN create-t3. This PR is specifically for this repo, not adding prettier or eslint into any created templates

nexxeln:
I would've liked it if we had discussed all the tools before you made a PR 😄 . 

### Prettier Config
Please change single quotes to *false*

### Eslint
This looks good to me

### Changeset
I've never used or heard of this package, what exactly will it do?

### Husky
I'm kind of 50/50 on this one. I would love to know what other people think.

MWhite-22:
AFK for a few hours. Will push an update in a bit.

nexxeln:
> Pushed new prettier config with double quotes.
Thank you, please format the files also with the new config.

Changesets seems cool so why not, but the first time I'll use it I'll poke you with questions on Discord 😄.

nexxeln:
> Already re-formatted!

Oh lol sorry I didn't look at the commit.

I'm still unsure about husky.

MWhite-22:
Totally happy to remove it if we need to. The current implementation just runs a lint and format before commit. 

Without it, it's up to people to lint and format on their own, and although most IDEs have built in settings to do so on save, not everyone is going have the same setups and extensions. 

This way we keep a relatively sane integrity check in the most "gentle" way possible.

juliusmarminge:
I think it adds value as the amount of contributors grow.

Looking at commits and seeing a bunch of prettier-changes is annoying and makes it harder to see what the commit is actually doing.

What tool/setup to use for this I have no opinion on.

nexxeln:
Okay let's do it then! I'll merge

PR #44, opened 2022-06-25
Fix a small linting issue and cleanup extraneous comments

MWhite-22:
As per the title, added 'type-fest' as a dev dependency to provide types for package.json (which was previously throwing a linting error).

Removed extra comments from my previous PR


nexxeln:
Thanks!

PR #46, opened 2022-06-27
remove windows specific commands 

asheeeshh:
changes made --
- removed windows specific cmds from package.json as we are using tsup.
- update start command to auto build.
- updated readme.


nexxeln:
Thank you!

PR #51, opened 2022-06-27
feat: add path alias

EvonuX:
This PR adds path aliases for better readability when importing modules.

To import a module it needs to be in the `src` directory.
Then, to import a component use the syntax below:
`import Button from '~/components/Button'`

I've replaced the current imports to use the alias.

Next.js docs: https://nextjs.org/docs/advanced-features/module-path-aliases


Dhravya:
Didn't know this can be done! Thanks 

From someone who doesn't know about the advanced features of NextJS, this can be confusing right?

asheeeshh (approved):

asheeeshh:
lol i was reviewing it

Issue #50, opened 2022-06-27
Take the name as an arg


Dhravya:
So I was creating a new project with create-t3-app and (like other create-x-apps), I used the project name in the main command itself
<img width="499" alt="image" src="https://user-images.githubusercontent.com/63950637/175955789-c07fefd5-da0d-46bb-95fd-2692c7a9675e.png">

I think, this should automatically set the project name, so it doesn't have to be asked again

PR #47, opened 2022-06-27
Added Spinners when downloading packages, initialising git and scaffolding 

Dhravya:
Used [ora](https://github.com/sindresorhus/ora) (Version 4) to create spinners

Here's the GIF of how it looks, let me know if any changes need to be done

GIF of spinners

Issue #42, opened 2022-06-24
Moving the "useX" questions into a select prompt, rather than individual questions


MWhite-22:
I think moving the "useTailwind", "useTrpc", etc., questions into a select block is a better UX for those that want to see all the options available before selecting.

juliusmarminge:
There is a multiselect example from a previous PR here: https://github.com/nexxeln/create-t3-app/commit/ead9ea84e36634eb1f2c3c5d5efd1cde5c8580e2

Issue #37, opened 2022-06-23
option to provide project name as argv


juliusmarminge:
Many CLI's allows you to include your name as a command-line-argument (argv), e.g. `npm create-turbo <project_name>`. It would be nice to support this too.

- If name is provided as argv, that should be used and the name-prompt be skipped. 
- If no name is provided, the prompt should be displayed.

nexxeln:
Yeah we could implement this, shouldn't be too hard.

MWhite-22:
We should think about adding in a --noGit and --noInstall flag (or cli options) as well. 

In a future when the t3 stack is used to spin up a new app inside of an already existing monorepo, having the app create its own git and node_modules would just require the end user to then go in and delete those folders before reinstalling node_modules at their monorepo root.

juliusmarminge:
I was actually doing this yesterday and thought that would be cool to have

MWhite-22:
Anyone currently tackling this? 

@asheeeshh You mentioned in the discord you were interested? LMK if not and i'll dive right in

nexxeln:
@asheeeshh has some technical issues with his keyboard. So feel free @MWhite-22!

MWhite-22:
This should be ready to go today. I had to do some pretty decent refactoring to support a --noInstall flag

nexxeln:
Great! You might have some merge conflicts tho because I just merged #47.

MWhite-22:
I'll work through that before pushing up to GH

MWhite-22:
This is completed in 4.1 and can be closed

PR #58, opened 2022-06-27
Adding in cli args, refactoring file structure, and changing prompts for inquirer

MWhite-22:
I went a little nuts, so, apologies, but i'll outline all the rationale below:

### Accepting args (#37 #50):
- Added commander as a parsing agent for the command line
- the first non-flagged argument will used as the app name and directory name
- Flags include:
  - `-y` or `--default`: Skips over the prompts and goes straight to initializing a new repo with opinionated defaults. Can still accept a directory argument so `npx create-t3-app myApp -y` will create a "default" app named "myApp" inside the "./myApp" directory
  - `--noGit`: Skip the git initialization step
  - `-v` or `--version`: Print the current create-t3-app version
  - `-h` or `--help`: All of this and more in a nicely formatted help menu

### Inquirer vs. Prompts:
- Inquirer has a much smaller package size and allows for assigning some generic types into the prompt instantiation. 
- Inquirer is slowly moving over to a full TS package where type safety will be massively improved
- Prompts was weirdly not exiting the CLI if the user pressed ctrl-c mid questions. Instead, it jumped straight into trying to run the scaffold scripts, which can't be cancelled once started. With inquirer, the user can press ctrl-c any time before the scaffolding starts, and the cli will correctly exit
- Inquirer comes with a check that the terminal is interactive. If not, it will throw a specific error. The current implementation catches that error, and will scaffold a default app for the user. This should help is someone not using an interactive terminal tries to use the cli and doesn't understand why no prompts are coming up.

### Add packages as a list, not individual questions (#42):
- Changed this question type to select from a list, rather than answer one by one
- Refactoring some of the installer types into an `as const` object means this list can be quickly and easily expanded in a type safe way

### A few bug fixes for ora (#47):
- Ora was blocking the scaffold function from asking the user if they wanted to overwrite a directory that already existed. Fixed those issues, and a few stylistic changes that bring the entire cli experience to a cohesive feel.

That should be everything. If you have any questions, please ask away.


juliusmarminge:
> On top of the above, I tried finding a way to easily add in a --noInstall flag to skip the actual install part of the scaffolding, but with the current architecture thats going to be really tough.

I guess a hack would be to just run a `rm -rf node_modules` afterwards. Can't believe there isn't an native argument like `npm install --add-only` for adding the latest version of a package to the package.json...

nexxeln:
I think `--noInstall` can be implemented later when the layouts RFC thing comes out. We will have kind of rewrite then anyway. Fo now I think not having this is fine.

nexxeln:
@MWhite-22 Can you also please unselect all of the prompts by default

nexxeln:
No, what I meant was that in the prompts to select the packages, initially nothing should be selected.

MWhite-22:
Right right, but I am building those choices programmatically from the "default" option.

So the new build will have all packages by `-y` default, but if they user is going through the prompts, the prompts will start off will all packages deselected. Correct?

MWhite-22:
Rebased to the new readme doc

nexxeln:
Looks good! Very good PR. Thank you so much

MWhite-22:
It does! Any name arg is passed in still, but the interactive part (aka: the fun language selector, and the packages selector) will not be viewable.

MWhite-22:
Never mind, I now see what you mean. Updating the string template to ${cliResults.appName}

PR #62, opened 2022-06-28
Fix Contributing Steps

JacobMGEvans:
If you follow current steps you will attempt to run start on a non-existing

- Fix is to run the build step before start


nexxeln:
Yes this was changed #58 and after merging that I slept and was going to do this now. Thank you.

PR #67, opened 2022-06-28
bug-fix: Resolve pkg root

### LINKED ISSUE #65: Packages fail to install
**Describe the bug**
After selecting packages to install it fails and shows this error
```
Aborting installation...
Error: ENOENT: no such file or directory, lstat 'template/base'
```

**To Reproduce**

1. `npx create-t3-app@latest`
2. Select any packages
3. See the error

**Expected behavior**
Packages should be installed

**Screenshots**
ss

juliusmarminge:
Closes #65 

Investigated lots of options using `process.mainModule`, `require.main` etc but these options are not compatible with ESM. The solution I [found](https://bobbyhadz.com/blog/javascript-dirname-is-not-defined-in-es-module-scope), which was also proposed by @devkevbot in [#65 (comment)](https://github.com/nexxeln/create-t3-app/issues/65#issuecomment-1168430972) was to use `import.meta.url` instead.

```ts
import path from "path";
import { fileURLToPath } from "url";

const __filename = fileURLToPath(import.meta.url); // path/to/create-t3-app/dist/index.js
const distPath = path.dirname(__filename); // path/to/create-t3-app/dist
export const PKG_ROOT = path.join(distPath, "../"); // path/to/create-t3-app
```

when PKG_ROOT is accurately pointing to the package root again, the template folder can be found no matter what directory the app is scaffolded from. 
CleanShot 2022-06-28 at 12 16 04

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # create-t3-app
 
+## 4.2.9
+
+- fix resolved path to package root
+
 ## 4.2.8
 
 ### Patch Changes
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "create-t3-app",
-  "version": "4.2.8",
+  "version": "4.2.9",
   "description": "Create web application with the t3 stack",
   "author": "Shoubhit Dash <shoubhit2005@gmail.com>",
   "license": "MIT",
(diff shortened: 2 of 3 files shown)


juliusmarminge:
> Haha I was also looking at that blog post and had the changes ready, was just investigating to see if there's a better solution.

Hehe. I didn't find any better approach and I don't think it is that hacky tbf.

nexxeln:
I'll merge this then I guess. Thank you again.

juliusmarminge:
It's not a big change so if another solution surfaces you can just change it again.

PR #66, opened 2022-06-28
feat: add publish workflow when creating releases

Omar-Belghaouti:
Hi, I really liked your work here and I wanted to contribute to your project by adding a workflow for a GitHub Action that publish this package automatically to npm registry when creating releases in your repo. You just need to add `NODE_AUTH_TOKEN` to your repo secrets and you should test it before using it if any thing occurs.

PR #70, opened 2022-06-28
feat: add release-it in workflow

asheeeshh:
this pr adds `release-it` for auto releasing the package to github and npm. 

things you need to do to make it work @nexxeln 
1. create a env file and follow the content of `.env.example` file i have added in the repo.
2. whenever you want to release a new version, run `pnpm run release` on terminal and proceed with yes for each prompt. It will automatically release the version on github and npm for you.

ps: it auto increments the version in pkg json so make sure you don't increment it yourself.

moreover, i don't think its interfering with the change set we are maintaining, i would still like to have @MWhite-22 's thoughts on this before you merge the pr.

image

image

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+GH_TOKEN=your_github_access_token_with_repo_scope
--- a/.release-it.json
+++ b/.release-it.json
@@ -1,0 +1,9 @@
+{
+  "github": {
+    "release": true,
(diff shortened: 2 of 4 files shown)


MWhite-22:
So I think, rather than adding in a package to do this, a GitHub action (as implemented in #66) is the more imperative way of releasing updates. 

Ultimately the decision lies with @nexxeln though.

PR #74, opened 2022-06-28
feat: add trpc-auth middleware

### LINKED ISSUE #38: Scaffold NextAuth TRPC getSession middleware if both are selected.
If NextAuth and TRPC are both selected, It's very likely that the user wants to have protected routes/queries.

The TRPC middleware to enable to enable protected routes/queries should be included so users don't have to write the boilerplate code themselves.

There's an example on the [TRPC Kitchen Sink](https://kitchen-sink.trpc.io/next-auth) site.

juliusmarminge:
Closes #38.

Adds protected routes when choosing both trpc and next-auth.

Added a simple CredentialsProvider so that you can test sessions out of the box without having to configure anything.

Haven't added a "showcase" homepage yet. Can get on that if requested.


juliusmarminge:
side note, the "multiple-copies-of-the-same-file" approach is becoming exhausting ngl... maybe time to look into jscodeshift or something to build the files instead of copying

juliusmarminge:
added the showcase since I had it available. I can remove it if you'd like.

Note. It (the showcase) doesn't support prisma (that's why there is the `!usingPrisma` check), that would require more setup I think, or perhaps it's cause I'm using sqlite?

nexxeln:
I have never used trpc middleware so I can't verify if this is correct, I'd love some confirmation.

> side note, the "multiple-copies-of-the-same-file" approach is becoming exhausting ngl... maybe time to look into jscodeshift or something to build the files instead of copying

I agree, I'll also look into it.

PR #76, opened 2022-06-29
feat: trpc next middleware

### LINKED ISSUE #38: Scaffold NextAuth TRPC getSession middleware if both are selected.

juliusmarminge:
Closes #38 

Sorry, I messed up the previous PR (#74).


iDarkLightning:
It's nicer to implement auth with the [route metadata](https://trpc.io/docs/metadata) feature in my opinion since it enabled routes to easily opt in and out of requiring authentication

juliusmarminge:
Good call. Made a poll on Discord to see which is more used

iDarkLightning:
Yeah performance wise meta might be slightly worse considering that. As far as I am aware you are correct that the middleware will only run for endpoints inserted after the middleware (unless there's some technicalities I'm missing).

nexxeln:
Is this ready to be merged now?

juliusmarminge:
Versioning and changelog left

nexxeln:
Alright

nexxeln:
Thanks!

PR #77, opened 2022-06-29
feat: add main as initial branch

Fedeorlandau:
Hello there,

When you run this tool it may initialize the repository with "master" as the principal branch. 

Is it ok if we specify "main"? Should we ask the user for the default branch name?

docs: https://git-scm.com/docs/git-init

Cheers,


nexxeln:
Sure why not haha. Thank you.

PR #81, opened 2022-07-03
fix-feat: extend validate app name to support paths and scoped packages

### LINKED ISSUE #72: project-name can no longer be used to change directory
**Describe the bug**
It is no longer possible to scaffold your app outside of the CWD due to the CLI's [validator](https://github.com/nexxeln/create-t3-app/blob/main/src/utils/validateAppName.ts).

**To Reproduce**

1. Run the CLI `npx create-t3-app`
2. Under project name, enter a path, e.g. `apps/web`.
3. The CLI's validator will warn that the name is not valid, see screenshots below.

**Expected behavior**
The app should scaffold with the name `web` under the `apps` directory.

**Screenshots**
CleanShot 2022-06-28 at 17 58 40
CleanShot 2022-06-28 at 18 00 39

juliusmarminge:
Closes #72 

Not quite sure what we agreed on here, but made some changes to extend the support of the `validateAppName` feature.

## Validation
- [x] The validation is only made on the app's name and not the full path, e.g. `some/dir/to/my/app` only validates the `app` part of the path.
- [x]  The validation supports scoped packages so `my-mono/@my-mono/app` validates `@my-mono/app` as valid. `my-mono/@my-mono/apps/web` is still not valid (should it? in that case the regex has to be changed)

## Parsing the name and path from the user's input
- [x] Inputting a scoped name as part of the path now parses that correctly, i.e. `some/dir/@mono/app` scaffolds the app under `some/dir/app`, but the app's name in the package.json is `@mono/app`.

## Examples
|Input | OK? | Parsed Path | Parsed App Name|
|------|-----|-------------|----------------|
|some/path/app | Yes | some/path/app | app|
|some/path/ju[nk | No | - | - |
|some/path/mono/@mono/app | Yes | some/path/mono/app | @mono/app |

What do you think?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,6 +8,7 @@ import { initializeGit } from "./helpers/initGit.js";
 import { logNextSteps } from "./helpers/logNextSteps.js";
 import { buildPkgInstallerMap } from "./installers/index.js";
 import { logger } from "./utils/logger.js";
+import { parseNameAndPath } from "./utils/parseNameAndPath.js";
 import { renderTitle } from "./utils/renderTitle.js";
 
 const main = async () => {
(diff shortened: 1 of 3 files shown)


nexxeln:
I'm not too sure tbh. I would like other people to leave their thoughts here. We have three options:

1. No validation
2. The validation in this PR
3. Change the validation to support all possible inputs

juliusmarminge:
Yeah most important for me is to be able to scaffold using a path.

Supporting scoped names may be overkill.

mauris:
How about we prompt 2 things?

1. the name of the app
2. the path to create the app (with default being cwd+app name)?

juliusmarminge:
That seems redundant and doesn't solve the problem of what we want to validate.

juliusmarminge:
Aouch that branch must have been accidentally deleted, sorry bout that. Should I redo it?

Issue #88, opened 2022-07-04
Shouldn't there be only one h1 tag per page?


mnik01:
**Describe the bug**
For example, in with-trpc-tw.tsx file, we have 5 h1 tags.

**Expected behavior**
To have one h1 and others h2

mnik01:
If i getting it right
Even if we use sections, it's recommended to use nesting of h1
As it mentioned here (with link to W3C)

![image](https://user-images.githubusercontent.com/39025497/177070596-92d4af02-42dd-4587-9b40-eababd37ba7c.png)
https://ahrefs.com/blog/h1-tag/#use-one-h1-per-page

nexxeln:
Thank you for opening this issue, we definitely want to stick with best practices. Will take a look at the PR in a bit.

nexxeln:
Implemented in #89

PR #89, opened 2022-07-04
fix(markup): multiple h1 replaced to h2

mnik01:
As it mentioned in issue 88 we have multiple h1 tags in one page. This PR fixing it

https://github.com/t3-oss/create-t3-app/issues/88


TheoBr:
This is good semantic html practice, lgtm 🙏

nexxeln:
Thank you for this.

PR #95, opened 2022-07-04
Text formatting and remove cursor for non link elements 

LeahyCC:
Small update to some UX things while poking around.

1. Remove `-` char, and lowercased `stack` on heading text... `This Stack uses:-` -> `This stack uses:` within templates.
2. Removed `cursor-pointer` on `div` element. Will still have hover effect but now only the link will show as clickable.


nexxeln:
Thanks!

PR #104, opened 2022-07-04
docs: update readme with links to npm and the contribution guide

Fedeorlandau:
Right now the badges are pointing to the svg file. When you click them it redirects you to that. I fixed it by adding the npm url and the contribution guide url


nexxeln:
This is great thanks!

PR #107, opened 2022-07-04
fix(cli): use correct spelling, fix short choices

jsjoeio:
Consistent use of the word "TypeScript" with a capital "S". This also fixes the `short` value to show "JavaScript" when that choice is selected.

See https://github.com/SBoudrias/Inquirer.js#prompt


nexxeln:
Thanks!

Issue #87, opened 2022-07-03
Consider setting tabWidth option in .prettierrc


devkevbot:
**Is your feature request related to a problem? Please describe.**
- The project does not have a default `tabWidth` setting in the `.prettierrc` file, which appears to allow for a local user's indentation settings to override the in-place indentation when editing files.
- This override behaviour is annoying when contributing because conveniences such as "format on save" will run using my editor's default settings (i.e. in VS Code) which causes _other lines_ to be formatted even though I'm only change one line.

**Describe the solution you'd like**
I surmise that adding `"tabWidth": <some preferred project value>` (see configuration [here](https://prettier.io/docs/en/configuration.html#basic-configuration)) would alleviate this annoyance.

**Describe alternatives you've considered**
Alternative solutions would also include creating a `.vscode` folder in the project root and setting indentation there, but that would not affect users who contribute through `Vim, Neovim, Sublime`, etc.

nexxeln:
Thank you, will add that soon.

amar-jay:
Done!!

PR #110, opened 2022-07-04
Merge

amar-jay:


amar-jay:
A prettier config option for vs code added
- tabWidth
-  printWidth
- semi-colons
- tabs or spaces

nexxeln:
Would have appreciated if you opened an issue before making to effort to make a PR. 

We won't be adding a prettier config to the template, see #25

nexxeln:
Just looked at your comment in #87. The issue was for adding prettier the repo not the template. I think you misinterpreted it 😄

Issue #108, opened 2022-07-04
next-auth: better env.example


wlechowicz:
**Is your feature request related to a problem? Please describe.**
When choosing to use next-auth, the generated .env-example file does not include config values template of things typically used with next-auth, which leads to warnings when running the app.

**Describe the solution you'd like**
When next-auth is selected, .env-example files should include `NEXTAUTH_SECRET=` and `NEXTAUTH_URL` as memos to teach and/or remind users that these need to be set up.

**Describe alternatives you've considered**
While console warnings such as `[next-auth][warn][NEXTAUTH_URL]` and `[next-auth][warn][NO_SECRET]` might be considered a "good enough" alternative, this app already includes required config keys in .env-example for Prisma (even when user does not opt in on Prisma!) and Github.

juliusmarminge:
NEXTAUTH_URL is only for production (right?) so no need to include it in a dev .env-example.

wlechowicz:
It's the secret that will sort of work in dev but will not in prod (they auto-generate a secret for dev).
The URL should be set always, but for things like Vercel previews and Vercel deployments, it will use VERCEL_URL.
https://next-auth.js.org/warnings#nextauth_url
https://next-auth.js.org/warnings#no_secret
https://next-auth.js.org/configuration/options#secret
The last one lists all the reasons why should not depend on the auto secret generation and that it will be removed, and it boggles my mind why they would ever come up with something like that in the first place.

juliusmarminge:
Oh right. You dont need it with Vercel was probably what I was thinking about.

Add 2 rows to the .env-example then. I dont think we need to parametrize it or autogenerate a secret to use. The file is an example and it is up to the user to set their own secrets.

juliusmarminge:
Oh right. You dont need it with Vercel was probably what I was thinking about.

Add 2 rows to the .env-example then. I dont think we need to parametrize it or autogenerate a secret to use. The file is an example and it is up to the user to set their own secrets.

wlechowicz:
I totally agree that generating a secret goes too far beyond scaffolding so it's not a concern of this app.

juliusmarminge:
Closed by #114

PR #114, opened 2022-07-04
feat: add next-auth config to env-example #108

wlechowicz:
Adds next-auth config fields to .env-example to help users opting in for next-auth resolve related warnings.


template/base/.env-example
```diff
@@ -4,5 +4,9 @@
 DATABASE_URL=
 
 # Next Auth
+NEXTAUTH_SECRET=
+NEXTAUTH_URL=
+
+# Github
```
juliusmarminge, line 10: The GITHUB_CLIENT and SECRET are still NextAuth env-vars

wlechowicz: I moved them out because I thought they are more generic (like GH_TOKEN for workflows), but they belong to the GithubProvider, so either section makes sense to me. Should I move them back?

juliusmarminge: Its not a big deal and I dont know whats more confusing: having a Github variable for Auth or having it separate and left wonder what it does.

wlechowicz: Case in point, it confused me too. I renamed it, how about now?

PR #115, opened 2022-07-04
feat: better ui for templates

asheeeshh:
discussed on discord

PR #118, opened 2022-07-05
fix(templates): h1 restored

mnik01:
When added UI changes to templates h1 tags was removed  😆

see https://github.com/t3-oss/create-t3-app/pull/115


mnik01:
Opened this PR to main cause dev on that moment does not have changes from 
https://github.com/t3-oss/create-t3-app/pull/115

mnik01:
> Opened this PR to main cause dev on that moment does not have changes from #115

Changed target to dev. Cause now it cause this changes

Dhravya:
Looks good to me~

asheeeshh:
i don't think it will make any difference as the font size remains the same.

nexxeln:
This is good semantics, thank you!

PR #120, opened 2022-07-05
Follow patterns of the project.

Fizzify:
Fix an issue in the _app.tsx where it would give this error. This is because the curly braces aren't needed, it is already getting it from "next/app", so you wouldn't need the curly braces anymore.

Before changes:
image

After changes:
image

### DIFF
--- a/template/base/src/pages/_app.tsx
+++ b/template/base/src/pages/_app.tsx
@@ -1,5 +1,5 @@
 import "../styles/globals.css";
-import type AppProps from "next/app";
+import type { AppProps } from "next/dist/shared/lib/utils";
 
 function MyApp({ Component, pageProps }: AppProps) {
   return <Component {...pageProps} />;
--- a/template/base/src/pages/_app.tsx
+++ b/template/base/src/pages/_app.tsx
@@ -1,5 +1,5 @@
 import "../styles/globals.css";
-import type { AppProps } from "next/app";
+import type AppProps from "next/app";
 
 function MyApp({ Component, pageProps }: AppProps) {
   return <Component {...pageProps} />;
--- a/template/base/src/pages/_app.tsx
+++ b/template/base/src/pages/_app.tsx
@@ -1,7 +1,7 @@
 import "../styles/globals.css";
 import type { AppProps } from "next/dist/shared/lib/utils";
 
-function MyApp({ Component, pageProps }: AppProps) {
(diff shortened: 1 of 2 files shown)


juliusmarminge:
The base app is created using `create-next-app`, see _app file here: https://github.com/vercel/next.js/blob/canary/packages/create-next-app/templates/typescript/pages/_app.tsx

I usually do: 
```ts
import type { AppType } from "next/dist/shared/lib/utils";

const MyApp: AppType = ({ Component, PageProps }) => { ... }
export default MyApp
```

and this is how we have in all other _app.tsx so maybe migrate that pattern here too?

Fizzify:
Ohhh wait I get what you said now.

template/base/src/pages/_app.tsx
```diff
@@ -1,5 +1,5 @@
 import "../styles/globals.css";
-import type { AppProps } from "next/app";
+import type { AppProps } from "next/dist/shared/lib/utils";
 
 function MyApp({ Component, pageProps }: AppProps) {
```
juliusmarminge, line 4: I was actually thinking of making this an arrow function like we have in most other places. 
```ts
import type { AppType } from "next/dist/shared/lib/utils";

const MyApp: AppType = ({ Component, pageProps }) => { ... }
export default MyApp;
```

What do you think?

template/page-studs/_app/with-auth.tsx
```diff
@@ -1,5 +1,5 @@
 import "../styles/globals.css";
-import type { AppProps } from "next/app";
+import type { AppProps } from "next/dist/shared/lib/utils";
 import { SessionProvider } from "next-auth/react";
 
 function MyApp({ Component, pageProps: { session, ...pageProps } }: AppProps) {
```
juliusmarminge, line 5: Same as above

template/base/src/pages/_app.tsx
Fizzify: Yeah, I think it will be cool, so that the code will be consistent throughout the project.

Fizzify:
Alright I made the PR to the main branch.

template/base/src/pages/_app.tsx
juliusmarminge: > If you want then I can just add those changes to this PR as well.

Yes, please do!

PR #122, opened 2022-07-05
feat: add middleware for protected routes

Fedeorlandau:
Hi there, on almost every next.js app I needed to add some sort of middleware to protect routes like /dashboard or /admin

What do you think of adding this middleware example on the next-auth boilerplate?

Regards.

P.S : I've also upgraded eslint-config-next to match the latest nextjs version


wlechowicz:
IMO feels a step over the border of the scaffolding realm.

Fedeorlandau:
> IMO feels a step over the border of the scaffolding realm.

I'd argue that is doing the same as the existing protected endpoint but for routes. Maybe removing the protected endpoint example is the real deal. 

I can do it if the people agree.

wlechowicz:
Actually, I think you're right. It's a valuable example of a working middleware and route matcher. One thing I'd suggest is that the comment could be moved to the top, like in the restricted.ts file, so it's the first thing that the user hits, and that it spells out that it's an example, just like the other file.

src/installers/next-auth.ts
```diff
@@ -36,8 +36,12 @@ export const nextAuthInstaller: Installer = async (
     "src/pages/api/restricted.ts",
   );
 
+  const middlewareSrc = path.join(nextAuthAssetDir, "middleware.ts");
+  const middlewareDest = path.join(projectDir, "src/middleware.ts");
```
juliusmarminge, line 40: Next middleware goes in `_middleware.ts`

template/addons/next-auth/middleware.ts
```diff
@@ -0,0 +1,20 @@
+// Example of a restricted route that only authenticated users can access https://next-auth.js.org/configuration/nextjs#middleware
+
+import { NextResponse } from "next/server";
+import type { NextRequest } from "next/server";
+import { getToken } from "next-auth/jwt";
+
+export async function middleware(request: NextRequest) {
```
juliusmarminge, line 7: Should we not use the [default middleware](https://next-auth.js.org/configuration/nextjs#basic-usage) prescribed by NextAuth.js?

src/installers/next-auth.ts
Fedeorlandau: They changed it to middleware.ts on the latest stable release. And now it goes outside pages

template/addons/next-auth/middleware.ts
Fedeorlandau: It depends. That middleware will automatically redirect you to your signin page. The one I added takes you to the homepage. If you think that is correct I can replace it with the one from the docs

src/installers/next-auth.ts
juliusmarminge: You get errors when trying to import: CleanShot 2022-07-06 at 01 26 23

template/addons/next-auth/middleware.ts
juliusmarminge: Yea I guess and that would be a 404 page...

Having a custom one seems over the top for a scaffold, but at the same time it better showcases how to use NextAuth so I think you are right in keeping it like this. Good call!

template/addons/next-auth/middleware.ts
```diff
@@ -0,0 +1,20 @@
...
+  const token = await getToken({
+    req: request,
+    secret: process.env.NEXTAUTH_SECRET,
+  });
+
+  if (!token) {
+    return NextResponse.redirect(new URL("/", request.url));
```
juliusmarminge, line 14: Tested this locally (and on a test-repo on Vercel) and I get an error that I got redirected too many times. Have you tested this successfully? Using the [other redirect method](https://nextjs.org/docs/messages/middleware-relative-urls) seems to work better.

juliusmarminge (approved):
Works as intended. Now it's just deciding if we want this or not and I leave that to @nexxeln

nexxeln:
I think this would be valuable. 
@TheoBr What do you think?

template/addons/next-auth/middleware.ts
TheoBr, line 14: This should probably redirect to "login" and have a comment explaining what it does

template/addons/next-auth/middleware.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+}
+
+export const config = {
+  matcher: "/profile/:path*",
```
TheoBr, line 19: Might be good to rename this in a way where it's more clear that "this is to protect a route"

TheoBr (commented):
Down to include this. Would like to be way more explicit in the code, file naming etc that "this is a way to BLOCK ACCESS TO ROUTES" so we're not guiding users down this path incorrectly _just because it's there_

template/addons/next-auth/middleware.ts
Fedeorlandau: Done!

huylee99: Hello @Fedeorlandau and @TheoBr,

Should we add `return NextResponse.next();` if there is token? My linter warns that "Not all code paths return a value", so I think we should add it here. Furthermore, it would be perfect if there is a `redirect` query in the URL for redirect after login process.

Fedeorlandau: > Hello @Fedeorlandau and @TheoBr,
> 
> Should we add `return NextResponse.next();` if there is token? My linter warns that "Not all code paths return a value", so I think we should add it here. Furthermore, it would be perfect if there is a `redirect` query in the URL for redirect after login process.

Done! Thank you! Would you mind checking now?

huylee99: Yeah, it's good now!

> > Hello @Fedeorlandau and @TheoBr,
> > Should we add `return NextResponse.next();` if there is token? My linter warns that "Not all code paths return a value", so I think we should add it here. Furthermore, it would be perfect if there is a `redirect` query in the URL for redirect after login process.
> 
> Done! Thank you! Would you mind checking now?

Yeah, it's good now!

Fedeorlandau:
> I thought about this more and it feels like something we're not building. This is more an "example", one that's already in other sample repos 🤔🤔

It's an example indeed. Again, the reason why I added this is because we have another example for restricted endpoints which is a copy of the [next-auth protected endpoint example](https://next-auth.js.org/getting-started/example#backend---api-route). 
IMO if this PR doesn't go in, then the [restricted endpoint](https://github.com/t3-oss/create-t3-app/blob/main/template/addons/next-auth/restricted.ts) must go away.

I really don't know what do lol

CarlosGomez-dev:
I think it's valuable to have both, it's common to want to protect endpoint and routes, and we already have the endpoint example

Fedeorlandau:
Closing this PR as it's implying that this is the right way to do protected routes. Something that we can't really guarantee.

If you're looking to do something like this please refer to next-auth official docs.

Cheers!

PR #127, opened 2022-07-05
fix: incompatible git version

### LINKED ISSUE #96: Git fails to innitialize with create t3-app
**Describe the bug**
When I run the create `yarn create t3-app`
After choosing my preferred options, the scaffold builds successfully, however during the build, it attempts to initialise git and it fails.
Git can be initialised manually later by running `git init` after the build

**To Reproduce**
Steps to reproduce the behaviour:
- Run  yarn create  t3-app

2. See  error `✖ Failed: could not initialise git`

**Expected behaviour**
The git initialisation should not fail 

**Additional information**
- I am on a linux distro i.e Zorin Os
- I have git installed

**Screenshots**
t3

juliusmarminge:
Closes #96 

As discussed in above referenced issue.

The flag `initial-branch` will only be used if the version is compatible (>=2.28)

Don't know if we should switch branch to `main` by `git checkout -b main` otherwise or just leave them on `master`.

### DIFF
--- a/src/helpers/initGit.ts
+++ b/src/helpers/initGit.ts
@@ -10,7 +10,18 @@ export const initializeGit = async (projectDir: string) => {
   logger.info("Initializing Git...");
   const spinner = ora("Creating a new git repo...\n").start();
   try {
-    await execa("git init --initial-branch=main", { cwd: projectDir });
+    let initCmd = "git init --initial-branch=main";
+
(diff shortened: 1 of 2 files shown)


asheeeshh:
i think we should change the branch in second case too, to keep the same behavior

nexxeln:
Looks good to me.

nexxeln:
Can you make this PR to the main branch lol

Issue #128, opened 2022-07-05
Adding git (and no-install) prompt


juliusmarminge:
**Is your feature request related to a problem? Please describe.**
- I noticed when creating a new Astro project that they have prompts for installing dependencies and initializing the git repo. I find this easier than having to remember a flag to provide.

**Describe the solution you'd like**
- Adding in prompts similar to the ones below (we currently don't support no-install but eventually we probably should):
![CleanShot 2022-07-06 at 01 18 17](https://user-images.githubusercontent.com/51714798/177432417-966567ab-eae8-474d-9ee0-d24dc8ee620a.png)

**Describe alternatives you've considered**
- We have flags right now but for new users that doesn't know they exist or even old users who just forget to provide them, a prompt seems easier and more intuitive.

MWhite-22:
Agreed, but we should keep both as those without interactive terminals, or those that just want o bypass the options, can still pass in flags.

Alexanderdunlop:
Noticed when skipping npm install the console still logs "Adding packages..."
Is that intentional?

<img width="529" alt="image" src="https://user-images.githubusercontent.com/22138454/177910051-c9a6b6fe-b650-443b-b2e1-6792fbbd0143.png">

juliusmarminge:
Implemented in #130 and #133.

juliusmarminge:
> Noticed when skipping npm install the console still logs "Adding packages..."
Is that intentional?

I intentionally made the message different from when it is actually installing the dependencies. Maybe change to like `Adding boilerplate for <package>`? Also should probably not say anything if a blank app is requested without any of the extras.

PR #133, opened 2022-07-06
feat: add noInstall flag and prompt

juliusmarminge:
Has been discussed previously (#37, #58, #128 etc.).

Enables the `--noInstall` CLI flag and adds a prompt for it. This skips the installation process and only adds the dependency to the `package.json` and copies over the boilerplate.

I ended up using `npm show <pkg> version` for this, so for tRPC it's still kinda slow (see snippet below) since there are a lot of dependencies whose versions needs to be fetched. An alternative is to keep a config-file with all the latest versions and update that from time to time. 

There are probably also some awaits we could remove to optimize and do some more stuff async to speed up the process if we care about the time it takes to scaffold. I timed 36s to scaffold a full app with install and 18s (half) without installs.

Will document and clarify with some comments tomorrow.

CleanShot 2022-07-07 at 01 30 00

### DIFF
--- a/src/helpers/createProject.ts
+++ b/src/helpers/createProject.ts
@@ -5,13 +5,17 @@ import { installPackages } from "./installPackages.js";
 import { scaffoldProject } from "./scaffoldProject.js";
 import { selectAppFile, selectIndexFile } from "./selectBoilerplate.js";
 
-export const createProject = async (opts: {
+interface CreateProjectOptions {
   projectName: string;
   packages: PkgInstallerMap;
   noInstall: boolean;
-}) => {
-  const { projectName, packages, noInstall } = opts;
+}
 
+export const createProject = async ({
+  projectName,
+  packages,
+  noInstall,
+}: CreateProjectOptions) => {
   const pkgManager = getUserPkgManager();
   const projectDir = path.resolve(process.cwd(), projectName);
 
--- a/src/helpers/createProject.ts
+++ b/src/helpers/createProject.ts
@@ -5,18 +5,21 @@ import { installPackages } from "./installPackages.js";
 import { scaffoldProject } from "./scaffoldProject.js";
 import { selectAppFile, selectIndexFile } from "./selectBoilerplate.js";
 
-export const createProject = async (
-  projectName: string,
-  packages: PkgInstallerMap,
-) => {
+export const createProject = async (opts: {
+  projectName: string;
+  packages: PkgInstallerMap;
+  noInstall: boolean;
+}) => {
+  const { projectName, packages, noInstall } = opts;
+
   const pkgManager = getUserPkgManager();
   const projectDir = path.resolve(process.cwd(), projectName);
 
   // Bootstraps the base Next.js application
-  await scaffoldProject(projectName, projectDir, pkgManager);
+  await scaffoldProject({ projectName, projectDir, pkgManager, noInstall });
 
   // Install the selected packages
-  await installPackages(projectDir, pkgManager, packages);
+  await installPackages({ projectDir, pkgManager, packages, noInstall });
 
   // TODO: Look into using handlebars or other templating engine to scaffold without needing to maintain multiple copies of the same file
   await selectAppFile(projectDir, packages);
--- a/src/helpers/installPackages.ts
+++ b/src/helpers/installPackages.ts
@@ -5,19 +5,28 @@ import ora from "ora";
 import { logger } from "../utils/logger.js";
 
 // This runs the installer for all the packages that the user has selected
-export const installPackages = async (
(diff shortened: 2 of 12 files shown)


nexxeln (commented):
Maybe we could not print `Using npm` if noInstall is picked?

juliusmarminge:
> Maybe we could not print Using npm if noInstall is picked?

Makes sense, done. 
CleanShot 2022-07-07 at 09 44 54

nexxeln:
Looks good to me. Just waiting for @MWhite-22 's review

src/helpers/createProject.ts
```diff
@@ -5,18 +5,21 @@ import { installPackages } from "./installPackages.js";
 import { scaffoldProject } from "./scaffoldProject.js";
 import { selectAppFile, selectIndexFile } from "./selectBoilerplate.js";
 
-export const createProject = async (
-  projectName: string,
-  packages: PkgInstallerMap,
-) => {
+export const createProject = async (opts: {
```
Fedeorlandau, line 8: Could you extract this type like this:

```tsx
interface CreateProjectProps {
    opts: {
         projectName: stirng;
         ....
    }
}
```

I usually put the interfaces/types just right above the function.

src/helpers/installPackages.ts
```diff
@@ -5,19 +5,28 @@ import ora from "ora";
 import { logger } from "../utils/logger.js";
 
 // This runs the installer for all the packages that the user has selected
-export const installPackages = async (
-  projectDir: string,
-  pkgManager: PackageManager,
-  packages: PkgInstallerMap,
-) => {
-  logger.info("Installing packages...");
+export const installPackages = async (opts: {
```
Fedeorlandau, line 8: You can reuse the type I said before here.

src/helpers/installPackages.ts
```diff
@@ -5,19 +5,28 @@ import ora from "ora";
...
-      const spinner = ora(`Installing ${name}...`).start();
-      await opts.installer(projectDir, pkgManager, packages);
+  logger.info(`${noInstall ? "Adding" : "Installing"} packages...`);
+
+  for (const [name, pkgOpts] of Object.entries(packages)) {
+    if (pkgOpts.inUse) {
+      const spinner = ora(
+        `${noInstall ? "Adding" : "Installing"} ${name}...`,
+      ).start();
+      await pkgOpts.installer({ projectDir, pkgManager, packages, noInstall });
```
Fedeorlandau, line 23: Can this fail? If it does we're triggering the spinner.succeed anyway. Maybe try/catch? Not sure if we're throwing errors to a top-level handler

src/helpers/logNextSteps.ts
```diff
@@ -3,14 +3,20 @@ import { getUserPkgManager } from "../utils/getUserPkgManager.js";
 import { logger } from "../utils/logger.js";
 
 // This logs the next steps that the user should take in order to advance the project
-export const logNextSteps = (
-  projectName: string,
-  packages: PkgInstallerMap,
-) => {
+export const logNextSteps = (opts: {
```
Fedeorlandau, line 6: Reuse types

src/helpers/scaffoldProject.ts
```diff
@@ -9,14 +9,22 @@ import { type PackageManager } from "../utils/getUserPkgManager.js";
 import { logger } from "../utils/logger.js";
 
 // This bootstraps the base Next.js application
-export const scaffoldProject = async (
-  projectName: string,
-  projectDir: string,
-  pkgManager: PackageManager,
-) => {
+export const scaffoldProject = async (opts: {
```
Fedeorlandau, line 12: Reuse types here as well

src/installers/index.ts
```diff
@@ -15,11 +15,12 @@ export const availablePackages = [
 
 export type AvailablePackages = typeof availablePackages[number];
 
-export type Installer = (
-  projectDir: string,
-  packageManager: PackageManager,
-  packages: PkgInstallerMap,
-) => Promise<void>;
+export type Installer = (opts: {
```
Fedeorlandau, line 18: Oh you have it here, maybe you could use it in the scaffoldProject and lognextstep

Fedeorlandau (commented):
Nice feature! Left a few comments ts related. Thanks!

juliusmarminge:
@Fedeorlandau I extracted the types but didn't bother reusing them. There is probably a good way to do this but I don't have time for that now. If someone wants to reuse the types it should be easier now at least.

src/helpers/installPackages.ts
juliusmarminge: Not that I'm aware of

PR #138, opened 2022-07-07
fix: remove duplicated types

Fedeorlandau:
from: https://github.com/t3-oss/create-t3-app/pull/133#issuecomment-1177332751

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,5 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
+npx tsc --noEmit --skipLibCheck
 npx lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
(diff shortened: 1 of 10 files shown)


MWhite-22 (changes requested):
Can we rename "CommonOptions" into "InstallerOptions" to provide a little more semantic clarity on what the general purpose of the type is?

Otherwise looks OK to me

Fedeorlandau:
Done @MWhite-22 ! Thanks for the feedback

juliusmarminge:
Is it just me who get a bunch of type-errors?

CleanShot 2022-07-07 at 15 32 02

juliusmarminge:
> Maybe instead of running build --onsuccess tsc or tsc on pre_commit, we can do it on pre-push? is that possible?

Still the problem of getting stuff onto GitHub to share if they aren't 100% ready, but again that might just be me doing that?

Fedeorlandau:
Makes sense to do it on CI. Let's remove it from husky. Typescript can become an issue if you're doing some experimental work and you want to commit your stuff

Fedeorlandau:
> would you be open to resubmitting the pr so we can test the CI?

Done! All green

juliusmarminge:
perfect 👍🏼

Issue #140, opened 2022-07-07
A lot of tiny big wins


KATT:
## Table of contents

- [Table of contents](#table-of-contents)
- [Background](#background)
- [Areas of improvements](#areas-of-improvements)
  - [Safe environment variables](#safe-environment-variables)
  - [Type-safety](#type-safety)
  - [Continuous Integration](#continuous-integration)
  - [E2E testing](#e2e-testing)
  - [Integration testing](#integration-testing)
  - [Scripts](#scripts)
  - [VSCode extensions and settings](#vscode-extensions-and-settings)
  - [tRPC inference helpers](#trpc-inference-helpers)
- [Describe alternatives you've considered](#describe-alternatives-youve-considered)
- [Ending notes](#ending-notes)

## Background

I'm the creator of [tRPC](https://github.com/trpc/trpc) and Internet-friend of @theobr.

https://twitter.com/alexdotjs/status/1545036195331055616

I've been maintaining [this repo](https://github.com/trpc/examples-next-prisma-starter) for over a year and I've kept updating it with many tiny big details over time which represents how I do things in production.

### Safe environment variables

Environment variables should be validated when a project is built. Having the wrong env vars is a very common production issue that is hellish as you usually don't notice it straight-away. It's better to just crash the build if something's wrong.

Zod to the rescue!

- https://github.com/trpc/examples-next-prisma-starter/blob/a07a8d1bd1ed4a9af4fd7cd6130ee92383082419/src/server/env.js#L1-L24
- https://github.com/trpc/examples-next-prisma-starter/blob/a07a8d1bd1ed4a9af4fd7cd6130ee92383082419/next.config.js#L3
- It's a `.js`-file because Next.js [doesn't support `next.config.ts`](https://github.com/vercel/next.js/issues/5318), but it can still be validated by TypeScript: https://github.com/trpc/examples-next-prisma-starter/blob/a07a8d1bd1ed4a9af4fd7cd6130ee92383082419/tsconfig.json#L25-L31

### Type-safety

You can have `next.config.js` typesafe.

https://github.com/trpc/examples-next-prisma-starter/blob/main/next.config.js

If you use `publicRuntimeConfig`, you can have that inferred in a typesafe manner: https://github.com/trpc/examples-next-prisma-starter/blob/main/src/utils/publicRuntimeConfig.ts

### Continuous Integration

It's very time consuming to setup CI with a proper DB, Playwright, testing, build, etc. Most production projects need it.

[...]

PR #147, opened 2022-07-07
feat: add environment variables validation

piotrski:
# Add runtime environment variables validation

---

In #140 @KATT suggested adding runtime environment variables validation using zod. Since `env.js` is imported inside `next.config.js` it can't ba a typescript file. There are two templates for `env.js`, one for just prisma `env-prisma.js` and another one for prisma + next-auth `env-prisma-auth.js`

---

## Screenshots
Screenshot 2022-07-07 at 20 58 14

💯


piotrski:
> On my phone, but think you should add something to the tsconfig too in order to get the file validated with ts ☺

I've added `**/*.js` to tsconfig.

piotrski:
> Hi there! Please remember to update the PR title so we get it on the changelog.
> 
> feat: add environment variables validation
> 
> Cheers!

Should I change it to something else?

Fedeorlandau:
> > Hi there! Please remember to update the PR title so we get it on the changelog.
> > feat: add environment variables validation
> > Cheers!
> 
> Should I change it to something else?

It's perfect now! Thanks

Quick question, what happens when just "next-auth + trpc" or just "next-auth"  is enabled?

TheoBr (approved):
I do really like this

juliusmarminge:
The only problem with this as it stands, is that it would make the initial app fail to build because of the example .env not following this validation-schema. 

I just added a CI step that builds the default t3-app and this would fail if we do not include a better `.env` by default that can be parsed with this validation.

juliusmarminge:
> There are two templates for env.js, one for just prisma env-prisma.js and another one for prisma + next-auth env-prisma-auth.js

Also include one for just NextAuth.js `env-auth.js` to cover all the cases.

piotrski:
I've moved schema to `env-schema.js` to keep the code DRY and added env validation to the base template. Now addons are just need to override `env-schema.js`. I've tested **base config**, **prisma**, **prisma + auth**, **auth**, **auth + tailwind**

src/installers/next-auth.ts
```diff
@@ -38,8 +38,12 @@ export const nextAuthInstaller: Installer = async ({
     "src/pages/api/restricted.ts",
   );
 
+  const envSchemaSrc = path.join(nextAuthAssetDir, "env-auth.js");
+  const envSchemaDest = path.join(projectDir, "src/server/env-schema.js");
```
juliusmarminge, line 42: you should probably do some more checking to make sure the correct schema is selected. 

as it is now, are you sure that the Prisma-installer will overwrite this and select the `env-prisma-auth.js` _after_ this? if it runs the other way around this will override in the other direction if that makes sense?

_maybe_ having a separate installer for this would make handling this logic easier?

src/installers/prisma.ts
```diff
@@ -41,6 +41,12 @@ export const prismaInstaller: Installer = async ({
   const sampleApiRouteSrc = path.join(prismaAssetDir, "sample-api.ts");
   const sampleApiRouteDest = path.join(projectDir, "src/pages/api/examples.ts");
 
+  const envSchemaSrc = path.join(
```
juliusmarminge, line 44: same as above

template/addons/next-auth/api-handler-prisma.ts
```diff
@@ -5,14 +5,15 @@ import CredentialsProvider from "next-auth/providers/credentials";
 // Prisma adapter for NextAuth, optional and can be removed
 import { PrismaAdapter } from "@next-auth/prisma-adapter";
 import { prisma } from "../../../server/db/client";
+import { env } from "../../../server/env";
```
juliusmarminge, line 8: in your installers you copy this file to `server/env-schema` or am I missing something?

template/addons/prisma/env-prisma-auth.js
```diff
@@ -0,0 +1,12 @@
...
+const envSchema = z.object({
+  DATABASE_URL: z.string().url(),
+  NODE_ENV: z.enum(["development", "test", "production"]),
+  NEXTAUTH_SECRET: z.string(),
+  NEXTAUTH_URL: z.string().url(),
+  GITHUB_ID: z.string(),
+  GITHUB_SECRET: z.string(),
+});
+
+module.exports.envSchema = envSchema;
```
juliusmarminge, line 12: these can not be ESM, right?
```js
export const envSchema = ...
```

juliusmarminge (changes requested):
I like this. Left some comments. As for the concern in my comment previously, would this prevent the app from building if the .env schema is not valid or is it just runtime?

template/addons/next-auth/api-handler-prisma.ts
piotrski: I decided to split it into two separate files to keep it dry
- `env-schema.js` - contains the schama overwritten by the installers
- `env.js` - parses schema, formats errors and exports validated env variables

template/addons/prisma/env-prisma-auth.js
piotrski: Yes, but I think the `.mjs` might be supported

.github/workflows/PR-CI.yml
```diff
@@ -95,4 +95,5 @@ jobs:
       # has to be scaffolded outside the CLI project so that no lint/tsconfig are leaking
       # through. this way it ensures that it is the app's configs that are being used
       - run: pnpm start -y ../ci-test-app
+      - run: mv ../ci-test-app/.env-example ../ci-test-app/.env
```
juliusmarminge, line 98: oh I like this approach. this way we still leave it up to the user to make sure they set their env-vars whilst allowing the CI to test-build the app 👍🏼

src/installers/next-auth.ts
piotrski: I could move it to a separate installer for now. We only have 3 variants, this should be manageable

juliusmarminge: yes, until there are other stuff being made with ast/codeshift stuff there is no reason to do that for this one

juliusmarminge:
Looks good! Update the branch and I'll merge

piotrski:
> Looks good! Update the branch and I'll merge

Updated 👍

juliusmarminge:
Thank you 🙏!

PR #156, opened 2022-07-09
chore: install pnpm with corepack

nickmccurdy:
# Install pnpm with corepack

---

Install pnpm with corepack automatically instead of installing it globally, which can interfere with existing corepack users.

---

💯


nexxeln:
Corepack is still experimental and we won't be recommending it.

Please open an issue first from next time so that we can discuss the issue before you make a PR.

PR #159, opened 2022-07-09
fix: replaced process with typeof window since it's deprecated

thegeeko:
# replaced process.browser with typeof window

---
replaced process.browser with typeof window since it's depreciated refer to this [comment](https://github.com/vercel/next.js/issues/5354#issuecomment-520305040)

---
💯

### DIFF
--- a/template/page-studs/_app/with-trpc.tsx
+++ b/template/page-studs/_app/with-trpc.tsx
@@ -13,7 +13,7 @@ const getBaseUrl = () => {
   if (typeof window !== "undefined") {
     return "";
   }
-  if (process.browser) return ""; // Browser should use current path
+  if (typeof window !== "undefined") return ""; // Browser should use current path
   if (process.env.VERCEL_URL) return `https://${process.env.VERCEL_URL}`; // SSR should use vercel url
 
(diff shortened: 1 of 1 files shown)


nexxeln:
The tRPC docs use this, so we're using it too. Are you sure this works with tRPC?

thegeeko:
I'm not sure but I can't see why tRPC could have a problem with it .. I'll do some testing

thegeeko:
it works just with and without tRPC ssr enabled

nexxeln:
We already have this, no need to check for the same condition twice. Thank you but we'll be closing this.

template/page-studs/_app/with-trpc.tsx
```diff
@@ -13,7 +13,7 @@ const getBaseUrl = () => {
   if (typeof window !== "undefined") {
     return "";
   }
```
asheeeshh, line 15: here ^

asheeeshh, line 15: here ^

thegeeko: idk how I missed this .. sry .. but then isn't the next process.browser technically the same thing shouldn't u  remove it?? or  it's there just to be sure

PR #167, opened 2022-07-10
chore: ci improvements

JacobMGEvans:
# CI Improvements 

- [x] I performed a functional test on my final commit [(tests CI in my fork)](https://github.com/JacobMGEvans/create-t3-app/runs/7272582273?check_suite_focus=true)

---

Installing deps globally once for all jobs, caching, and actions updates.
- Significantly faster CI per Pull Request or Push.
- Currently using minimal caching, can add caching later that actually skips steps when nothing has changed. 
- Updated all the GitHub Actions to their latest versions, has performance, security and other benefits.
- BIG change is that moved "jobs" into "steps" which allows the usage of installed deps, pnpm, and node to be reused between commands rather than rebuilding the environment for each check.
- Very different jobs can be moved into their own Workflow or Job, similar tasks should all be in the same job i.e. "PR Checks"
---

## Screenshots
"Fork Testing" 
image

💯

### DIFF
--- a/.github/workflows/PR-CI.yml
+++ b/.github/workflows/PR-CI.yml
@@ -4,85 +4,20 @@ name: PR-CI
 
 on:
   pull_request:
-    branches: "*"
-
-  # Allows you to run this workflow manually from the Actions tab
-  workflow_dispatch:
-
+    branches: ["*"]
+  
 jobs:
-  lint:
-    name: Run ESLint
+  install-deps:
     runs-on: ubuntu-latest
+    name: Run pnpm install
     steps:
-      - uses: actions/checkout@v2
-      - uses: pnpm/action-setup@v2
+      - uses: actions/cache@v3
+        id: pnpm-cache
         with:
-          version: 7.2.1
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 16
-          cache: 'pnpm'
-      - run: pnpm install
-      - run: pnpm lint:check
-      
-  lint-pr:
-    name: Lint the PR title
-    runs-on: ubuntu-latest
-    steps:
-      - uses: amannn/action-semantic-pull-request@v4
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      
-  prettier:
-    name: Run Prettier Check
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: pnpm/action-setup@v2
-        with:
-          version: 7.2.1
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 16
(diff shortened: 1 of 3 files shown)


JacobMGEvans:
Oh no.... how did that get in the PR? That `if` is haunting me lmao

JacobMGEvans:
> Oh no.... how did that get in the PR? That `if` is haunting me lmao

Destroyed it's existence! lol

juliusmarminge:
So I think the downside to having them as steps instead of jobs is that you only get a single output in the PR and to see which one failed you have to go into the action in order to see the steps. Maybe the "performance gain" is worth it?

Output with jobs:
image

Output with steps:
image
image

JacobMGEvans:
> So I think the downside to having them as steps instead of jobs is that you only get a single output in the PR and to see which one failed you have to go into the action in order to see the steps. Maybe the "performance gain" is worth it?

Yeah the other thing you're doing is not rebuilding an environment for each task that is all part of the PR check, like lint, typecheck, etc... Each one is checking out the repo, setting up pnpm, setting up Node, even installing dependencies. This moves those into one setup for the environment then runs the tasks for PR checking. 

Something else this makes possible is now you can cache, I was planning on doing next, the modules, linting, typechecking,  once and it would be used by each task, potentially hitting the cache and skipping the task altogether.

JacobMGEvans:
The only one I see potentially being a good one to move into its own `job` here is the `Build an outputted t3-app` which could be dependent on the `PR check` job passing 
```yaml 
 build-t3-app:
    name: Build an outputted t3-app
    needs: pull-request-check # This would mean if PR check fails for any reason it doesn't try to run the job
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: pnpm/action-setup@v2
        with:
          version: 7.2.1
      - uses: actions/setup-node@v3
        with:
          node-version: 16
          cache: "pnpm"
      - run: pnpm install
      - run: pnpm build
      # has to be scaffolded outside the CLI project so that no lint/tsconfig are leaking
      # through. this way it ensures that it is the app's configs that are being used
      - run: pnpm start -y ../ci-test-app
      - run: cd ../ci-test-app && pnpm build
```

juliusmarminge:
How doyou want to do @nexxeln ? I know you were a fan of having the separate output?

> Something else this makes possible is now you can cache

Although the CI right now is quite minimal and doesnt really take too long, this benefit seems good to have for future expansion.

Another thing i mentioned in DC, can the steps run in parallel, like the format and lint? That would also speed up the process.

Keeping the `build-t3-app` as its separate job is a good idea. Since it's dependent on the other steps, can you use the build from that? So the only thing that job has to do is run and build the output?

PR #68, opened 2022-06-28
Warn node 18

juliusmarminge:
Closes #59 

CLI now warns the user if they are using Node 18 about the current [issue with Next-Auth](https://github.com/nextauthjs/next-auth/issues/4575).
CleanShot 2022-06-28 at 13 36 17

Issue #57, opened 2022-06-27
yarn create t3-app crashes my M1 mac.


matt123miller:
**Describe the bug**
When running `yarn create t3-app` and saying yes to everything my M1 mac eventually crashes while trying to install prisma. This has happened twice in a row now. I include M1 because it might be ARM mac specific.

**To Reproduce**

M1 Macbook Air
Installed through nvm
node 18.4.0
yarn 3.2.1

1. Run this '...'

`yarn create t3-app`

Say yes to everything.

Wait

2. See error

The terminal hung while installing prisma, eventually crashing my whole laptop. The emergency mac Force Quit Applications modal appeared, spinning beachball I closed the terminal, then everything else while my laptop crawled along, then it restarted itself.

I've included the error details I received on restart.

I'll retry this again tomorrow and record a video on my phone.

```
panic(cpu 0 caller 0xfffffe002eac3a94): userspace watchdog timeout: no successful checkins from WindowServer in 149 seconds
service: logd, total successful checkins since load (1218 seconds ago): 119, last successful checkin: 0 seconds ago
service: WindowServer, total successful checkins since load (1188 seconds ago): 101, last successful checkin: 149 seconds ago
service: opendirectoryd, total successful checkins since load (1218 seconds ago): 118, last successful checkin: 0 seconds ago

Debugger message: panic
Memory ID: 0xff
OS release type: User
OS version: 21F79
Kernel version: Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:29 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T8101
Fileset Kernelcache UUID: 6761FE24D814CDDD6BDA1ADB023146F8
Kernel UUID: B747B891-04EC-3617-95EC-05E037610520
iBoot version: iBoot-7459.121.3
secure boot?: YES
Paniclog version: 13
KernelCache slide: 0x00000000267fc000
KernelCache base:  0xfffffe002d800000
Kernel slide:      0x0000000026fa4000
Kernel text base:  0xfffffe002dfa8000
Kernel text exec slide: 0x000000002708c000
Kernel text exec base:  0xfffffe002e090000
mach_absolute_time: 0x6f651c1f0
Epoch Time:        sec       usec
  Boot    : 0x62b9e73a 0x000c5373
  Sleep   : 0x00000000 0x00000000
  Wake    : 0x00000000 0x00000000
  Calendar: 0x62b9ec13 0x0006859c

Zone info:
[...]

PR #173, opened 2022-07-11
feat: add check for yarn 3

c-ehrlich:
# Add check for Yarn 3

---

This commit adds a warning when the CLI is run using Yarn 3. This is to alert the users of possible crashes as seen in https://github.com/t3-oss/create-t3-app/issues/57.

I followed the format of https://github.com/t3-oss/create-t3-app/pull/68 as it is a similar issue.

The `.yarnrc.yml` file is necessary so that the CLI doesn't crash before we can even check whether the user is on yarn 3. See https://github.com/chalk/chalk/issues/531.

---

## Screenshots

(screenshot has a typo in it. i fixed it, but didn't want to create another screenshot as I had already uninstalled yarn 3. Hope that's ok)

Screenshot

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,0 +1,5 @@
+packageExtensions:
+  chalk@5.0.1:
+    dependencies:
+      "#ansi-styles": npm:ansi-styles@6.1.0
+      "#supports-color": npm:supports-color@9.2.2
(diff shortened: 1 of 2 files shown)


nexxeln:
Thank you. 

We don't recommend any package manager, so you can change the message to this. Also can we link that issue in the warning?
> package manager such as npm, pnpm, or Yarn Classic.

juliusmarminge (approved):
I like this. Seems like a lot of people are running Yarn 3 and warning them pushing them onto another pkgManager seems like a good idea.

nexxeln:
Looks good thank you!

Issue #176, opened 2022-07-11
Include `user.id` on session


TheoBr:
I feel like I see this particular issue come up in the discord at least a few times a week, and it's always bothered me.

Should be as simple as:

```ts
// in [...nextAuth].ts
callbacks: {
  async session({ session, user }) {
    session.user.id = user.id;
    return Promise.resolve(session);
  },
}

// in a random next-auth.d.ts
declare module "next-auth" {
  /**
   * Returned by `useSession`, `getSession` and received as a prop on the `Provider` React Context
   */
  interface Session {
    userId: string;
  }
}
```

Ideally NextAuthJS would include this 😅 hoping us including it can work as a forcing function for them to get on it

TheoBr:
Cut an issue on NextAuthJS because this should be fixed there IMO

https://github.com/nextauthjs/next-auth/issues/4903

TheoBr:
Unhappy w/ response on NextAuth side, let's do this here instead

bhatvikrant:
Hey I can pick this up @TheoBr

bhatvikrant:
Yeah, that happened by mistake LoL @nexxeln

TheoBr:
@bhatvikrant looks correct to me

PR #180, opened 2022-07-12
chore: include user.id on session

bhatvikrant:
# Include `user.id` on session 

---

## Summary
fixes https://github.com/t3-oss/create-t3-app/issues/176
1. added callbacks in `api-handler-prisma.ts` and `api-handler.ts`
2. included `next-auth.d.ts` only when it is a `next-auth` scaffolding
---


template/base/next-auth.d.ts
```diff
@@ -0,0 +1,12 @@
+import { DefaultSession } from "next-auth";
```
bhatvikrant, line 1: `next-auth.d.ts` should probably not be included in `base/` since it should not be always present. 
I was not sure to how to add it conditionally. Please give me some idea as to how I can do it, I will update the PR then.

ochicf: I think these would be the steps in order to do it conditionally:
1. move the file to `template/addons/next-auth` folder
2. in `src/installers/next-auth.ts`, copy the file by using the same pattern as done with the others files:
   - add src and dest variables (be sure to set the correct paths)
   - copy the file by adding the call in the `Promise.all` array

This way the file should only be copied when the next-auth package in installed.

bhatvikrant: Thanks, @ochicf, it helped a lot! 
I have done the changes

template/addons/next-auth/api-handler-prisma.ts
```diff
@@ -7,6 +7,13 @@ import { PrismaAdapter } from "@next-auth/prisma-adapter";
 import { prisma } from "../../../server/db/client";
 
 export const authOptions: NextAuthOptions = {
+  // Include user.id on session
+  callbacks: {
+    async session({ session, user }) {
+      session.user.id = user.id;
+      return Promise.resolve(session);
+    },
```
balazsorban44, line 15:
```suggestion
    session({ session, user }) {
      session.user.id = user.id;
      return session;
    },
```

template/addons/next-auth/api-handler.ts
```diff
@@ -3,6 +3,13 @@ import GithubProvider from "next-auth/providers/github";
 import CredentialsProvider from "next-auth/providers/credentials";
 
 export const authOptions: NextAuthOptions = {
+  // Include user.id on session
+  callbacks: {
+    async session({ session, user }) {
+      session.user.id = user.id;
+      return Promise.resolve(session);
+    },
```
balazsorban44, line 11:
```suggestion
    session({ session, user }) {
      session.user.id = user.id;
      return session;
    },
```

template/addons/next-auth/next-auth.d.ts
```diff
@@ -0,0 +1,12 @@
+import { DefaultSession } from "next-auth";
+
+declare module "next-auth" {
+  /**
+   * Returned by `useSession`, `getSession` and received as a prop on the `SessionProvider` React Context
+   */
+  interface Session {
+    user: {
+      id?: string | null;
```
balazsorban44, line 9:
```suggestion
      id: string;
```

Since you are enforcing it, this should always be defined now.


Make sure this `.d.ts` file is `include`d in the `tsconfig.json` to be picked up https://github.com/t3-oss/create-t3-app/blob/main/template/base/tsconfig.json#L19

balazsorban44 (changes requested):
Added my suggestions.

bhatvikrant:
> Added my suggestions.

Hey @balazsorban44 nice to see you here! 
Thanks for the inputs

bhatvikrant:
Have made the changes suggested by @balazsorban44

juliusmarminge (approved):

template/addons/next-auth/api-handler.ts
```diff
@@ -4,6 +4,13 @@ import CredentialsProvider from "next-auth/providers/credentials";
 import { env } from "../../../server/env";
 
 export const authOptions: NextAuthOptions = {
+  // Include user.id on session
+  callbacks: {
+    session({ session, user }) {
+      session.user.id = user.id;
```
balazsorban44, line 10: On second look. this won't do anything. there is no `user` when not using a database adapter. `token` may or may not have `id` in this case.

balazsorban44 (changes requested):

template/addons/next-auth/api-handler.ts
bhatvikrant: Should I remove callbacks then for this case?

TheoBr: @balazsorban44 if we only include this when Prisma is also included, would that make more sense?

TheoBr (approved):
This looks good to me, I'm fine with the session binding being "incorrect" if you don't use a DB persistence (can't imagine many if any of our users would NOT use a DB)

juliusmarminge:
> if we only include this when Prisma is also included, would that make more sense?

> I'm fine with the session binding being "incorrect" if you don't use a DB persistence

I think modifying the installer to have two separate api-handlers would be better? Also not include the next-auth.d.ts if no db is included?

juliusmarminge (changes requested):

juliusmarminge (approved):

template/addons/next-auth/next-auth.d.ts
```diff
@@ -0,0 +1,12 @@
+import { DefaultSession } from "next-auth";
```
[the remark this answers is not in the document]
juliusmarminge: FYI; Made a little demo of why I had the issue here: https://share.cleanshot.com/UQKyOO

When having path-aliasing setup, the `next-auth.d.ts` isn't being picked up.

template/addons/next-auth/api-handler.ts
```diff
@@ -4,6 +4,13 @@ import CredentialsProvider from "next-auth/providers/credentials";
 import { env } from "../../../server/env";
 
 export const authOptions: NextAuthOptions = {
+  // Include user.id on session
```
TheoBr, line 7: Why do we have a separate prisma and non-prisma version? These look identical

template/base/tsconfig.json
```diff
@@ -16,6 +16,12 @@
     "incremental": true,
     "noUncheckedIndexedAccess": true
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "**/*.js"],
+  "include": [
+    "next-env.d.ts",
+    "next-auth.d.ts",
```
TheoBr, line 21: Ooo clever

bhatvikrant: Yup

template/addons/next-auth/next-auth.d.ts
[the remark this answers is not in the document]
bhatvikrant: oh I see

bhatvikrant:
> > > if we only include this when Prisma is also included, would that make more sense?
> > 
> > 
> > > I'm fine with the session binding being "incorrect" if you don't use a DB persistence
> > 
> > 
> > I think modifying the installer to have two separate api-handlers would be better? Also not include the next-auth.d.ts if no db is included?
> 
> On second thought it doesn't really matter since the user-object's type exists whether there is a db or not. This is from a fresh t3 app with just next-auth: CleanShot 2022-07-12 at 19 48 41
> 
> So having an id on there doesn't really matter I guess...
> 
> The above solution would also require two different tsconfigs and now it's really getting annoying...
> 
> But making the `user` (and the id) optional in the `next-auth.ts` I guess would be better to keep the typesafety going?
> 
> ```ts
>   interface Session {
>     user?: {
>       id?: string;
>     } & DefaultSession["user"];
>   }
> ```
> 
> The current setup where they are not optional:
> 
> ```ts
>   interface Session {
>     user: {
>       id: string;
>     } & DefaultSession["user"];
>   }
> ```
> 
> would result in a runtime error if trying to access the id:
> 
> ```ts
> console.log(session?.user.id); // no type-errors, but the user-object is undefined
> ```

I agree, making `user` and `id` optional. This should be better for type-safety

template/addons/next-auth/api-handler.ts
bhatvikrant: `template/addons/next-auth/api-handler.ts` and `template/addons/next-auth/api-handler-prisma.ts` are different. 
The latter sets the Adapter as `PrismaAdapter`

https://github.com/t3-oss/create-t3-app/blob/b2d542a44eed0aa699fe91e52f25f6ac68a0021c/template/addons/next-auth/api-handler-prisma.ts#L12

bhatvikrant:
Hey @juliusmarminge @TheoBr, I've made the suggested changes and the PR already has 2 approvals. 
Whats next to get it merged?

TheoBr (approved):

PR #199, opened 2022-07-15
fix: Increase size of columns for tokens

Decad:
Bumped into this when trying to use google provider and getting `provided value for the column is too long for the column's type`. Noticed this was missing from the schema on the next-auth [docs](https://next-auth.js.org/adapters/prisma)


Decad:
Ah I see why its not here, as by default this is using SQLite which wont have this issue. Not sure what the best thing to do here is, maybe just needs documenting?

Issue #200, opened 2022-07-15
Improve minor config friction when scaffolding prisma


cyremur:
**Problem Description**

As a developer trying out the stack for the first time, I ran into some friction when switching from local `db.sqlite` to a hosted postgres DB.

```
npx prisma db push
Environment variables loaded from .env
Prisma schema loaded from prisma\schema.prisma
Error code: P1012
error: Error validating datasource `db`: the URL must start with the protocol `file:`.
  -->  schema.prisma:11
   | 
10 |     // url      = "file:./db.sqlite"
11 |     url      = env("DATABASE_URL")
```

In `schema.prisma`, I had to adjust `db.provider` from `sqlite` to `postgresql`. 

Additionally, the `.gitignore` file didn't include the `db.sqlite` file created on intended first local setup.

**Describe the solution you'd like**

1.  Include `DATABASE_PROVIDER` setting in the `.env-example` and commented out in `schema.prisma` to make it obvious that it needs to be set as well when using a `DATABASE_URL`.
2. Use partial, feature dependent templates for `.env-example` and `.gitignore / _gitignore` that are concatted into the full file during scaffolding based on selected features (i.e. concat `base/_gitignore` and `prisma/_gitignore`).

Happy to work on this if considered beneficial.

juliusmarminge:
> Include DATABASE_PROVIDER setting in the .env-example and commented out in schema.prisma to make it obvious that it needs to be set as well when using a DATABASE_URL

I dont think the provider is a environment variable so putting it there would make no sense to me.

> Maybe just add a comment instead? Or should this be obvious to people in general?
// url = env("DATABASE_URL") // remember to adjust provider as well

I think if you are switching to another database solution you will have to make changes no matter what, like if you are moving to Planetscale you have to enable referential-integrity and stuff. Another thing you have to keep track of yourself is if using Next-Auth.js you cannot use sqlite because it is not supported. These types of stuff are things that, in my opinion, the user have to change themselves.

c-ehrlich:
I agree that the provider should absolutely not be an environment variable. Prisma is not designed to work like that.

Overall I believe that doing as little custom stuff as possible with Prisma is the way to go - the goal should be that people who have used Prisma before in some other context will not run into any surprises when using ct3a.

juliusmarminge:
> the goal should be that people who have used Prisma before in some other context will not run into any surprises when using ct3a

[...]

PR #201, opened 2022-07-16
feat: add db.sqlite to gitignore

c-ehrlich:
# Add db.sqlite to gitignore

---

_[Short summary/description of story/feature/bugfix]_

As mentioned in https://github.com/t3-oss/create-t3-app/issues/200, the `db.sqlite` file created by the default prisma config was not included in .gitignore. 

I believe it's better to include it - Because we're not automatically creating a 'project initialized commit', it's easy to overlook the db file among all the other files that exist in version history after you initialize the app, and accidentally commit it.

If desired this could be changed to only include that line in the .gitignore if prisma is selected during initialization, but this way is much simpler and personally I think don't think it's an issue to just include it in .gitignore for everyone.

---

💯

### DIFF
--- a/template/base/_gitignore
+++ b/template/base/_gitignore
@@ -9,7 +9,7 @@
 /coverage
 
 # database
-db.sqlite
+/prisma/db.sqlite
 
 # next.js
 /.next/
--- a/template/base/_gitignore
+++ b/template/base/_gitignore
@@ -8,6 +8,9 @@
 # testing
 /coverage
 
+# database
+db.sqlite
+
 # next.js
 /.next/
 /out/


nexxeln:
Thank you

template/base/_gitignore
```diff
@@ -8,6 +8,9 @@
 # testing
 /coverage
 
+# database
+db.sqlite
```
nexxeln, line 12: I think `db.sqlite` exists inside the prisma folder. So please change the path to `/prisma/db.sqlite`

nexxeln (changes requested):

template/base/_gitignore
c-ehrlich: It works this way (see screenshots) but I think you're right that being specific about the path is better. Will change.
 
Screenshot 2022-07-16 at 12 43 09
Screenshot 2022-07-16 at 12 42 55

nexxeln (approved):

nexxeln:
Cool thanks!

PR #202, opened 2022-07-16
refactor: inject curried `runPkgManagerInstall` to package installers

ochicf:
# Inject curried `runPkgManagerInstall` to package installers

---

I noticed that every installer had to call `runPkgManagerInstall` with a lot of configuration, but IMHO the only configuration they should specify is the `packages` list. So, this MR adds a curryable version of `runPkgManagerInstall` that is curried by `createProject` and injected to every installer as an option. This injected function has been curried with all of the necessary configuration, allowing to clear the call in each installer.

Benefits:
- cleaner code of each installer
- adding new options to `runPkgManagerInstall` will not force adding that parameter in each installer, but instead just once in the `createProject.ts` helper where it's curried
- curried options serve only as defaults, so each installer could still override the options if needed

Another small change added here passing the entirety of the options received in `installPackages.ts` to each package installer, since they share the same shape. This will also prevent from having to pass newly added options here.

### DIFF
--- a/src/helpers/createProject.ts
+++ b/src/helpers/createProject.ts
@@ -1,7 +1,7 @@
 import type { PkgInstallerMap } from "../installers/index.js";
 import path from "path";
 import { getUserPkgManager } from "../utils/getUserPkgManager.js";
-import { curryRunPckgManagerInstall } from "../utils/runPkgManagerInstall.js";
+import { curryRunPkgManagerInstall } from "../utils/runPkgManagerInstall.js";
 import { installPackages } from "./installPackages.js";
 import { scaffoldProject } from "./scaffoldProject.js";
 import { selectAppFile, selectIndexFile } from "./selectBoilerplate.js";
@@ -19,7 +19,7 @@ export const createProject = async ({
 }: CreateProjectOptions) => {
   const pkgManager = getUserPkgManager();
   const projectDir = path.resolve(process.cwd(), projectName);
-  const runPkgManagerInstall = curryRunPckgManagerInstall({
+  const runPkgManagerInstall = curryRunPkgManagerInstall({
     projectDir,
     pkgManager,
     devMode: false,
--- a/src/helpers/createProject.ts
+++ b/src/helpers/createProject.ts
@@ -1,6 +1,7 @@
 import type { PkgInstallerMap } from "../installers/index.js";
 import path from "path";
 import { getUserPkgManager } from "../utils/getUserPkgManager.js";
+import { curryRunPckgManagerInstall } from "../utils/runPkgManagerInstall.js";
 import { installPackages } from "./installPackages.js";
 import { scaffoldProject } from "./scaffoldProject.js";
 import { selectAppFile, selectIndexFile } from "./selectBoilerplate.js";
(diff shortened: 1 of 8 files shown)


ochicf:
Not sure about the naming convention with currying functions. I've named `curryRunPkgManagerInstall` because it's the most obvious but maybe you prefer something different?

I've also kept the original `runPkgManagerInstall` export, though it's not used outside of that module anymore. Maybe it causes confusion?

nexxeln:
Oh wow this is much cleaner.

nexxeln (approved):

juliusmarminge (approved):
Looks good.

I actually had a draft to remove the `runPackageManagerInstall` entirely and instead add all the packages to package.json and then run a single `npm/pnpm/yarn install` command at the end, but haven't had time to complete it.

src/installers/prisma.ts
```diff
@@ -4,27 +4,19 @@ import path from "path";
...
+  runPkgManagerInstall,
   pkgManager,
   packages,
   noInstall,
 }) => {
   await runPkgManagerInstall({
-    pkgManager,
-    projectDir,
     packages: ["prisma"],
-    devMode: true,
```
ochicf, line 19: I missed this, so this is not correct.

src/installers/tailwind.ts
```diff
@@ -2,19 +2,13 @@ import type { Installer } from "./index.js";
...
   projectDir,
-  pkgManager,
-  noInstall,
+  runPkgManagerInstall,
 }) => {
   await runPkgManagerInstall({
-    pkgManager,
-    projectDir,
     packages: ["tailwindcss", "postcss", "autoprefixer"],
-    devMode: true,
```
ochicf, line 16: Same as previous comment.

ochicf:
Just noticed in another branch that some installs were actually passing `devMode` and I missed them in the refactor.

src/utils/runPkgManagerInstall.ts
```diff
@@ -50,3 +54,26 @@ export const runPkgManagerInstall = async (opts: {
...
+export type CurryRunPkgManagerInstallOptions = Omit<
+  RunPkgManagerInstallOptions,
+  "packages"
+>;
+
+export type CurriedRunPkgManagerInstallOptions =
+  Partial<CurryRunPkgManagerInstallOptions> &
+    Omit<RunPkgManagerInstallOptions, keyof CurryRunPkgManagerInstallOptions>;
+
+export const curryRunPckgManagerInstall = (
```
juliusmarminge, line 67:
```suggestion
export const curryRunPkgManagerInstall = (
```

src/utils/runPkgManagerInstall.ts
```diff
@@ -50,3 +54,26 @@ export const runPkgManagerInstall = async (opts: {
...
+  baseOptions: CurryRunPkgManagerInstallOptions,
+) => {
+  const curriedRunPckgManagerInstall = async (
```
juliusmarminge, line 70:
```suggestion
  const curriedRunPkgManagerInstall = async (
```

src/utils/runPkgManagerInstall.ts
```diff
@@ -50,3 +54,26 @@ export const runPkgManagerInstall = async (opts: {
...
+    options: CurriedRunPkgManagerInstallOptions,
+  ) =>
+    runPkgManagerInstall({
+      ...baseOptions,
+      ...options,
+    });
+
+  return curriedRunPckgManagerInstall;
```
juliusmarminge, line 78:
```suggestion
  return curriedRunPkgManagerInstall;
```

src/helpers/createProject.ts
```diff
@@ -18,12 +19,30 @@ export const createProject = async ({
 }: CreateProjectOptions) => {
   const pkgManager = getUserPkgManager();
   const projectDir = path.resolve(process.cwd(), projectName);
+  const runPkgManagerInstall = curryRunPckgManagerInstall({
```
juliusmarminge, line 22:
```suggestion
  const runPkgManagerInstall = curryRunPkgManagerInstall({
```

src/helpers/createProject.ts
```diff
@@ -1,6 +1,7 @@
 import type { PkgInstallerMap } from "../installers/index.js";
 import path from "path";
 import { getUserPkgManager } from "../utils/getUserPkgManager.js";
+import { curryRunPckgManagerInstall } from "../utils/runPkgManagerInstall.js";
```
juliusmarminge, line 4:
```suggestion
import { curryRunPkgManagerInstall } from "../utils/runPkgManagerInstall.js";
```

juliusmarminge (changes requested):
Found a typo

ochicf:
Oh sorry just after I pushed I saw that there was the commit suggestion option 🤦‍♂️ Never saw that before, will use next time.

juliusmarminge (approved):

PR #205, opened 2022-07-17
feat: migrate env + next config to esm

danielroe:
# Migrate CJS next config -> ESM

---

This PR renames the default `next.config` file (and its dependencies) to `.mjs`, and converts syntax within those files to ESM. Because ... why not? That's the direction we're moving in anyway, and this enables importing ESM-only packages too.


juliusmarminge:
Mind also implementing the next config typesafe as suggested in #140?

@KATT any downsides of having this as `.mjs`?

PR #208, opened 2022-07-18
feat: generate index page dynamically based on selected packages

### LINKED ISSUE #190: enhancement: index.tsx doesn't include the correct cards
**Describe the bug**
Create t3 app with all possible option will not return the correct page studs seems to have broke in : https://github.com/t3-oss/create-t3-app/commit/c4363821664fe95b4301d46b3e2d34fbeb14952d

**To Reproduce**

1. Run ct3 app
2. Enable all possible feature
3. Run the 3 commands
4. Go to localhost
5. See that only 4 cards on the page stud are showing missing prisma and next auth

**Expected behavior**
Page stud should give out the according cards that were selected when scaffolding (verify all page stud seems like they all use the default component)

**Screenshots**

Sorry i'm on phone je

CarlosGomez-dev:
# Generate index page dynamically based on selected packages

---

This resolves #190, home component is now created dynamically based on the selected packages, this allows us to have a base index and a tailwind index, both are then modified according to the selected packages.

Looked into codemods and jscodeshift, but they seem a bit too complex for this problem, so I resorted to simply replacing strings. I'm open to suggestions for a simpler way to do it.

---


juliusmarminge:
@ochicf How far off are you into getting codemods working? Seems redundant to do this now if we soon are moving to codemods anyways.

ochicf:
> @ochicf How far off are you into getting codemods working? Seems redundant to do this now if we soon are moving to codemods anyways.

@juliusmarminge I was focusing on attempting the reinstalls, so basically going deep instead of going wide. If codemods is chosen as the tool we want to use event for regulars installs (which sounds it does) then I could focus on writing them and refactor the code. Not sure on ETA as I'm pretty new with it and I expect encountering walls, so it won't be in a matter of days.

IMHO, the modifications done in this MR would be better done with a template engine (if not considering codemods) as parsing code with `replace` it's not the safest nor the most maintainable. With that said: if this fixes an issue that is causing problems, then go for it and we will refactor it with codemods anyway.

Issue #195, opened 2022-07-14
Typesafe Client Side Environment Variables


juliusmarminge:
Currently with the new typesafe env-vars, they are all placed under `server`. We should probably have some distinction for client-side env-vars as Theo pointed out on stream when using Pusher for example.

nramkissoon:
Mind if I pick this up as a good first issue?

juliusmarminge:
Implemented in #209

PR #209, opened 2022-07-18
feat: add typesafe client side env variables

nramkissoon:
Add typesafe client side env variables

---

Implements https://github.com/t3-oss/create-t3-app/issues/195 

Separates client and server env schemas. Also added an additional check to ensure client env variable names start with `NEXT_PUBLIC_`.


bhatvikrant (approved):

juliusmarminge:
I would kind of like them to be part of the same env schema when they reach the developer. Example:

```ts
// client envschema
// define and export env schema
```

```ts
// server envschema
// define and export env schema
```

```ts
// env validator
// validate client
// validate server 
// merge and export validated env-vars
```

```ts
// usage
import { env } from "envpath";

const pusherKey = env.NEXT_PUBLIC_PUSHER_KEY; // begins with next_public so we know its client

const dbUrl = env.DATABASE_URL; // doesn't begin with next_public so we know its server
```

Or would this be confusing that you write the schemas in two separate files but import only a single?

nramkissoon:
> I would kind of like them to be part of the same env schema when they reach the developer. Example:
> 
> ```ts
> // client envschema
> // define and export env schema
> ```
> 
> ```ts
> // server envschema
> // define and export env schema
> ```
> 
> ```ts
> // env validator
> // validate client
> // validate server 
> // merge and export validated env-vars
> ```
> 
> ```ts
> // usage
> import { env } from "envpath";
> 
> const pusherKey = env.NEXT_PUBLIC_PUSHER_KEY; // begins with next_public so we know its client
> 
> const dbUrl = env.DATABASE_URL; // doesn't begin with next_public so we know its server
> ```
> 
> Or would this be confusing that you write the schemas in two separate files but import only a single?

I think merging the schemas makes sense for importing env vars server-side because both public and server-only should be available in that case. It would be cumbersome to have to import 2 different env files in a API route file, for example. I can make the changes to do this.

As for client only env vars, turns out you cannot destructure or dynamically access properties in `process.env` client side, so this validation does not work client side.
```js
const _clientEnv = clientEnvSchema.safeParse(process.env);
export const env = _clientEnv
```
```js
//usage
import {env} from "clientEnvPath"

// process.env is empty
```
We can still use this validation to verify env vars at build time and server-side along with the server env vars however, so no concerns there.


The user will either have to resort to accessing client-side env vars explicitly in `process.env` like `process.env.NEXT_PUBLIC_foo` or define and export an env object alongside the schema in order to get autocompletion like so: 
```js
// client-env-schema.mjs
import { z } from "zod";

export const envSchema = z.object({
  // Specify your environment variables schema here
  // Be sure to name your environment variables with the prefix "NEXT_PUBLIC_"
  NEXT_PUBLIC_foo: z.string(),
});
```
```js
// client-env.mjs
import {envSchema} from 'client-env-schema.mjs'
export const env = envSchema.parse({
  NEXT_PUBLIC_DDDD: process.env.DDDD
})
```
```js
//usage
import {env} from 'client-env.mjs'
...
```

I'd like some thoughts on which approach is better here.

juliusmarminge:
My main concern is we will have 4 different files and it will be confusing what goes where, as well as there being 4 files + the actual .env seems kinda "bloated"?

nramkissoon:
> My main concern is we will have 4 different files and it will be confusing what goes where, as well as there being 4 files + the actual .env seems kinda "bloated"?

Spent some time playing with different approaches.

I could not find a way to validate both server and client variables in the same file while being able to import client-side. 
Server variables don't show up in the browser so validation naturally breaks there. Wrapping the server var validation in a 
`if (typeof window === "undefined")` will work on the client, but messes up the types on the server such that server vars will appear as `string | undefined`.

Instead, we can put both schemas in the same file so that the user only needs to edit a single file and reduce any confusion. Then, have a `client-env.mjs` and a `server-env.mjs` for validating and importing client and server env vars, respectively. `server-env.mjs` would have the merged client and server env vars.

So it would be 3 files under an `env` directory.

juliusmarminge:
> Spent some time playing with different approaches.
> 
> I could not find a way to validate both server and client variables in the same file while being able to import client-side. Server variables don't show up in the browser so validation naturally breaks there. Wrapping the server var validation in a `if (typeof window === "undefined")` will work on the client, but messes up the types on the server such that server vars will appear as `string | undefined`.
> 
> Instead, we can put both schemas in the same file so that the user only needs to edit a single file and reduce any confusion. Then, have a `client-env.mjs` and a `server-env.mjs` for validating and importing client and server env vars, respectively. `server-env.mjs` would have the merged client and server env vars.
> 
> So it would be 3 files under an `env` directory.

Yep. Came to the same conclusion. Let's see how that looks

juliusmarminge:
Hotfixing some tRPC stuff so my bad for the format errors and stuff. Sorry, wanted to get the fix out asap

template/base/src/env/server-env.mjs
```diff
@@ -0,0 +1,50 @@
...
+ ) =>
+   Object.entries(errors)
+     .map(([name, value]) => {
+       if (value && "_errors" in value)
+         return `${name}: ${value._errors.join(", ")}\n`;
+     })
+     .filter(Boolean);
+ 
+ const _serverEnv = serverEnvSchema.safeParse(process.env);
+ const _clientEnv = clientEnvSchema.safeParse(process.env);
```
juliusmarminge, line 20: Do we need to validate client here? Can we not just import the env from client instead of the schema?

Validating here and checking status leads to code duplication and error messages needing to be changed in two places.

template/base/src/env/server-env.mjs
```diff
@@ -0,0 +1,50 @@
...
+     console.warn(
+       '❌ Invalid public environment variable name:',
+       key,
+     );
+ 
+     throw new Error("Invalid public environment variable name");
+   }
+ }
+ 
+ export const env = {..._serverEnv.data, ..._clientEnv.data};
```
juliusmarminge, line 49: So this would just be

```suggestion
export const env = {..._serverEnv.data, ...clientEnv }
```

template/base/src/env/server-env.mjs
```diff
@@ -0,0 +1,50 @@
+// @ts-check
+/**
+ * This file is included in `/next.config.mjs` which ensures the app isn't built with invalid env vars.
+ * It has to be a `.mjs`-file to be imported there.
+ */
+ import { serverEnvSchema, clientEnvSchema } from "./env-schema.mjs";
```
juliusmarminge, line 6:
```suggestion
import { serverEnvSchema } from "./env-schema.mjs";
import { env as clientEnv } from "./client-env.mjs";
```

template/base/src/env/server-env.mjs
nramkissoon: Yea that makes more sense, I'll refactor.

template/addons/env/env-auth.mjs
```diff
@@ -1,8 +1,24 @@
...
+});
+
+/**
+ * Next.js client-side environment variables are evaluated at build time, 
+ * so only environment variables actually used will be included.
+ * 
+ * Define your client-side environment variables in this object to
+ * be able to use autocompletion and destructuring in your code. 
+ */
+export const clientEnv = {
```
juliusmarminge, line 22: What does this do? Same for all installers

template/base/src/env/client-env.mjs
```diff
@@ -0,0 +1,36 @@
...
+if (!_clientEnv.success) {
+  console.error(
+    "❌ Invalid environment variables:\n",
+    ...formatErrors(_clientEnv.error.format()),
+  );
+  throw new Error("Invalid environment variables");
+}
+   
+for (let key of Object.keys(_clientEnv.data)) {
+  if (!key.startsWith("NEXT_PUBLIC_")) {
```
juliusmarminge, line 25: Add the "reverse" validation for server so that you can't "accidentally" expose secrets?

template/addons/env/env-auth.mjs
nramkissoon: Mentioned this earlier. This is a workaround for `process.env` not being a standard JS object client-side in Next.js. From the [docs](https://nextjs.org/docs/basic-features/environment-variables):
```
Note: In order to keep server-only secrets safe, environment variables are 
evaluated at build time, so only environment variables actually used will 
be included. This means that process.env is not a standard JavaScript object, 
so you’re not able to use object destructuring. Environment variables must be referenced as e.g. process.env.PUBLISHABLE_KEY, not const { PUBLISHABLE_KEY } = process.env.
```
By explicitly defining the variables in an actual js object you can get autocomplete when importing it in your code among other things and allows for the zod validation to work on the client.

template/base/src/env/client-env.mjs
nramkissoon: Yea, also prevents typos and having to waste time figuring out why your public env vars aren't showing up in the browser.

template/addons/env/env-auth.mjs
juliusmarminge: Hmm that sucks having to "duplicate" the keys but if its necessary it is. I tested some stuff out and this is as "minimal" I could get (compare last commit)...
https://github.com/juliusmarminge/trpc-rqv3

nexxeln:
Is this ready to be merged now?

nramkissoon:
> Is this ready to be merged now?

I think so. @juliusmarminge I've updated my changes to include what you have in your example https://github.com/juliusmarminge/trpc-rqv3

juliusmarminge (changes requested):

juliusmarminge (approved):
Looks good! Thanks for your patience :)

PR #220, opened 2022-07-19
fix: handle pkg with version in noInstall mode

### LINKED ISSUE #219: Getting a "EINVALIDPACKAGENAME" on fresh install
**Describe the bug**
After a fresh t3 install using the latest version, `npm install` fails because of invalid package name "react-query@3.39.2".

**To Reproduce**

1. Run `npx create-t3-app@latest`
2. Answer no to run npm install 
3. Then manually run it

**Expected behavior**
Dependencies should be installed with no failure

**Screenshots**
image

amyisdev:
# Handle adding package with hardcoded version in noInstall mode

---

Fix #219 by extracting the package name using regex before putting the pkg string into dependency object.

### DIFF
--- a/src/utils/runPkgManagerInstall.ts
+++ b/src/utils/runPkgManagerInstall.ts
@@ -34,11 +34,13 @@ export const runPkgManagerInstall = async (
         continue;
       }
 
+      const pkgName = pkg.replace(/^(@?[^@]+)(?:@.+)?$/, "$1");
+
       // Note: We know that pkgJson.[dev]Dependencies exists in the base Next.js template so we don't need to validate it
       if (devMode) {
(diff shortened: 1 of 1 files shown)


src/utils/runPkgManagerInstall.ts
```diff
@@ -34,11 +34,13 @@ export const runPkgManagerInstall = async (
         continue;
       }
 
+      const pkgName = pkg.replace(/^(@?[^@]+)(?:@.+)?$/, "$1");
```
juliusmarminge, line 37: Not good with regex, would this support scoped packages too?

For example if we need to lock tRPC to a version like `@trpc/react@9.26.2`?

amyisdev: yup, it works with scoped package.

here is some examples
image

juliusmarminge:
Thank you for the swift support. Will publish 5.1.2 when I get home!

PR #228, opened 2022-07-19
feat: add vscode ext and settings

juliusmarminge:
# Add recommended vscode extensions and settings

---

Shamefully stolen from @KATT's suggestion in #140.

---

### DIFF
--- a/template/base/.vscode/extensions.json
+++ b/template/base/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+  "recommendations": [
+    "esbenp.prettier-vscode",
+    "dbaeumer.vscode-eslint",
+    "prisma.prisma",
+    "heybourn.headwind"
+  ]
+}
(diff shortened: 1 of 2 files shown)


juliusmarminge:
Putting this up for discussion, both regarding if we include it or not, as well as what extensions we recommend?

nexxeln:
Yeah I'm also on the side of not including it tbh. Not everyone uses vscode and I don't want to add a folder to a template which is just going to be deleted

TheoBr:
Two thoughts:
- This should recommend Tailwind vs code extension
- Prolly shouldn't recommend prettier b/c opinions

I think this fits in w/ using userId in session, it's common enough that it should probably be included and it doesn't hurt the people who choose not to use VS Code

gcavanunez:
Though understandable the take on not including prettier due to getting into opinionated territory, wouldn't the out of the box eslint take from Next, the tailwindcss prettier plugin and the example that the create-t3-app uses prettier, be compelling arguments to include prettier?

PR #236, opened 2022-07-22
feat: check for conflicting files

AlexanderHOtt:
# Check for conflicting files

- [-] I performed a functional test on my final commit

It works fine if given a project name, but I couldn't figure out how to test it with `.` as the name because `pnpm start .` fails with
 ```
> next start "."

'next' is not recognized as an internal or external command,
operable program or batch file.
```
when inside a test folder.

---

_[Short summary/description of story/feature/bugfix]_
Checks if existing files will be overwritten and adds the option to overwrite files if there are conflicts.
```
ℹ project is not empty, but no conflicting files found, continuing...
```
If there are conflicts
```
? Warning: project already exists and has conflicting files:

  project\.env-example
  project\.eslintrc.json
  project\next-env.d.ts
  project\next.config.mjs
  project\package.json
  project\public\favicon.ico
  project\README.md
  project\env\client-env.mjs
  project\env\env-schema.mjs
  project\env\server-env.mjs
  project\pages\index.tsx
  project\pages\_app.tsx
  project\styles\globals.css
  project\tsconfig.json
  project\.gitignore

 How would you like to proceed? (Use arrow keys)
❯ Clear directory
  Abort installation
  Overwrite files (dangerous)
```
---

## Screenshots

_[Screenshots]_
image
image

💯

### DIFF
--- a/src/helpers/scaffoldProject.ts
+++ b/src/helpers/scaffoldProject.ts
@@ -6,6 +6,7 @@ import ora from "ora";
 import { PKG_ROOT } from "../consts.js";
 import { InstallerOptions } from "../installers/index.js";
 import { execa } from "../utils/execAsync.js";
+import { getAllFiles } from "../utils/getAllFiles.js";
 import { logger } from "../utils/logger.js";
 
 // This bootstraps the base Next.js application
@@ -26,18 +27,37 @@ export const scaffoldProject = async ({
   const spinner = ora(`Scaffolding in: ${projectDir}...\n`).start();
 
   if (fs.existsSync(projectDir)) {
-    if (fs.readdirSync(projectDir).length === 0) {
+    const userFilesSet = new Set(getAllFiles(projectDir));
+    const srcFiles = [...getAllFiles(srcDir), ".gitignore"]; // template files
+    const conflictingFiles = srcFiles.filter((element) =>
+      userFilesSet.has(element),
+    );
+
+    if (userFilesSet.size === 0) {
       spinner.info(
         `${chalk.cyan.bold(projectName)} exists but is empty, continuing...\n`,
       );
+    } else if (conflictingFiles.length === 0) {
+      spinner.info(
+        `${chalk.cyan.bold(
+          projectName,
+        )} is not empty, but no conflicting files found, continuing...\n`,
+      );
     } else {
+      let conflictingFilesStr = "";
+      for (const file of conflictingFiles) {
+        conflictingFilesStr += `\n  ${chalk.cyan.bold(
+          projectName,
+        )}${chalk.reset.red(path.sep + file)}`;
+      }
+
       spinner.stopAndPersist();
       const { overwriteDir } = await inquirer.prompt<{ overwriteDir: string }>({
         name: "overwriteDir",
         type: "list",
         message: `${chalk.redBright.bold("Warning:")} ${chalk.cyan.bold(
           projectName,
-        )} already exists and isn't empty. How would you like to proceed?`,
+        )} already exists and has conflicting files:\n${conflictingFilesStr}\n\n How would you like to proceed?`,
         choices: [
           { name: "Clear directory", value: "clear", short: "Clear" },
           { name: "Abort installation", value: "abort", short: "Abort" },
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by AlexanderHOtt** [comment]
Fix for #230

File: src/helpers/scaffoldProject.ts [hunk 1]
```diff
@@ -32,24 +32,35 @@ export const scaffoldProject = async ({
...
+        choices: [
+          { name: "Clear directory", value: "clear", short: "Clear" },
+          { name: "Abort installation", value: "abort", short: "Abort" },
+          {
+            name: "Overwrite files (dangerous)",
+            value: "overwrite",
+            short: "Overwrite",
+          },
+        ],
+        default: "clear",
```
juliusmarminge, line 70: the default should be `abort` so that people don't accidentally press enter and get all their files removed.

```suggestion
        default: "abort",
```

src/helpers/scaffoldProject.ts
```diff
@@ -26,30 +27,62 @@ export const scaffoldProject = async ({
...
-          name: "overwriteDir",
-          type: "confirm",
-          message: `${chalk.redBright.bold("Warning:")} ${chalk.cyan.bold(
-            projectName,
-          )} already exists and isn't empty. Do you want to overwrite it?`,
-          default: false,
-        },
-      );
-      if (!overwriteDir) {
+      const { overwriteDir } = await inquirer.prompt<{ overwriteDir: string }>({
```
juliusmarminge, line 55: Making this string type safe will help preventing typos if modified later.

```suggestion
      const { overwriteDir } = await inquirer.prompt<{ overwriteDir: "clear" | "abort" | "overwrite" }>({
```

src/utils/getAllFiles.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Find all files inside a dir, recursively.
+ * https://gist.github.com/kethinov/6658166?permalink_comment_id=3379782#gistcomment-3379782
+ * https://gist.github.com/kethinov/6658166?permalink_comment_id=2389484#gistcomment-2389484
+ * @function getAllFiles
+ * @param {string} dir Directory path string.
+ * @param {string} base Base path string.
+ * @return {string[]} Array with all the relative paths that are inside the directory.
+ */
```
juliusmarminge, line 12: Not a fan of jsdoc and we don't use it anywhere else. Using typescript is clear enough and inlined. Leave the reference link if it is taken externally.

juliusmarminge (commented):
Not available to test this today so will have another look over the weekend/next week. 

I can see some edge-cases on the top of my head. E.g. user is in a git repo which is not caught but this logic. We should probably extend the initializeGit to handle this. @JacobMGEvans sent me this a while back that may be helpful: [cloudflare/wrangler2 - git-client.ts](https://github.com/cloudflare/wrangler2/blob/main/packages/wrangler/src/git-client.ts)

AlexanderHOtt:
> user is in a git repo which is not caught but this logic
The `git init` should fail but that won't overwrite any files.

src/helpers/scaffoldProject.ts
```diff
@@ -26,30 +27,64 @@ export const scaffoldProject = async ({
   const spinner = ora(`Scaffolding in: ${projectDir}...\n`).start();
 
   if (fs.existsSync(projectDir)) {
-    if (fs.readdirSync(projectDir).length === 0) {
+    const userFilesSet = new Set(getAllFiles(projectDir));
+    const srcFiles = [...getAllFiles(srcDir), ".gitignore"]; // template files
```
juliusmarminge, line 31: this means we are just checking against the base app without any addons. If there is files that would conflict with any of these other ones these would not be matched and potentially removed by accident

src/helpers/scaffoldProject.ts
```diff
@@ -26,30 +27,64 @@ export const scaffoldProject = async ({
...
+    const conflictingFiles = srcFiles.filter((element) =>
```
juliusmarminge, line 32: userFilesSet might have e.g. `prisma/schema.prisma` which would not be in srcFiles and therefore not in conflictingFiles. Continuing to scaffold and letting the user pick prisma would override the existing file which is not what we want.

juliusmarminge (changes requested):
Sorry for the delay on this review...

I believe there are a few gotchas which have to be addressed before we can merge this feature

juliusmarminge:
Closed as inactive

PR #238, opened 2022-07-22
feat: improved logo typography

c-ehrlich:
# Improve logo typography

---

This PR improves the typography of the A's in the ascii logo. It also removes a dependency.

There are two obvious ways to implement this - make the ascii text a const, or read it from an external text file. I decided to make it a const, which is simpler but makes the ascii harder to read in code due to having to escape backslashes. Let me know if you prefer the other solution.

---

## Screenshots
top: before, bottom: after

Screenshot 2022-07-22 at 14 02 08

💯

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
   "dependencies": {
     "chalk": "5.0.1",
     "commander": "^9.3.0",
-    "figlet": "^1.5.2",
     "fs-extra": "^10.1.0",
     "gradient-string": "^2.0.1",
     "inquirer": "^9.0.0",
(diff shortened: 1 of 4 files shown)


juliusmarminge:
> top: before, bottom: after

That comparison aint fair, compare it to the beta where the offset is fixed ;)

Anyway, the A looks good. Bit messy to read in sourcecode like you said but idk if this matters?

nexxeln:
> Anyway, the A looks good. Bit messy to read in sourcecode like you said but idk if this matters?

doesn't matter

PR #240, opened 2022-07-22
feat: add JSDoc type annotation to `clientEnv` ensuring variables are set

Duckapple:
# Added JSDoc type annotation to ensure variables are set

---

Zod already ensures that the build fails when environment variables declared in the `clientEnvSchema` are not represented in `clientEnv`.
This PR seeks to improve that experience, making the errors show up right in the editor.

It is achieved via a JSDoc type declaration, which loosens the definition of the zod object, resulting in errors whenever a key in the schema object isn't defined in the `clientEnv` object. This loosening is needed to ensure that environment variables (of type `string | undefined`) can be passed to the object.

I also ran Prettier on the .mjs files, partly due to "format on save" with Prettier in my editor. 

---

## Screenshots

A screenshot of VSCode showing an error when a client-side environment variable isn't declared
A screenshot of VSCode showing an error when a client-side environment variable isn't declared.

A comparison shot of how things were before this change
A comparison shot of how things were before this change. It wouldn't build (I hope), but don't you want this info before build-time?


template/base/src/env/env-schema.mjs
```diff
@@ -11,12 +11,13 @@ export const clientEnvSchema = z.object({
...
 /**
- * Next.js client-side environment variables are evaluated at build time, 
+ * Next.js client-side environment variables are evaluated at build time,
  * so only environment variables actually used will be included.
- * 
+ *
  * Define your client-side environment variables in this object to
- * be able to use autocompletion and destructuring in your code. 
+ * be able to use autocompletion and destructuring in your code.
+ * @type {{ [k in keyof z.infer<typeof clientEnvSchema>]: z.infer<typeof clientEnvSchema>[k] | undefined }}
```
Duckapple, line 19: This is the relevant change, which is also present in the other versions of `env-schema`

nexxeln (approved):

juliusmarminge (commented):
> I also ran Prettier on the .mjs files, partly due to "format on save" with Prettier in my editor.

Nice, Reminder that we should probably add .mjs files when running prettier

https://github.com/t3-oss/create-t3-app/blob/126aaaecbb2e58c04c335032fd79d193e5227b06/package.json#L35-L36

> This loosening is needed to ensure that environment variables (of type string | undefined) can be passed to the object.

Makes sense since `process.env.X` might not exist. That's alright i think. Still gives you the error that you have missed to put something in the "destruction-object".

> but don't you want this info before build-time?

I like this a lot. Not at my computer right now so cant test it.

juliusmarminge:
When #245 is merged you can update the branch and I'll merge this afterwards.

nexxeln:
Conflicts 😩

juliusmarminge (approved):

Duckapple:
I have updated the PR to match the new clientEnv files, and tested it locally

juliusmarminge:
Great thanks :)

PR #242, opened 2022-07-23
feat: updating tailwind and postcss config's to use .cjs

JohnDaly:
## Updating `postcss.config` and `tailwind.config` files to use `.cjs` extension.

- [X] I reviewed linter warnings + errors, resolved formatting, types and other issues related to my work

---

### Why?
Other configuration files in this project use the `.mjs` extension for files that use ESM and `.cjs` for CommonJS. 

This change also allows users of `create-t3-app` to add [`"type": "module"` to their `package.json`](https://nodejs.org/api/packages.html#type) without the need to change the extensions on these config files.

### Why not make these files .mjs?
Internally, Next.js and Tailwind both use `require` to load the configs (Next.js loads `postcss.config` and Tailwind loads `tailwind.config`). Since CommonJS can't easily `require` ESM, we will have to stick with the CommonJS format for these config files.


nexxeln:
What benefits does this give over the normal config files?

juliusmarminge:
> What benefits does this give over the normal config files?

Just had a thought, if we explicitly name all the js stuff with mjs or cjs, it's more clear that they are js files on purpose and probably should not be changed. Just me?

PR #245, opened 2022-07-23
fix: clarify some comments and rename some files in env

juliusmarminge:
# Clarifications regarding env

---

We get a lot of questions regarding the typesafe env so this PR tries to clarify some of this, as well as making some personal enhancements.

- I renamed some files, the previous paths, e.g. `src/env/env-schema.mjs` had redundant double-env, so renamed to `src/env/schema.mjs`
- Made comments for the env-stuff ts-doc style, so they appear on hover.
- Clarified (imo) why you have to define your client-env twice.

---

## Additional notes

Should we prompt for typesafe env? Currently, the base project includes [the files by default](https://github.com/t3-oss/create-t3-app/tree/main/template/base/src/env) stuff for this and is not part of any installer. We had #196 that makes the env installer only go if either Prisma or NextAuth.js is selected, but the base stuff is still present regardless.


nexxeln (commented):
Should we also add these things in the generated README?

juliusmarminge:
> Should we also add these things in the generated README?

Yeah the t3-app/README can probably include some stuff for getting started and more "next steps" stuff

juliusmarminge:
This is blocking #240 so merge when you looked it over. 

The README stuff should probably be a separate PR anyways.

template/base/src/env/schema.mjs
```diff
@@ -0,0 +1,28 @@
...
+export const clientSchema = z.object({
+  // NEXT_PUBLIC_BAR: z.string(),
+});
+
+/**
+ * You can't destruct `process.env` as a regular object, so you have to do
+ * it manually here. This is because Next.js evaluates this at build time,
+ * and only used environment variables are included in the build.
+ */
+export const clientEnv = {
```
corydeppen, line 26: Since this isn't schema config, would it be appropriate to include it in `client.mjs` instead?

juliusmarminge: Then you would have to change between files which seems messy. As it is now you dont even have to look st client/server.mjs. Atleast i prefer that

JacobMGEvans (approved):
I think these are great quality improvements!

k4u5hik:
This is going to be very helpful for newbies like myself.

PR #255, opened 2022-07-26
fix: missing next-auth types

zomars:
# Fixes missing type in `next-auth` module augmentation.

- [x] I reviewed linter warnings + errors, resolved formatting, types and other issues related to my work
- [x] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] I performed a functional test on my final commit

---

fixes #254 Fixes missing type in `next-auth` module augmentation.

---

## Screenshots

_[Screenshots]_

image

Issue #254, opened 2022-07-26
Adding "paths" and "baseUrl" to `tsconfig.json` breaks module augmentations


zomars:
**Describe the bug**
It seems like the declared `next-auth.d.ts` is overriding the initial `next-auth` types instead of augmenting them.

**To Reproduce**

1. `pnpm dlx create-t3-app@next`
2. Select all packages
3. See `src/pages/api/auth/[...nextauth].ts` file

**Expected behavior**

Having default `next-auth` types PLUS the augmented ones.

**Screenshots**

<img width="777" alt="image" src="https://user-images.githubusercontent.com/3504472/180909657-519df48f-6fcf-4ee5-b7c0-46ff7b769751.png">

<img width="648" alt="image" src="https://user-images.githubusercontent.com/3504472/180909681-3f245988-9a14-4dbc-b4f5-74e75cb33d07.png">

nexxeln:
Unable to reproduce, works fine for me.
![user has User type](https://us-east-1.tixte.net/uploads/nexxel.needs.rest/Code_JhXLUy5Krx.png)

zomars:
@nexxeln it may be out of scope, but it seems this happens as soon as you add path aliases (took them from this project btw):

<img width="387" alt="image" src="https://user-images.githubusercontent.com/3504472/180913329-1cbb182d-0161-491d-9e79-5b6c9603fcd0.png">

nexxeln:
Yup thats the issue, I added path aliases and am now getting a type error.

zomars:
It seems like even with #255 adding both "paths" and "baseUrl" to `tsconfig.json` keeps messing up with the module augmentations.

nexxeln:
> It seems like even with #255 adding both "paths" and "baseUrl" to `tsconfig.json` keeps messing up with the module augmentations.

Yeah I'm not sure how to fix this, I'll open an issue in `next-auth`

kroucher:
Some info for troubleshooting, when the `next-auth.d.ts` file is moved into any subdirectory, the types start working again.

![image](https://user-images.githubusercontent.com/99768034/180942900-1ab5377b-f1da-48de-819e-88f0b2708d9c.png)

Not sure if having a 'types' folder by default is too opinionated, but either way, the issue lies with the `next-auth.d.ts` file being in the root directory.

JohnDaly:
[...]

PR #257, opened 2022-07-26
revert: extend regular next-auth module

juliusmarminge:
# Revert changes made in #255

---

Reverting changes to the `next-auth.d.ts` change made in #255 and discussed in #254 and https://github.com/nextauthjs/next-auth/issues/5024

I think we should probably follow the [official docs](https://next-auth.js.org/getting-started/typescript#module-augmentation) as long as it is possible and augmenting a nested module may cause confusion when users compare the two.

I think the proper way to fix the issue that the types are not recognized when adding path aliasing is to move the typedefinition file into some subdirectory of the baseUrl (such as `src/types`).

I also think this more aligns with Balazs' response (https://github.com/nextauthjs/next-auth/issues/5024#issuecomment-1195192426)?

Alternatively we could move the file into `src/types` by default, but not sure if we really need that...

Feel free to convince me otherwise, but this is the approach I usually take and advice others to take too

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/src/installers/next-auth.ts
+++ b/src/installers/next-auth.ts
@@ -33,7 +33,10 @@ export const nextAuthInstaller: Installer = async ({
   );
 
   const nextAuthDefinitionSrc = path.join(nextAuthAssetDir, "next-auth.d.ts");
-  const nextAuthDefinitionDest = path.join(projectDir, "next-auth.d.ts");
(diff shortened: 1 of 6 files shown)


nexxeln (commented):
I'm down to put it in `src/types`

PR #259, opened 2022-07-27
refactor: Perf gains

juliusmarminge:
# Improve CLI Performance

---

Scaffolding a full T3 app can take time. This aims to reduce that time.

For initialization stuff like we do, fs.copySync() is faster than fs.copy() etc. (from what I've read you should only use async version on server stuff when needing cpu power for other stuff)

So I have refactored to take advantage of this. By using sync over async I decreased scaffolding times on my machine from 32 to 28 seconds for a full app.

I also refactored so that we only run a single `npm install` command at the end, instead of running 5-10 per app. 

Might be more things we can do to further decrease scaffolding time?

---

## Screenshots

_[Screenshots]_

💯


juliusmarminge:
Oh shit used old next branch...

PR #260, opened 2022-07-27
refactor: perf gains

juliusmarminge:
# Improve CLI Performance

---

Scaffolding a full T3 app can take time. This aims to reduce that time.

For initialization stuff like we do, fs.copySync() is faster than fs.copy() etc. (from what I've read you should only use async version on server stuff when needing cpu power for other stuff)

So I have refactored to take advantage of this. By using sync over async I decreased scaffolding times on my machine from 32 to 28 seconds for a full app.

I also refactored so that we only run a single npm install command at the end, instead of running 5-10 per app.

Might be more things we can do to further decrease scaffolding time?
---

## Screenshots

_[Screenshots]_

💯


juliusmarminge:
Correct branch this time

src/installers/prisma.ts
```diff
@@ -41,21 +37,10 @@ export const prismaInstaller: Installer = async ({
...
-    }),
-  ]);
-
-  // only generate client if we have installed the dependencies
-  if (!noInstall) {
-    const generateClientCmd =
-      pkgManager === "npm"
-        ? "npx prisma generate"
-        : `${pkgManager} prisma generate`;
-    await execa(generateClientCmd, { cwd: projectDir });
```
juliusmarminge, line 59: this will be executed in post install script anyway so no need to run it here too

src/installers/index.ts
```diff
@@ -15,9 +15,36 @@ export const availablePackages = [
...
+  postcss: "^8.4.14",
+
+  // tRPC
+  "@trpc/client": "10.0.0-alpha.41",
+  "@trpc/server": "10.0.0-alpha.41",
+  "@trpc/react": "10.0.0-alpha.41",
+  "@trpc/next": "10.0.0-alpha.41",
+  "react-query": "^3.39.2",
+  superjson: "^1.9.1",
+} as const;
```
nexxeln, line 45: Love this

nexxeln (commented):
So basically most of the async stuff is sync now, so its faster?

juliusmarminge:
> So basically most of the async stuff is sync now, so its faster?

Yes changing "fs.async" -> "fs.sync" made some difference (about 5 seconds I think).

Then by just adding the dependency to package.json and running a single `npm install` command at the end also does a lot.

nexxeln:
> Then by just adding the dependency to package.json and running a single npm install command at the end also does a lot.

Makes sense, so is this ready to be merged? We can test it out in beta.

juliusmarminge:
> > Then by just adding the dependency to package.json and running a single npm install command at the end also does a lot.
> 
> 
> 
> Makes sense, so is this ready to be merged? We can test it out in beta.

I targetted next but maybe we can publish beta from there?

nexxeln:
Sure

Issue #269, opened 2022-07-31
Ask for project name when creating project on current dir "yarn create t3-app ."


ibrahimyaacob92:
**Describe the bug**

**To Reproduce**

1. Run this '...'
2. See error

**Expected behavior**

**Screenshots**

juliusmarminge:
Could you provide some more information please?

c-ehrlich:
I believe what the OP is referring to is that on `npx create-t3-app .`, the `name` in package.json also gets set to `.`.

A more sensible default to use in this case would be the name of the current directory - that's what create-next-app does for example. I personally don't think the extra step of asking the user to input a name is worth it, as you will need to edit large parts of package.json either way if you're building a large/serious app.

I'd be happy to implement using the folder name if you think it's a good idea.

juliusmarminge:
> I believe what the OP is referring to is that on `npx create-t3-app .`, the `name` in package.json also gets set to `.`.
> 
> 
> 
> A more sensible default to use in this case would be the name of the current directory - that's what create-next-app does for example. I personally don't think the extra step of asking the user to input a name is worth it, as you will need to edit large parts of package.json either way if you're building a large/serious app.
> 
> 
> 
> I'd be happy to implement using the folder name if you think it's a good idea.

Sure, sounds sensible.

To the OP, next time when opening an issue, please fill in the form with information and not just submit the preset.

ibrahimyaacob92:
sorry bad habit of mine, but thanks! love to see you guys making this project better and better

PR #273, opened 2022-08-01
feat: set appName to directory on 'npx create-t3-app .'

c-ehrlich:
# set appName to directory on 'npx create-t3-app .'

---

resolves https://github.com/t3-oss/create-t3-app/issues/269

Before, when running create-t3-app with '.' as the name argument (ie to init inside the current folder), the `name` in package.json was also set to `.`. With this pr, the folder name is used instead in that case. This is the same behaviour as many other tools such as create-next-app.

💯


juliusmarminge:
Thanks

Issue #271, opened 2022-08-01
Next-auth fails with OAuthAccountNotLinked on second signin with Google OAuth + MySQL


cyremur:
**Describe the bug**
When using Google OAuth with MySQL, the next-auth Account creation silently fails.

**To Reproduce**

1. Setup create-t3-app with next-auth with Google OAuth and MySQL DB.
2. Push default schema `npx prisma db push`
3. Sign in via `/api/auth/signin`
4. Sign out via `/api/auth/signout`
5. Sign in via `/api/auth/signin`

**Expected behavior**
User is signed back into their account.

**Actual behavior**
Next-auth throws an error.
OAuthAccountNotLinked
![image](https://user-images.githubusercontent.com/5361181/182052916-7d50015f-bf8c-4459-9f67-154c9b686b08.png)

**Debugging results**
When using OAuth, next-auth creates an entry in both the User and the Account table. The Account table is used to verify repeated OAuth sign ins.

However, it turns out that some OAuth tokens (in my case from Google) are very long. The default DB Mapping for the type String is Text in PostgreSQL and SQLite but notably VARCHAR(191) in MySQL. This results in the INSERT statement into the Account table silently failing in the prisma adapter and a User being created without accompanying Account. That User has one and only one session where everything works as expected, and on second sign-in the missing entry in Account table will make the User inaccessible.
https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#string

Note: I used PlanetScale because it was referenced in https://init.tips/others and it uses MySQL. IMO, it would be nice if create-t3-app fully supported that one out of the box with next-auth and prisma.

**Solution**

I fixed this issue by annotating the two relevant token columns `access_token` and `id_token` with an explicit `@db.text`:
```
// Necessary for Next auth
model Account {
    id                       String  @id @default(cuid())
    userId                   String
    type                     String
    provider                 String
    providerAccountId        String
    refresh_token            String?
    access_token             String? @db.Text
    expires_at               Int?
    token_type               String?
    scope                    String?
    id_token                 String? @db.Text
    session_state            String?
    user                     User    @relation(fields: [userId], references: [id], onDelete: Cascade)

    @@unique([provider, providerAccountId])
}
```

This schema should work fine for PostgreSQL, MySQL, and Microsoft SQL Server. It might however cause issues for SQLite, CockroachDB and MongoDB. I only tested with MySQL and SQLite: MySQL this works, SQLite fails on `npx prisma db push`.

[...]

PR #275, opened 2022-08-02
fix(prisma): make db changes for next-auth on mysql obvious

cyremur:
# auth-schema migration hints

---

As discussed in #271, switching to MySQL with the current `auth-schema.prisma` can result in unexpected login behavior due to different default `String?` mappings in prisma. This change allows the user to quickstart with `sqlite` but makes it obvious that changes to the schema need to be made when switchting to `postgresql`, `mysql` or `sqlserver`.


juliusmarminge (approved):
Thanks 🙏

Issue #281, opened 2022-08-05
Make ctx.session.user.id non-nullable


c-ehrlich:
### **Is your feature request related to a problem? Please describe.**
Currently the session object defines user as:
```
(property) Session.user?: {
    id?: string | undefined;
} & {
    name?: string | null | undefined;
    email?: string | null | undefined;
    image?: string | null | undefined;
}
```
As far as I understand, there is no way to have a session.user without also having session.user.id, as id is non-nullable on the user in the DB schema. Making the id just `id: string` would allow writing fewer guard clauses. 

### **Describe the solution you'd like**
Perhaps the best possibility would be to change how `Session` is declared to:
```ts
interface Session {
    user?: {
      id: string;
    } & DefaultSession['user'];
  }
```
User remains nullable, but if user exists then we assume they have an id. 

Comments from @TheoBr :
<img width="635" alt="image" src="https://user-images.githubusercontent.com/8353666/183128854-544ecffd-e722-4dc0-a958-2fd79ce6aaa2.png">

I don't know enough about auth and possible issues/edge cases to feel comfortable just opening a PR about it, would prefer to get some comments first.

### **Describe some other solutions you have considered**
Doing the following before returning the user object doesn't work - it still remains typed as `id?: string | undefined`
```ts
if (!ctx.session || !ctx.session.user || !ctx.session.user.id) {
    throw new TRPCError({ code: 'UNAUTHORIZED' });
  }
```

Running the user object through a zod schema works, but seems like a terrible idea.

There are probably other possible solutions. Open to any other suggestions.

gcavanunez:
The proposed solutions looks really good! 
I think in the context of a session, we would either have no User or when we have a User, the id would always have data.

TheoBr:
I think this makes a lot of sense

PR #282, opened 2022-08-05
feat: make session.user.id non-nullable

c-ehrlich:
# feat: make session.user.id non-nullable

---

Resolves https://github.com/t3-oss/create-t3-app/issues/281

Makes `session.user.id` non-nullable, resulting in fewer guard clauses without . For further thoughts see https://github.com/t3-oss/create-t3-app/issues/281.

This changes behaviour that was discussed extensively in https://github.com/t3-oss/create-t3-app/pull/180#issuecomment-1182760580, however in that PR the only options discussed were both user and user.id nullable, or both user and user.id non nullable, not user nullable but user.id non nullable.

Open for comments though :)

---

## Screenshots

image

💯

### DIFF
--- a/template/addons/next-auth/next-auth.d.ts
+++ b/template/addons/next-auth/next-auth.d.ts
@@ -6,7 +6,7 @@ declare module "next-auth" {
    */
   interface Session {
     user?: {
-      id?: string;
+      id: string;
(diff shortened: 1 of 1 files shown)


juliusmarminge:
Looks good can you also make this change on next?

PR #284, opened 2022-08-06
feat: make session.user.id nullable (next)

c-ehrlich:
applying https://github.com/t3-oss/create-t3-app/pull/282 to next


juliusmarminge:
@c-ehrlich You applied the inverse ;)

Issue #263, opened 2022-07-28
Property 'error' does not exist on type 'SafeParseReturnType<{}, {}>'.   Property 'error' does not exist on type 'SafeParseSuccess<{}>'.


CoenWarmer:
**Describe the bug**
I started a new project with Prisma, NextAuth, tRPC. Adjusted `.env` and `env/schema.mjs` to use the minimal variables for my app at this point.

In Dev mode everything works great, but on build `env/server.mjs` throws a linting error: 

```Property 'error' does not exist on type 'SafeParseReturnType<{}, {}>'.   Property 'error' does not exist on type 'SafeParseSuccess<{}>'.```

**To Reproduce**

1. Scaffold a new project
2. Try to `yarn build`

**Expected behavior**
This type error does not happen.

**Screenshots**
<img width="1030" alt="Screenshot 2022-07-28 at 13 14 24" src="https://user-images.githubusercontent.com/535564/181492316-2a2a07d9-8e1f-45c2-9104-26a5b6f68872.png">

<img width="516" alt="Screenshot 2022-07-28 at 13 15 48" src="https://user-images.githubusercontent.com/535564/181492552-194c6a57-2c7d-4357-bac8-1607d9789dfa.png">

<img width="509" alt="Screenshot 2022-07-28 at 13 16 01" src="https://user-images.githubusercontent.com/535564/181492571-53d92681-3a51-47e7-b376-ca5f4619e67c.png">

Any idea?

**Comment by juliusmarminge** [comment]
Sorry for the delay.

Unfortunatly, I am unable to reproduce this behavior. I have created multiple apps both on latest, beta and next and with different package setups and they all build fine for me. Do you have any reproduction repo to share?

**Comment by khalil-omer** [comment]
@juliusmarminge I just reproduced this when trying to deploy to Vercel. Please take a look at the PR I opened at #293. To be clear, this PR is just to demonstrate what I think the fix is -- I'm not super familiar with this repo so not saying that the PR is ready to merge.

**Comment by juliusmarminge** [comment]
@CoenWarmer @khalil-omer Can you please try reproducing now? Its published so make sure you use the latest tag.
```sh
pnpm dlx create-t3-app@latest
```

CoenWarmer:
Apologies @juliusmarminge! Am currently on holiday and not near a computer. I trust the issue is fixed and if it isn't I'll reopen. Thanks for looking into this!

PR #293, opened 2022-08-08
fix: typeerror in env validation

khalil-omer:
# Fix typeerror in env validation 

- [X] I reviewed linter warnings + errors, resolved formatting, types and other issues related to my work

---

This is a POC fix to demonstrate the problem in issue #263

### DIFF
--- a/template/base/src/env/client.mjs
+++ b/template/base/src/env/client.mjs
@@ -14,7 +14,7 @@ export const formatErrors = (
     })
     .filter(Boolean);
 
-if (!_clientEnv.success) {
+if (_clientEnv.success === false) {
   console.error(
     "❌ Invalid environment variables:\n",
     ...formatErrors(_clientEnv.error.format()),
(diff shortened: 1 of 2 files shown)


template/base/src/env/client.mjs
```diff
@@ -14,7 +14,7 @@ export const formatErrors = (
     })
     .filter(Boolean);
 
-if (!_clientEnv.success) {
+if (_clientEnv.success === false) {
```
juliusmarminge, line 17: How come this would help?

khalil-omer: Honestly I'm not sure @juliusmarminge. The bang operator can be strange sometimes, but in this case `!_clientEnv.success` _should_ behave the same way as `_clientEnv.success === false`. However, it clearly exhibits unexpected behavior given that at least two of us have gotten this error now, which the change I am proposing would fix. 

If you look at the types for Zod, you can see that `success` is a required property, so we are not changing the logic by switching from `!` to `=`. 

https://github.com/colinhacks/zod/blob/master/src/types.ts#L128

juliusmarminge: Yea I guess we could try. Would you like to submit this change to the next-branch as well?

PR #303, opened 2022-08-10
fix: parse path returned from process.cwd() before using it

### LINKED ISSUE #302: on Windows, path returned by `process.cwd()` breaks the path parser
**Describe the bug**

https://github.com/t3-oss/create-t3-app/blob/f060d59a4e53a6932c5dd9620a20efa7d1c96241/src/utils/parseNameAndPath.ts#L16-L19

On windows, process.cwd() returns something like: `D:\\path\\to\\project`, which breaks the above code since it uses `\\` instead of `/` to separate efferent segments of path.
Consequently, project generated by `create t3 app` will use an invalid app name:
image

**Steps To Reproduce**

1. (on any windows computer i guess) `npx create-t3-app .`
2. see `package.json`

**Possible Solution**

Use nodejs's builtin path module to parse the path returned by process.cwd():
```ts
  // If the user ran `npx create-t3-app .` or similar, the appName should be the current directory
  if (appName === ".") {
    const parsedCwd = pathModule.resolve(process.cwd());
    appName = pathModule.basename(parsedCwd);
  }
```

BWsix:
# Fix package name error mentioned in #302 

---

### Describe the bug

https://github.com/t3-oss/create-t3-app/blob/f060d59a4e53a6932c5dd9620a20efa7d1c96241/src/utils/parseNameAndPath.ts#L16-L19

On windows, process.cwd() returns something like: `D:\\path\\to\\project`, which breaks the above code since it uses `\\` instead of `/` to separate efferent segments of path.
Consequently, project generated by `create t3 app` will use an invalid app name:
image

### Changes

This PR uses nodejs's builtin path module to parse the path returned by process.cwd() before consuming it, ensures windows users get correct package name.

### DIFF
--- a/src/utils/parseNameAndPath.ts
+++ b/src/utils/parseNameAndPath.ts
@@ -1,3 +1,5 @@
+import pathModule from "path";
+
 /**
  *  Parses the appName and its path from the user input.
  * Returns an array of [appName, path] where appName is the name put in the package.json and
@@ -15,7 +17,8 @@ export const parseNameAndPath = (input: string) => {
 
   // If the user ran `npx create-t3-app .` or similar, the appName should be the current directory
   if (appName === ".") {
-    appName = process.cwd().split("/").pop();
+    const parsedCwd = pathModule.resolve(process.cwd());
+    appName = pathModule.basename(parsedCwd);
   }
 
   // If the first part is a @, it's a scoped package


juliusmarminge:
I am not on Windows so cannot confirm n'or deny if this is working. How bout you @nexxeln?

nexxeln:
> I am not on Windows so cannot confirm n'or deny if this is working. How bout you @nexxeln?

Seems to be working on my end. Can you also confirm it works on mac?

juliusmarminge:
> > I am not on Windows so cannot confirm n'or deny if this is working. How bout you @nexxeln?
> 
> Seems to be working on my end. Can you also confirm it works on mac?

The name in package.json is correctly parsed but the logs are a bit awkward (although I assume they are for the current setup as well) 
CleanShot 2022-08-11 at 21 27 56@2x

PR #305, opened 2022-08-10
feat: Include ct3a version as metadata in generated app

### LINKED ISSUE #298: Include ct3a version as metadata in generated app
**Is your feature request related to a problem? Please describe.**
Manny bug-reports are cause people don't use the latest version, I.e only running `npx create-t3-app`. Having a way to tell which version was used may help debugging the issue

**Describe the solution you'd like**
Perhaps include a field in the package.json with what version was used to scaffold? It would "litter" the project and leave a trace which it currently doesn't have.

**Additional**
Alternative would be just leaving things as is, and maybe adding in a prerequisite of filing an issue is checking if the issue persists using latest version.

c-ehrlich:
# Include ct3a version as metadata in generated app

---

Resolves #298

Let me know if you want:
- the type moved to src/index.ts (I feel like it putting it next to the other types makes sense, but don't care strongly about it)
- you want the same on the next branch

💯

### DIFF
--- a/src/installers/index.ts
+++ b/src/installers/index.ts
@@ -1,3 +1,4 @@
+import type { PackageJson } from "type-fest";
 import type { PackageManager } from "~/utils/getUserPkgManager.js";
 import type { CurriedRunPkgManagerInstallOptions } from "~/utils/runPkgManagerInstall.js";
 import { envVariablesInstaller as envVariablesInstaller } from "~/installers/envVars.js";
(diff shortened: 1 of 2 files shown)


nexxeln (approved):
Nice

src/installers/index.ts
```diff
@@ -62,3 +63,9 @@ export const buildPkgInstallerMap = (
     installer: envVariablesInstaller,
   },
 });
+
+export type CT3APackageJSON = PackageJson & {
```
juliusmarminge, line 67: yea I think this should probably be moved. When I skim through this file I would wonder `wtf is this doing?`

juliusmarminge (commented):
Looks good, I do think the type shouldn't be under installer though

src/installers/index.ts
nexxeln: Yeah true

juliusmarminge:
Thanks. Mirror on next please 🙏

Issue #306, opened 2022-08-10
Various ideas to discuss


Gejsi:
**Is your feature request related to a problem? Please describe.**
First of all, I'd like to say that I really like this project and appreciate all the work that has been done by the community.
I'm opening this issue to discuss some features that I think could be added (I don't want to spam the issues section).

## Starting from the minor adjustments
1. Dismantle Prisma client on `teardown`, this can be easily done by adding one line:
```ts
// src/pages/api/trpc/[trpc].ts
[...]
import { prisma } from '../../../server/db/client'
export default createNextApiHandler({
  router: appRouter,
  createContext,
  teardown: () => prisma.$disconnect(),
})
```
2. Loggers are really useful during dev: both tRPC and prisma offer some kind of logging. I noticed that @TheoBr already opened an issue about the former (#296). For Prisma, a possible approach may be:
```ts
export const prisma =
  global.prisma ||
  new PrismaClient({
    log:
      env.NODE_ENV === 'development' ? ['query', 'error', 'warn'] : ['error'],
  })

// unsure about this if statement,
// can be improved for the testing env?
if (env.NODE_ENV !== 'production') {
  // prisma middleware to show elapsed time
  prisma.$use(async (params, next) => {
    const before = Date.now()
    const result = await next(params)
    const after = Date.now()
    console.log(
      `Query ${params.model}.${params.action} took ${after - before}ms`
    )

    return result
  })

  global.prisma = prisma
}
```
## Major features
3. Currently all projects scaffolded by ct3a use a session-based strategy for auth.
Do you guys think that having a jwt-based strategy as a option could benefit the project? For example, a possible cli step:
```bash
- next-auth(database-strategy)
- next-auth(jwt-strategy)
- tRPC
- tailwind
- prisma
```
4. Add support for [Per-page layouts](https://nextjs.org/docs/basic-features/layouts#per-page-layouts):
```ts
export type NextPageWithLayout = NextPage & {
  getLayout?: (page: ReactElement) => ReactNode
}

type AppPropsWithLayout = AppProps & {
  Component: NextPageWithLayout
}

const MyApp = (({ Component, pageProps }: AppPropsWithLayout) => {
  const getLayout =
    Component.getLayout ?? ((page) => <MainLayout>{page}</MainLayout>)

  return getLayout(<Component {...pageProps} />)
}) as AppType

// at EOF
export default withTRPC<AppRouter>({})(MyApp)
```
[...]

Issue #296, opened 2022-08-09
Include tRPC client 'loggerLink' for dev debugging


TheoBr:
tRPC has an awesome logger client built in for React Query that is a lifesaver

<img width="377" alt="image" src="https://user-images.githubusercontent.com/6751787/183571692-d7d9436b-f727-41ad-bc2d-27ee3f53f9ff.png">

Adding the following to the `_app.tsx` trpc config will add them in dev

```ts
{
...
      links: [
        loggerLink({
          enabled: (opts) =>
            process.env.NODE_ENV === "development" ||
            (opts.direction === "down" && opts.result instanceof Error),
        }),
      ],
 }
 ```

**Comment by TheoBr** [comment]
Wait this appears to be broken now rip :(

**Comment by nexxeln** [comment]
> 

lol, do we still want this then?

**Comment by c-ehrlich** [comment]
if its possible to get it working in both tRPC v9 and v10 (or only v10 i guess), i think it would be a good thing to add. those console outputs are nice to have by default.

**Comment by TheoBr** [comment]
If we can get it working I definitely want it in dev mode

**Comment by hnqg** [comment]
Well, it worked here. But I had to use the `httpBatchLink` as well. With only the `loggerLink` I was getting a `TypeError: link is not a function`. Tested in other projects with trpc in version `9.26.2` and got the same error.

That log really helps in development, and I also find the `httpBatchLink` a good thing to add to trpc config. What do you think @TheoBr , do you still think it's useful to add the `loggerLink` and the `httpBatchLink`?

```
const links = [
      loggerLink({
        enabled: (opts) =>
          process.env.NODE_ENV === 'development' ||
          (opts.direction === 'down' && opts.result instanceof Error),
      }),
      httpBatchLink({
        maxBatchSize: 10,
        url,
      }),
    ];

return {
      url,
      transformer: superjson,
      links,
      /**
       * @link https://react-query.tanstack.com/reference/QueryClient
       */
      // queryClientConfig: { defaultOptions: { queries: { staleTime: 60 } } },
    };
```

![image](https://user-images.githubusercontent.com/25696006/183981719-67634c5c-23ac-4d29-8343-34fb764f1d1d.png)

Gejsi:
Tbh, I don't like the idea of having `httpBatchLink` enabled by default, it feels a little too opinionated.
I will open a PR soon for this issue.

TheoBr:
@hnqg sorry just saw this now - yeah would be nice to add this I think

[...]

PR #312, opened 2022-08-12
feat: add logs during dev

Gejsi:
# Add loggers during development

---

Ref #306, ref #296
- [x] Dismantle prisma client on `teardown`
- [x] Add prisma logs
- [x] Add tRPC logs 

---

## Screenshots
Example prisma logs:
image

Example tRPC logs:
image

💯


Gejsi:
For checkbox n1, I had to create a separate file called `api-handler-prisma.ts`, because putting the 1-liner
inside `api-handler.ts` would also add the line for users that don't use prisma.

Commit c4ffe36 fixes an incorrect query I found while testing some stuff out.

For checkbox n2: the prisma middleware has a small caveat, the amount of the same logs increases in a weird way due to HMR.
For example, the query shown in the screenshot, after **three** hot reloads becomes:
```
prisma:query [...]
Query Exampe.findMany took 21ms
Query Exampe.findMany took 21ms
Query Exampe.findMany took 21ms
```
But, it's not really deterministic (sometimes 3 hot reloads only cause a duplicate). So I decided to comment it out, instead of having it on by default.

For checkbox n3: the tRPC logger works only if `httpBatchLink` or `httpLink` are present in the config. Since batching is the default, I've put `httpBatchLink` explicitly.
A smarter approach could be taken by [enabling or disabling batching for certain requests](https://trpc.io/docs/links#using-a-splitlink-to-control-request-flow).
Having something like:
```ts
splitLink({
  condition({ context }) {
    // check for context property `includeBatch`
    return context.includeBatch === true
  },
  // when condition is true, use batching...
  true: httpBatchLink({ url, maxBatchSize: 10 }),
  // ...otherwise use normal requests
  false: httpLink({ url }),
}),
```
Batching requests for two queries:
```ts
const firstQuery = trpc.useQuery(['example.hello'], {
  context: {
    includeBatch: true
  }
})
const secondQuery = trpc.useQuery(['example.getAll'], {
  context: {
    includeBatch: true
  }
})
```
One issue with this approach is that tRPC doesn't currently offer a _strict_ typesafety for the `context` object (e.g. no autocomplete for `context.includeBatch`), but this could be solved by extending the type definition in a `foo.d.ts` file.

juliusmarminge:
We have migrated the repo into a monorepo with the website baked in. All CLI stuff now goes into [`cli`](https://github.com/t3-oss/create-t3-app/tree/main/cli)

Gejsi:
Any review progress for this PR @TheoBr?
There are conflicts between the branches, @juliusmarminge how should I resolve them in this case?

juliusmarminge:
Sorry om travelling foot so not much time for coding. I'll try to look it over tonight.

As for the conflicts, just resolve so that they are made to cli/src instead of src

Gejsi:
> As for the conflicts, just resolve so that they are made to cli/src instead of src, or cli/template instead of template etc

Ok.
I noticed that PR #330 fixes the stuff that I already fixed in a slightly different way in commit c4ffe36.
If the two PRs get merged together the project would still be broken.
So I either revert my commit or that PR should be closed without merging.

juliusmarminge:
> > As for the conflicts, just resolve so that they are made to cli/src instead of src, or cli/template instead of template etc
> 
> 
> 
> Ok.
> 
> I noticed that PR #330 fixes the stuff that I already fixed in a slightly different way in commit c4ffe36.
> 
> If the two PRs get merged together the project would still be broken.
> 
> So I either revert my commit or that PR should be closed without merging.

I prefer nex's solution with renaming the actual router, so you can revert that specific change on your PR 👍

juliusmarminge:
Will you fix the conflicts?

When you are at it, point the PR to `next`. tRPC 10 is coming closer and closer to release so I think it's better we push new features there?

Gejsi:
Tbh, I haven't been working on it since I was waiting for the review: there are still some decisions to be made and I didn't want to add features that would just be deleted.
With that said, I will fix the conflicts soon, should I create a new PR only for the `next` version?
Because theo's review isn't coming, what's your opinion on the first comment I made @juliusmarminge (logger caveat, batching...)?

juliusmarminge:
Oh sorry...

I redid the loggerLink on next #352, but more than happy to let you submit a new PR doing just that so that you'll be credited accordingly 😄 

I think the explicit batchLink is required, but we don't need to overwrite the default batchSize so we should stick to the default on that

Gejsi:
No worries, I will close this PR and open a new one for the `next` branch.
If you want to close your PR it's fine with me because I can redo that :+1:

juliusmarminge:
Make yours and I'll close and link them

Gejsi:
I cannot open PRs to the `next` branch because I'm not a member.
Maybe, I can make open a simple PR to `main` with the trpc v9 stuff and you will re-add them from your new PR to `next`?

juliusmarminge:
> I cannot open PRs to the `next` branch because I'm not a member.
> 
> Maybe, I can make open a simple PR to `main` with the trpc v9 stuff and you will re-add them from your new PR to `next`?

FR? Yea just make to main then. Sorry for the troubles

PR #313, opened 2022-08-12
fix: process.browser is deprecated use process typeof window

chriscn:
Just a simple fix of a deprecated check. Thank you so much for this awesome project. Much easier than setting it up myself over and over again!


asrvd:
we're already doing it, just see the few lines right above that

template/page-studs/_app/with-trpc.tsx
```diff
@@ -13,7 +13,7 @@ const getBaseUrl = () => {
   if (typeof window !== "undefined") {
     return "";
   }
```
asrvd, line 15: here

PR #330, opened 2022-08-18
fix: use auth instead of question while merging the router

nexxeln:
## ✅ Checklist

## Changelog

<p>
While scaffolding a project with Next Auth and tRPC, there is a `protected-example-router` to show how to protect tRPC routers with Next Auth. This router is merged as `question.` in the index file. But in `src/pages/index.tsx`, the query is called using `auth.getSecretMessage` which will obviously fail because its `question.getSecretMessage`. This PR fixes it and merges the router as `auth.`.
</p>

### DIFF
--- a/.changeset/cyan-bags-sit.md
+++ b/.changeset/cyan-bags-sit.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+use auth instead of question while merging the router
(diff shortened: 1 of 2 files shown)


juliusmarminge:
Generate a patch changeset so you'll be credited in the changelog

PR #334, opened 2022-08-18
fix: build fails on trpc, trpc+prisma

c-ehrlich:
## ✅ Checklist

---

## Changelog

As of 5.8.0, build fails when selecting only tRPC or tRPC + Prisma:
```
./src/server/router/context.ts
5:29  Error: Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.  @typescript-eslint/ban-types
```

This commit fixes this by replacing the `{}` type with `Record<string, never>` and a comment telling users to use an object if they want to pass things.

An alternative would be to relax the eslint rules, but I think keeping them as they are is better.

💯

### DIFF
--- a/.changeset/rude-bags-fetch.md
+++ b/.changeset/rude-bags-fetch.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+fix: build fails on trpc, trpc+next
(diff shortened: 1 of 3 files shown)


c-ehrlich:
if you decide to merge, let me know if you also want this on next

nexxeln (approved):

juliusmarminge (approved):

juliusmarminge:
> if you decide to merge, let me know if you also want this on next

That would be great. No changesets needed for next though

Issue #182, opened 2022-07-12
NextAuth GitHub integration throws an error - `refresh_token_expires_in` missing in Prisma schema


EkaanshArora:
**Describe the bug**
When you scaffold a project with NextAuth and Prisma. Setting up the `env` file and using the GitHub provider throws an error as the Prisma Account schema is missing the `refresh_token_expires_in` field. The [NextAuth docs page](https://next-auth.js.org/providers/github) suggests adding it to the schema.

Exact error message:
```
Unknown arg `refresh_token_expires_in` in data.refresh_token_expires_in for type AccountUncheckedCreateInput. 
```

**To Reproduce**

1. Scaffold a new project
2. Select NextAuth & Prisma
3. Setup GitHub credentials in the `env` file
4. Use this snippet:
```tsx
import { signIn } from "next-auth/react"

export default function Component() {
  return <>
    <button onClick={() => signIn()}>Sign in</button>
  </>
}
```
5. Sign in with GitHub
6. See the error in the log

**Expected behavior**
The sign-in should succeed.

The default GitHub integration throwing an error can be confusing for someone new to NextAuth. I'm happy to create a PR and add the field to the schema if it makes sense. If we don't want to change the default Schema for NextAuth, it might make sense to explore another default provider instead of GitHub.

nexxeln:
I think Discord is a good idea.

ozgurrgul:
For anyone seeing this error like me when trying github provider: you basically need to extend `Account` schema in `schema.prisma`:

```
model Account {
    id                String  @id @default(cuid())
    userId            String
    .....
    refresh_token_expires_in Int <<< Added new to provide github support
}
```

I tested with only Github provider and it works, not sure with multiple providers.

PR #338, opened 2022-08-19
fix(next-auth): Prisma Model update for next-auth

KubaRocks:
## ✅ Checklist

---

## Changelog

Default `next-auth` with `prisma` setup doesn't work with GitHub OAuth and according to [this issue](https://github.com/nextauthjs/next-auth/issues/3465) probably with some others providers too.
Updated model fixes the issue.

### DIFF
--- a/cli/template/addons/prisma/auth-schema.prisma
+++ b/cli/template/addons/prisma/auth-schema.prisma
@@ -20,19 +20,23 @@ model Example {
 
 // Necessary for Next auth
 model Account {
-    id                String  @id @default(cuid())
-    userId            String
-    type              String
-    provider          String
-    providerAccountId String
-    refresh_token     String? //@db.Text
-    access_token      String? //@db.Text
-    expires_at        Int?
-    token_type        String?
-    scope             String?
-    id_token          String? //@db.Text
-    session_state     String?
-    user              User    @relation(fields: [userId], references: [id], onDelete: Cascade)
+    id                       String  @id @default(cuid())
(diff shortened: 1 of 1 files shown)


juliusmarminge:
See #182. We switched to Discord as default since Github is a bit special in that it requires more setup.

KubaRocks:
According to issue I've linked, this also fixes issues with Facebook and Google (although I didn't test it).
If the goal is to stay with default schema that next-auth provides in their docs, then we can close that PR.
Although it would be cool to mention this issue somewhere, so the newcomers are not confused.

From my perspective - schema currently has some excessive nullable fields to fulfill variety of different auth scenarios. Adding this 3 fields helps OTB integrating other providers without need of researching why auth is failing.
Small cost comparing to improved usability.

juliusmarminge:
I cant see anything of this field being provided by Facebook or Google on the documentation: https://next-auth.js.org/providers/

Every provider provides a bit different fields, and I'm afraid if we start doing special treatments for one, people are going to wonder why we haven't done so for another.

nexxeln:
I think discord is fine, not too keen on merging this. We want to be as close to the official docs as possible.

nexxeln (commented):
I think discord is fine, not too keen on merging this. We want to be as close to the official docs as possible.

juliusmarminge:
We can probably include a few words about this and refer to the appropriate docs in our upcoming docs page, but we shouldn't include it by default

PR #346, opened 2022-08-23
fix: tsconfig include mjs and cjs

juliusmarminge:
Mirrors #345 on next.


juliusmarminge:
Would you mind updating docs on this branch so we can get the CI working again @nexxeln?
Don't need to keep it up to date after but getting it to build would be nice.

juliusmarminge:
> Would you mind updating docs on this branch so we can get the CI working again @nexxeln?
> Don't need to keep it up to date after but getting it to build would be nice.

nvm merging main was easier than I thought, see #348

PR #348, opened 2022-08-23
chore: Next merge 2022 08 23

juliusmarminge:
Curious if this merge will work


cli/template/addons/trpc/auth-context.ts
```diff
@@ -32,7 +32,7 @@ export const createContext = async (
 
   return await createContextInner({
     session,
-  });
+  }));
```
juliusmarminge, line 35: looks odd

www/pnpm-lock.yaml
```diff
@@ -0,0 +1,3510 @@
+lockfileVersion: 5.4
```
juliusmarminge, line 1: this also. individual packages should not have lock files

juliusmarminge (commented):
mostly good :)

PR #349, opened 2022-08-25
docs: write tailwind section in Why page

bdsqqq:
## ✅ Checklist

---

## Changelog

- Added tailwind section to the Why page. Based on their landing page, Theo's stream & this [tweet from tailwind's creator](https://twitter.com/adamwathan/status/1560749403131895808).


bdsqqq:
I've been looking at the commits for the docs for a bit and will try to add some stuff to them in this PR, not really sure what yet but I'll at least try to get rid of the TODOs on the Why page.

I opened this VERY early because I want feedback on the stuff I write for these so let me know if discord would be a better place and I can close this and reopen once stuff is ready. <3

nexxeln (commented):
Thanks for contributing!

I think we can also mention the cool flow state it provides while styling.

bdsqqq:
The funny thing is I had a third point in this about the "zen-mode" it allows, great to know it should be included. Will work on it a bit and add it to this draft

bdsqqq:
Not sure about the "inline-like approach" in the zen-mode paragraph. For me writing tailwind feels like inlines and I heard others say the same but the creator of tailwind wrote about this comparison wayyy back [in this article](https://adamwathan.me/css-utility-classes-and-separation-of-concerns/). Might be worth changing this part but the previous paragraph adds the same context adam does in the article with his "It's kind of like inlines with constraints in the values™". 

What y'all think?

nexxeln (approved):
I like this

juliusmarminge (approved):
👍🏼 feel free or add or dismiss my suggestion. Just a bit of a banter for all who thinks it compares to finished design systems

bdsqqq:
Great point! this kind of thing makes me think we might eventually want a place in the docs to link to other resources for each piece of tech. Especially with Tailwind, being able to dump links to the docs, videos or comparisons would allow us to respond to "iS THis LikE maTERial Ui?" way faster, and in a perfect world, avoid the question being asked in the first place. 

Just a thought tho, can't imagine this being that useful right now, people can google stuff on their own.

juliusmarminge:
> Great point! this kind of thing makes me think we might eventually want a place in the docs to link to other resources for each piece of tech. Especially with Tailwind, being able to dump links to the docs, videos or comparisons would allow us to respond to "iS THis LikE maTERial Ui?" way faster, and in a perfect world, avoid the question being asked in the first place.
> 
> Just a thought tho, can't imagine this being that useful right now, people can google stuff on their own.

Yea we'll have sections for each tech so maybe we put that there?

PR #367, opened 2022-08-28
fix(www): build errors and CI improvements

juliusmarminge:
cc @kroucher 

Fixes build errors when running `pnpm build:www` and added CI check for it.

`www` could be successfully build on Vercel without catching some errors which is sub-optimal.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,1 +1,5 @@
 blank_issues_enabled: false
+contact_links:
+  - name: Ask a question
+    url: https://t3.gg/discord
+    about: Ask questions and discuss with other community members
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,5 +1,8 @@
 "area: cli":
-  - any: ["src/**"]
+  - any: ["cli/src/**"]
 
 "area: t3-app":
-  - any: ["template/**"]
+  - any: ["cli/template/**"]
+
+"documentation":
+  - any: ["www/**", "**/*.md"]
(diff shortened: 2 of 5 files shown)


juliusmarminge:
Astro needs to add some peer-deps it looks like xd

We could also just roll our own icons, or export something like `react-icons`/ from react-components? Wouldn't ship any JS anyways? `font-awesome` or `feather-icons` also options

juliusmarminge:
Yea this is too annoying. Changing icon package

nexxeln:
I vote for feathericons

juliusmarminge:
> I vote for feathericons

Feel free to take over... turns out even icons are hard in astro...

juliusmarminge:
btw reason I want this to pass in CI is cause currently if you clone the repo and run `pnpm build`, you'll fail with errors and the dev might think they did something wrong

juliusmarminge:
Alright found the issue. Turns out one of `astro-icon`'s reps (sharp) is a cjs library and can thus not be used.

kroucher:
Totally all for swapping it out for something else, if it's the cause of the issues, consider it gone - I think the only reason the library is there was lazy access to the brand icons (like nextjs/TS/tailwind etc), I'd be happy to just grab the SVGs and store them locally if that's easiest unless there's another icon lib with them

juliusmarminge:
> Totally all for swapping it out for something else, if it's the cause of the issues, consider it gone - I think the only reason the library is there was lazy access to the brand icons (like nextjs/TS/tailwind etc), I'd be happy to just grab the SVGs and store them locally if that's easiest unless there's another icon lib with them

https://simpleicons.org has them but no Astro integration so just grab the svgs I think will be the easiest

juliusmarminge:
bruh not a single __dirname reference in the entire www folder (including node_modules) wtf is this

juliusmarminge:
Cool that's what happens when deps (@astro/image in this case) hides shit under the hood...

juliusmarminge:
merge and continue iterating

PR #372, opened 2022-08-29
fix: astro build

c-ehrlich:
Closes #371

## ✅ Checklist

- [x] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-08-15).
- [x] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] I performed a functional test on my final commit

---

## Changelog

- Changes components that were arrow functions to named functions
- This fixes the build (edit: on my machine), see: https://github.com/withastro/astro/issues/4220
- Let me know if you want me to rename the files to match the new component names (named function components need to start with uppercase letters)

---

## Screenshots

image

💯

PR #373, opened 2022-08-29
fix: throw error on missing snowpack github token

c-ehrlich:
## ✅ Checklist

---

## Changelog

- the code block for throwing an error if the SNOWPACK_PUBLIC_GITHUB_TOKEN is missing was inaccessible because the value is either equal to the token itself or 'hello', both of which pass a null check
- might be related to https://github.com/t3-oss/create-t3-app/pull/372

---

## Screenshots

image

💯

### DIFF
--- a/www/src/components/blog/avatarList.astro
+++ b/www/src/components/blog/avatarList.astro
@@ -17,8 +17,8 @@ type Commit = {
 
 async function getCommits(url: string) {
   try {
-    const token = import.meta.env.SNOWPACK_PUBLIC_GITHUB_TOKEN ?? "hello";
-    if (!token || token === "hello") {
+    const token = import.meta.env.SNOWPACK_PUBLIC_GITHUB_TOKEN;
+    if (!token) {
       throw new Error(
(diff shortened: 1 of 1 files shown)


nexxeln (approved):
Good to have I guess

www/src/components/blog/avatarList.astro
```diff
@@ -18,7 +18,7 @@ type Commit = {
 async function getCommits(url: string) {
   try {
     const token = import.meta.env.SNOWPACK_PUBLIC_GITHUB_TOKEN ?? "hello";
-    if (!token) {
+    if (!token || token === "hello") {
```
juliusmarminge, line 21: I think we should probably remove the `?? 'hello'` instead?

c-ehrlich: yeah  i agree

juliusmarminge:
also I don't know if SNOWPACK_PUBLIC is the correct prefix anymore? https://docs.astro.build/en/guides/environment-variables/

nexxeln:
Thanks!

PR #383, opened 2022-08-31
feat: safeguards for git repos

### LINKED ISSUE #378: bug: git repo is kept when selecting `overwrite`

```
  System:
    OS: macOS 12.5.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 136.66 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.1 - ~/.nvm/versions/node/v16.15.1/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.15.1/bin/npm
```

### Describe the bug

Since introducing overwrite option in #375 it seems like it's not overwriting the git repo. This is how it looks after I have created an app and committed an "external" file:

CleanShot 2022-08-30 at 13 46 25@2x

Scaffolding a new app and selecting overwrite, doesn't overwrite the git repo, since this file is still committed afterwards.

### To reproduce

```bash
npx create-t3-app@beta
```
No need for any packages, just make sure to initialize Git.

Create a sample file, and commit it.

Now create another app in the same directory, once again, initialize Git.

You'll see the sample file is still committed, indicating the Git repo was not reset.

### Additional information

cc @c-ehrlich 

This should probably be validated at the step where the Git repo is initialized, something like:

```
The directory already includes a Git repo, would you like to delete it and create a new one?
```

can be prompted if selecting `Yes`?

> **NOTE:** I noticed the `.env` also isn't overwritten (probably goes for everything hidden (i.e. `.filename`).

juliusmarminge:
Closes #378
cc @c-ehrlich 

## ✅ Checklist

---

## Changelog

- fixes #378 by deleting the `.git` repo first. Copying files to hidden directories didn't seem to overwrite them.
- adds some safeguards for initializing git repos in current worktrees etc
- allow user to keep the current git repo if there is one

> Note: Also updated some scripts I noticed didn't work properly
---

## Screenshots

_[Screenshots]_
CleanShot 2022-08-31 at 11 02 09@2x


c-ehrlich:
This makes sense to me. An existing git repo should be either left alone, or wiped. Writing new files into it seems like a terrible idea, even if the user selected "overwrite".

juliusmarminge:
> This makes sense to me. An existing git repo should be either left alone, or wiped. Writing new files into it seems like a terrible idea, even if the user selected "overwrite".

Do you mean writing files in the .git? That will be prevented indeed.

But the 'continue' option here would still let the scaffolding proceed (well, in the order we do things, it's already done) and the files would be added to the existing git repository.

juliusmarminge:
Would you be willing to give this a proper review too @c-ehrlich ?

cli/src/helpers/git.ts
```diff
@@ -0,0 +1,112 @@
...
+      default: false,
+    });
+    if (!overwriteGit) {
+      spinner.info("Skipping Git initialization.");
+      return;
+    }
+    // Deleting the .git folder
+    fs.removeSync(path.join(projectDir, ".git"));
+  } else if (isInside && !isRoot) {
+    // Dir is inside a git worktree
```
c-ehrlich, line 71: need a `spinner.stopAndPersist()` here

c-ehrlich (changes requested):

cli/src/helpers/git.ts
juliusmarminge: 👍🏼

c-ehrlich (approved):

c-ehrlich:
lgtm :)

PR #385, opened 2022-08-31
feat: updated UI design of landing page

PiotrekPKP:
## ✅ Checklist

---

## Changelog
I updated UI of landing page - removed some "not that great" stuff and added better looking (at least for me :D) styles.

---

## Screenshots
localhost_3000_

💯

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,10 +12,6 @@
     "type": "git",
     "url": "https://github.com/t3-oss/create-t3-app.git"
   },
-  "workspaces": [
-    "www",
-    "cli"
-  ],
   "keywords": [
     "create-t3-app",
     "init.tips",
--- a/package.json
+++ b/package.json
@@ -12,6 +12,10 @@
     "type": "git",
     "url": "https://github.com/t3-oss/create-t3-app.git"
   },
+  "workspaces": [
+    "www",
+    "cli"
+  ],
   "keywords": [
     "create-t3-app",
     "init.tips",
--- a/www/src/components/landingPage/community/community.astro
+++ b/www/src/components/landingPage/community/community.astro
@@ -1,4 +1,6 @@
 ---
+import CommunityCard from "./communityCard.astro";
+
 // eslint-disable-next-line -- need to export something to make it a module
 export interface Props {}
 ---
@@ -7,7 +9,7 @@ export interface Props {}
   <div
     class="flex flex-col items-center justify-center max-w-7xl text-t3-purple-100 mx-auto"
   >
-    <h1 class="text-3xl py-3">Community</h1>
+    <h1 class="text-3xl py-3 font-bold">Community</h1>
     <p class="text-center pb-8 px-4">
       Join our community to get help, share your projects, and even contribute
       to the project!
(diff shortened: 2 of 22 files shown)


nexxeln (commented):
I definitely like this more than our current design for the tech cards

package.json
```diff
@@ -12,6 +12,10 @@
     "type": "git",
     "url": "https://github.com/t3-oss/create-t3-app.git"
   },
+  "workspaces": [
```
juliusmarminge, line 15: This is not needed since we are using pnpm workspaces: https://pnpm.io/pnpm-workspace_yaml

www/src/components/navigation/navbar.astro
```diff
@@ -1,44 +1,64 @@
...
-          href="https://init.tips"
-          target="_blank"
-          rel="noopener noreferrer"
-          ><span class="">T3 Stack</span>
-        </a>
+        {
+          navbarLinks.map((navbarLink) => (
+            <a
+              class={`inline-flex items-center border font-medium relative text-base px-3 py-2 rounded-md border-transparent ${
+                mode === "landing"
```
juliusmarminge, line 51: Reviewing this on mobile so can't see the full change but why do we need this? If there is contrast issues, is it not better to adjust the bg on the docs-layout?

www/src/components/landingPage/community/community.astro
```diff
@@ -7,80 +9,68 @@ export interface Props {}
...
-        target="_blank"
-        rel="noopener noreferrer"
-        class="flex flex-col items-start justify-center border border-t3-purple-100 rounded-lg p-3 space-y-3 h-48 sm:h-56 hover:bg-white/10 cursor-pointer w-full"
       >
         <svg
           role="img"
           viewBox="0 0 24 24"
           xmlns="http://www.w3.org/2000/svg"
           class="fill-[#5865F2]"
+          slot="icon"
```
juliusmarminge, line 30: Astro 1.1.2 is now out which fixes the typing of SVGs. Updating to that will silence the issues and pass CI 👍

juliusmarminge (changes requested):
Love the design changes. Looks awesome 👏

www/src/components/navigation/navbar.astro
PiotrekPKP: I wanted to keep the color (at least for now) in the docs, whereas I tweaked it on the landing page to be much lighter. I didn't want to touch docs page for now so I added the `mode` prop

kroucher (changes requested):
Looks really awesome, thanks heaps! Starting to really look like a professional site now :D Only things I would suggest:

1. Small bit of padding under Read More on mobile

image

2. Perhaps make the div containing card description + Read More a flex-col with `justify-between` to always slot the Read More at the bottom of the card on larger screens

image

www/src/components/landingPage/stack/card.astro
```diff
@@ -1,20 +1,33 @@
...
+      <slot name="icon" />
+      <p class="text-lg leading-6 md:text-xl font-medium text-t3-purple-200">
+        {title}
+      </p>
+    </dt>
+  </a>
+  <dd class="m-6 text-sm md:text-base text-t3-purple-100 subpixel-antialiased">
     <slot name="description" />
   </dd>
+  <div class="m-6">
```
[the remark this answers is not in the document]
PiotrekPKP: This and the previous problem will be fixed soon :smile:

PiotrekPKP:
> Looks really awesome, thanks heaps! Starting to really look like a professional site now :D Only things I would suggest:
> 
> 1. Small bit of padding under Read More on mobile
> 
> image
> 
> 2. Perhaps make the div containing card description + Read More a flex-col with `justify-between` to always slot the Read More at the bottom of the card on larger screens
> 
> image

Doing fixes now!

www/src/components/navigation/navbar.astro
juliusmarminge: Makes sense. There are other stuff needed on the docs page so it is perfectly fine for now :)

PiotrekPKP:
I think everything should be fixed right now :)

kroucher (approved):
<3!

Looks really nice, thanks so much!

My janky mobile menu on the landing page will get fixed with my existing PR today too 👍

www/src/components/landingPage/community/community.astro
```diff
@@ -7,80 +9,68 @@ export interface Props {}
...
-        <p class="text-left">
-          Contribute to the project by submitting issues and pull requests on
-          our Github repository.
-        </p>
-      </a>
-      <a
+      </CommunityCard>
+      <CommunityCard
+        title="Twitter"
+        text="Join us on Twitter for shitposts"
```
juliusmarminge, line 58: Not a problem for now, but just noting it down.

We should either fix the styles here or just add some more text so it aligns with the others 
CleanShot 2022-08-31 at 23 27 00@2x

juliusmarminge (approved):
Great!

www/src/components/landingPage/community/community.astro
PiotrekPKP: I'll try to update this tomorrow ;)

PR #387, opened 2022-09-01
feat: add basic i18n configuration in next.config

Vallerem:
## ✅ Checklist

~~- [] I performed a functional test on my final commit~~

---

## Changelog

Added basic next i18n configuration:

- The generated `next.config.mjs` for new projects will also include this small configuration, this can be helpful as a reference point to add new locales or at least as a hint on how to work with locales.
- Improves the generated project lighthouse accessibility score to 97 by adding the `lang` attribute to the html.

---

## Screenshots

Screenshot 2022-09-01 at 20 38 14


c-ehrlich:
I like pointing users towards i18n because it's an easy win, and overlooked so often. I don't know if setting a default makes sense though - I'd rather a page has no lang attribute than an incorrect one. 

Maybe it's better to include this in the next.config, but commented out?

What does everyone else think?

PiotrekPKP:
I think it would be nice to have it done. I'm just building an app that needed i18n and a comment would be nice to have :)

nexxeln:
I think this is too much for a template, not too keen on merging this.

juliusmarminge (commented):
Not sure either. The docs are quite clear on this so I don't know if it provides much value https://nextjs.org/docs/advanced-features/i18n-routing

Vallerem:
Maybe it's not that relevant as a small hint into localisation, as @juliusmarminge already pointed out the docs are very straightforward. I still think it would be nice to at least generate the main html with the `lang` attribute for accessibility :)

nexxeln:
> Maybe it's not that relevant as a small hint into localisation, as @juliusmarminge already pointed out the docs are very straightforward. I still think it would be nice to at least generate the main html with the `lang` attribute for accessibility :)

That we can do for sure

juliusmarminge:
> > Maybe it's not that relevant as a small hint into localisation, as @juliusmarminge already pointed out the docs are very straightforward. I still think it would be nice to at least generate the main html with the `lang` attribute for accessibility :)
> 
> 
> 
> That we can do for sure, but then if someone makes their app in spanish and they don't notice the next config, it'll lead to bad accessibility. I'm not sure what the correct solution is here.

Meh? A few points increase in accessibility on the scaffolded app that people will just rip out doesn't seem worth adding stuff that could me misleading imo.

juliusmarminge:
> > > Maybe it's not that relevant as a small hint into localisation, as @juliusmarminge already pointed out the docs are very straightforward. I still think it would be nice to at least generate the main html with the `lang` attribute for accessibility :)
> > 
> > 
> > 
> > That we can do for sure, but then if someone makes their app in spanish and they don't notice the next config, it'll lead to bad accessibility. I'm not sure what the correct solution is here.
> 
> Meh? A few points increase in accessibility on the scaffolded app that people will just rip out doesn't seem worth adding stuff that could me misleading and lead to future issues imo.

TheoBr:
fwiw this is a thing I often miss. I think this is more Next's problem to solve than ours tho. Leaning "close" but I definitely see the value and could be convinced

c-ehrlich:
I'm split between close, and add but comment out, don't feel strongly in either direction.

Vallerem:
To me it just felt a bit wrong not having a language stablished general html in the final build, specially for accessibility. Since the default text is already in english and i18 being is the default (and only?) way to generate this attribute it feels like something that can add some minor value or at least point into a good direction for localisation and accessibility. This being said I agree that this is maybe a more Next oriented thing but having it at least a s commented code could help.

juliusmarminge:
> To me it just felt a bit wrong not having a language stablished general html in the final build, specially for accessibility. Since the default text is already in english and i18 being is the default (and only?) way to generate this attribute it feels like something that can add some minor value or at least point into a good direction for localisation and accessibility. This being said I agree that this is maybe a more Next oriented thing but having it at least a s commented code could help.

Alright I'm actually leaning yes now. Setting English by default and then linking to the docs page for i18n

juliusmarminge (approved):
I'm okay with this. What are your thought @nexxeln @c-ehrlich ?

c-ehrlich:
I like it. After thinking about it again, I don't think having it set to EN initially is worse than having it commented - it's easy to change. And having it at all will significantly increase the percentage of apps using T3 that have it set at all, I've certainly forgotten to do it in the past.

eddyhdzg:
Please, without i18n I can't use t3. 😢

`https://github.com/i18next/next-i18next` support would be great.

PR #395, opened 2022-09-04
refactor: move createProtectedRouter to auth-context.ts

Pearcekieser:
Closes #<issue>

## ✅ Checklist

---

## Changelog

_Moved createProtectedRouter into auth-context.ts._

I'm experimenting with the t3 stack.

I think this makes the template a little easier to navigate.  Having `protected-router.ts` and `protected-example-router.ts`, based on filenames alone I didn't know one was more of a utility wrapper (defining the createProtectedRouter function) and the other the example.  I moved the createProtectedRouter function into the auth-context.ts so it goes into the same as the createRouter() definition when auth is also enabled.

---

## Screenshots

_[Screenshots]_

💯


Pearcekieser:
After getting a bit deeper into how the cli works, I see my change introduces some code duplication.  I'm going to leave this open for feedback but no hard feelings if this is not a direction supported by maintainers.

juliusmarminge (approved):
I don't have a strong opinion here but kinda makes sense that it's placed next to `createRouter` and would match the setup we have for the v10 version: https://github.com/t3-oss/create-t3-app/blob/next/cli/template/addons/trpc/auth-server-utils.ts

juliusmarminge:
> After getting a bit deeper into how the cli works, I see my change introduces some code duplication. I'm going to leave this open for feedback but no hard feelings if this is not a direction supported by maintainers.

What in particular do you mean? We have A LOT of code duplication which is required to get all the different branches working properly 🥲

PR #396, opened 2022-09-04
feat: forward pass trpc headers during ssr

### LINKED ISSUE #280: feat: forward pass trpc headers during ssr
**Is your feature request related to a problem? Please describe.**

As [called out by Theo](https://clips.twitch.tv/SleepyVibrantGrasshopperDxCat-ZJjzoC-yQHmSmyDE) and [mentioned in this tRPC issue](https://github.com/trpc/trpc/issues/1811) currently `tRPC` has a bit of an unintuitive behaviour where during SSR Client headers are not merged and so it is impossible to access cookies, for say getting the users session from `next-auth` during SSR. 

When looking for the solution to this problem I turned to `t3-app` to try and see if SSR was working here and out of the box it wasn't!

**Describe the solution you'd like**

It looks like this header merging is just going to be documented rather then done by default in the tRPC implementation and so the `tRPC` Template for `t3-app` should include this bit of code?  

[template/addons/trpc/utils.ts](https://github.com/t3-oss/create-t3-app/blob/d021b2325301f0b56615ca10715d801ac9ff446d/template/addons/trpc/utils.ts#L18)
```ts
.......
export const trpc = setupTRPC<AppRouter>({
  config() {
    return {
      url: `${getBaseUrl()}/api/trpc`,
      transformer: superjson,

      // Additional code to merge headers below:
      headers: () => {
        if (ctx?.req) {
          // on ssr, forward client's headers to the server
          return {
            ...ctx.req.headers,
            "x-ssr": "1",
          };
        }
        return {};
      },

    };
  },
  ssr: false, // In t3-app SSR is false by default, but above is needed if turned on! 
});
.......
```

NOTE: This is the fix for tRPC V10 ... Not looked into where this goes in V9 yet! 

**Describe alternatives you've considered**

for tRPC to add this into the implementation but this does not look like it is going that way!

rhenriquez28:
Closes #280 

## ✅ Checklist

- ~~[ ] I performed a functional test on my final commit~~

---

## Changelog

Added the implementation to forward client's headers when using SSR as a comment. For more info see https://github.com/t3-oss/create-t3-app/issues/280.

💯

### DIFF
--- a/cli/template/page-studs/_app/with-trpc.tsx
+++ b/cli/template/page-studs/_app/with-trpc.tsx
@@ -42,17 +42,17 @@ export default withTRPC<AppRouter>({
       // queryClientConfig: { defaultOptions: { queries: { staleTime: 60 } } },
 
       // To use SSR properly you need to forward the client's headers to the server
-      /*headers: () => {
-        if (ctx?.req) {
-          const headers = ctx?.req?.headers;
-          delete headers?.connection;
-          return {
-            ...headers,
-            "x-ssr": "1",
-          };
-        }
-        return {};
-      },*/
+      // headers: () => {
+      //   if (ctx?.req) {
+      //     const headers = ctx?.req?.headers;
+      //     delete headers?.connection;
+      //     return {
+      //       ...headers,
+      //       "x-ssr": "1",
+      //     };
+      //   }
+      //   return {};
+      // }
     };
   },
   /**
--- a/cli/template/page-studs/_app/with-trpc.tsx
+++ b/cli/template/page-studs/_app/with-trpc.tsx
@@ -40,6 +40,19 @@ export default withTRPC<AppRouter>({
        * @link https://react-query.tanstack.com/reference/QueryClient
        */
       // queryClientConfig: { defaultOptions: { queries: { staleTime: 60 } } },
+
+      // To use SSR properly you need to forward the client's headers to the server
+      /*headers: () => {
+        if (ctx?.req) {
+          const headers = ctx?.req?.headers;
+          delete headers?.connection;
+          return {
+            ...headers,
+            "x-ssr": "1",
+          };
+        }
+        return {};
+      },*/
     };
   },
   /**
--- a/cli/template/page-studs/_app/with-auth-trpc.tsx
+++ b/cli/template/page-studs/_app/with-auth-trpc.tsx
@@ -50,17 +50,17 @@ export default withTRPC<AppRouter>({
       // queryClientConfig: { defaultOptions: { queries: { staleTime: 60 } } },
 
       // To use SSR properly you need to forward the client's headers to the server
-      /*headers: () => {
-        if (ctx?.req) {
-          const headers = ctx?.req?.headers;
-          delete headers?.connection;
-          return {
-            ...headers,
-            "x-ssr": "1",
-          };
-        }
-        return {};
-      },*/
+      // headers: () => {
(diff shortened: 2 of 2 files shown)


cli/template/page-studs/_app/with-trpc.tsx
```diff
@@ -40,6 +40,19 @@ export default withTRPC<AppRouter>({
        * @link https://react-query.tanstack.com/reference/QueryClient
        */
       // queryClientConfig: { defaultOptions: { queries: { staleTime: 60 } } },
+
+      // To use SSR properly you need to forward the client's headers to the server
+      /*headers: () => {
```
TheoBr, line 45: This style of comment is not ECMAscript standard compliant. Can you comment the code back in and use VS Code to "comment it out"? Should result in // in front of each line

rhenriquez28: Sure thing. On it.

juliusmarminge:
Please do a changeset

juliusmarminge:
> > Please do a changeset
> 
> 
> 
> Sure thing. Just to be sure, this is a patch, right?

Not too strict on this, could argue both ways, i.e adding ssr or fixing ssr. Choose which you think is most appropriate

juliusmarminge:
On second look, if we have this here as is the user might think that uncommenting will enable ssr. Maybe we should add an explicit `ssr: false` to the config to make that part more clear?

juliusmarminge:
On second look, if we have this here as is the user might think that uncommenting will enable ssr. Maybe we should add an explicit `ssr: false` to the config to make that part more clear?

rhenriquez28:
Isn't that already explicitly set in the `withTRPC` config though? Unless you're referring to other config I'm not aware of?

juliusmarminge:
> Isn't that already explicitly set in the `withTRPC` config though? Unless you're referring to other config I'm not aware of?

https://trpc.io/docs/v9/nextjs#withtrpc-options

rhenriquez28:
I see what you mean, but the templates we have for the tRPC config already have an explicit `ssr: false`. Check https://github.com/t3-oss/create-t3-app/blob/main/cli/template/page-studs/_app/with-trpc.tsx and https://github.com/t3-oss/create-t3-app/blob/main/cli/template/page-studs/_app/with-auth-trpc.tsx

juliusmarminge:
> I see what you mean, but the templates we have for the tRPC config already have an explicit `ssr: false`. Check https://github.com/t3-oss/create-t3-app/blob/main/cli/template/page-studs/_app/with-trpc.tsx and https://github.com/t3-oss/create-t3-app/blob/main/cli/template/page-studs/_app/with-auth-trpc.tsx

Oooh haha I see... had to expand the diff, sorry bout that

rhenriquez28:
Haha, no worries. Although I do think the process to forward headers in SSR should be documented, at least as an optional step, in tRPC docs. I plan to head over there and reignite that convo again and see if they accept my help to make that doc more clear and complete.

juliusmarminge:
> Haha, no worries. Although I do think the process to forward headers in SSR should be documented, at least as an optional step, in tRPC docs. I plan to head over there and reignite that convo again and see if they accept my help to make that doc more clear and complete.

Please do! We have an open issue about it https://github.com/trpc/trpc/issues/2420

timotius02:
It looks like this change has been removed from current version. Have they been moved somewhere or are the manual header forwarding not needed anymore?

PR #402, opened 2022-09-05
docs: write Prisma section in Why page

bdsqqq:
## ✅ Checklist

---

## Changelog

Added Prisma section to the Why page. Based on a lot of sources so I'm throwing them into [this gist](https://gist.github.com/bdsqqq/25c2d2b5509b5793fd9ae1fd3bd491e1).

---

## Screenshots

image

💯


bdsqqq:
I'm not nearly as familiar with Prisma as I was with tailwind when I wrote the other Why, feedback is VERY appreciated.

bdsqqq:
I also wanted to add something about how Prisma doesn't care about the DB you're using as long as a connector exists for it but didn't know how to write it...

nexxeln (commented):
Yeah we can say that Prisma also doesn't care what db you're using, be it Postgres or MySQL, or even CockroachDB, it will give you the same DX everywhere. You can even connect to mongo if you're into that.

We can also talk about how Prisma allows for each database migrations, like moving from postgres to mysql, etc.

Also we can talk about prisma studio.

c-ehrlich:
Regarding "features":
- Agree that mentioning Prisma works on so many different DBs, from sqlite to various 'real' SQL or even Mongo, is important.
- Prisma Studio is good to mention I think.
- Introspection maybe? I've found that to be very powerful.

That being said I think the main focus should be on DX rather than features because that's the main thing where Prisma is the best. Maybe even put the TS metaphor first?

ajcwebdev:
> Yeah we can say that Prisma also doesn't care what db you're using, be it Postgres or MySQL, or even CockroachDB, it will give you the same DX everywhere. You can even connect to mongo if you're into that.
> 
> We can also talk about how Prisma allows for each database migrations, like moving from postgres to mysql, etc.

While all of this is technically true and is part of the big pitch for Prisma, there's a lot of nuance here that can lead to footguns if devs lean too heavily into this. It's not super straight forward to actually change your migrations and there's a lot of caveats. One of those databases you mentioned doesn't even have migrations for example. The concept of moving your migrations from a relational db to Mongo is nonsensical.

See this section from the Prisma docs: [You cannot automatically switch database providers](https://www.prisma.io/docs/concepts/components/prisma-migrate/prisma-migrate-limitations-issues#you-cannot-automatically-switch-database-providers).

the tldr is migrations create a `migration_lock.toml` file that is checked into your version history which requires you to wipe and rerun your migrations if you want to change the database provider. That can be okay, but there needs to be a clear call out that if you want to change providers you can't necessarily take your migrations with you, you can regenerate brand new ones that hopefully don't break.

juliusmarminge:
I'll merge and we can iterate on it

PR #407, opened 2022-09-08
refactor(cli): avoids big code blocks

minsk-dev:
Closes #<issue>

## ✅ Checklist

---

## Changelog

Moves some big blocks of code to functions, so the main function can be clearer/easier to read.

💯

### DIFF
--- a/cli/src/cli/index.ts
+++ b/cli/src/cli/index.ts
@@ -184,11 +184,15 @@ export const runCli = async () => {
         });
         cliResults.appName = appName;
 
-        await askLanguage();
-        cliResults.packages = await askPackages();
-        if (!cliResults.flags.noGit) cliResults.flags.noGit = !(await askGit());
-        if (!cliResults.flags.noInstall)
-          cliResults.flags.noInstall = !(await askInstall());
+        await promptLanguage();
+        cliResults.packages = await promptPackages();
+        if (!cliResults.flags.noGit) {
+          cliResults.flags.noGit = !(await promptGit());
+        }
+
+        if (!cliResults.flags.noInstall) {
+          cliResults.flags.noInstall = !(await promptInstall());
+        }
       }
     }
   } catch (err) {
@@ -208,7 +212,7 @@ export const runCli = async () => {
   return cliResults;
 };
 
-const askLanguage = async (): Promise<void> => {
+const promptLanguage = async (): Promise<void> => {
   const { language } = await inquirer.prompt<{ language: string }>({
     name: "language",
     type: "list",
@@ -227,7 +231,7 @@ const askLanguage = async (): Promise<void> => {
   }
 };
 
-const askPackages = async (): Promise<AvailablePackages[]> => {
+const promptPackages = async (): Promise<AvailablePackages[]> => {
   const { packages } = await inquirer.prompt<Pick<CliResults, "packages">>({
     name: "packages",
     type: "checkbox",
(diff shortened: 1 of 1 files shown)


cli/src/cli/index.ts
```diff
@@ -185,76 +183,12 @@ export const runCli = async () => {
...
-
-        if (runInstall) {
-          logger.success("Alright. We'll install the dependencies for you!");
-        } else {
-          cliResults.flags.noInstall = true;
-          logger.info(
-            `No worries. You can run '${pkgManager} install' later to install the dependencies.`,
-          );
-        }
+        await askLanguage();
```
juliusmarminge, line 187: I'd rename these to `promtX`

cli/src/cli/index.ts
```diff
@@ -185,76 +183,12 @@ export const runCli = async () => {
...
+        cliResults.packages = await askPackages();
+        if (!cliResults.flags.noGit) cliResults.flags.noGit = !(await askGit());
+        if (!cliResults.flags.noInstall)
+          cliResults.flags.noInstall = !(await askInstall());
```
juliusmarminge, line 191: kind of hard to read when some break lines and some don't. In these cases it's better to just force it I think:

```ts
if (!cliResults.flags.noGit) {
  cliResults.flags.noGit = !(await promptGit());
}
if () {
  ...
}
```

juliusmarminge (commented):
I think it's a good change.

Also seems like you "missed" the name-part.

cli/src/cli/index.ts
```diff
@@ -273,3 +207,81 @@ export const runCli = async () => {
...
+  return git;
+};
+
+const askInstall = async (): Promise<boolean> => {
+  const packageManager = getUserPkgManager();
+
+  const { install } = await inquirer.prompt<{ install: boolean }>({
+    name: "install",
+    type: "confirm",
+    message: `Would you like us to run ${packageManager} install?`,
```
juliusmarminge, line 274:
```suggestion
    message: `Would you like us to run '${packageManager} install'?`,
```

juliusmarminge:
> Also seems like you "missed" the name-part.

Bring that out to a separate function too for coherency

minsk-dev:
Oh that's what you meant. Sure

juliusmarminge (approved):
Great 👍

cli/src/cli/index.ts
```diff
@@ -166,94 +166,21 @@ export const runCli = async () => {
...
-          type: "confirm",
-          message: "Initialize a new git repository?",
-          default: true,
-        });
-        if (git) {
-          logger.success("Nice one! Initializing repository!");
-        } else {
-          cliResults.flags.noGit = true;
-          logger.info("Sounds good! You can come back and run git init later.");
+        cliResults.appName = await promptAppName();
```
juliusmarminge, line 175: Missing ending bracket here I think? On mobile though so may be wrong?

minsk-dev: It makes more sense if you try seeing only the last commit

c-ehrlich: I'm also pretty sure that the logic is incorrect on this. Try this:
```
cd cli
pnpm build
pnpm start someAppName
```

juliusmarminge: You are running every prompt conditionally depending on if the name was provided as option

minsk-dev: @c-ehrlich You are right. I did this while doing my own project, so I lost some context. Fixed now

juliusmarminge (approved):
Looks correct. Since this logic isn't caught by CI I suggest we do some manual tests to be sure before merging.

c-ehrlich:
behaves correctly both with and without name provided to the cli

image

minsk-dev:
I retested being pushing too

juliusmarminge:
Cool I'll merge, we can test @beta afterwards

Issue #440, opened 2022-09-13
feat(docs): fix lightmode


juliusmarminge:
### Is your feature request related to a problem? Please describe.

#439 disabled lightmode since there we're a few bugs here and there and didn't want to have that slow us down launching the site.

Should have it in the future though. Anyone feel free to add it back. I put `FIXME`'s whereever i disabled stuff.

### Describe the solution you'd like to see

- Enabling the theme toggle
- Sorting out all the theming and css for lightmode
- Fix issue where toggle wasn't selected by default (#436)

### Desribe alternate solutions

Only running darkmode, which isn't too bad at the beginning.

gabrielelpidio:
[#443](https://github.com/t3-oss/create-t3-app/pull/443) addresses this issue

PR #443, opened 2022-09-14
feat: added light mode back

gabrielelpidio:
Closes #<issue>

## ✅ Checklist

---

## Changelog

Addresses [#440](https://github.com/t3-oss/create-t3-app/issues/440)
- Fixed toggle initialization
- Fixed scrollbar on light mode
- Fixed code background on light mode
- Added darker color for hover on sidebar links

---

## Screenshots

image

### DIFF
--- a/www/astro.config.mjs
+++ b/www/astro.config.mjs
@@ -8,5 +8,11 @@ import tailwind from "@astrojs/tailwind";
  */
 export default defineConfig({
   site: `https://create.t3.gg`,
+  markdown: {
+    shikiConfig: {
+      theme: "poimandres",
+      wrap: true,
+    },
+  },
   integrations: [react(), tailwind()],
 });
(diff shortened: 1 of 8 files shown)


juliusmarminge:
Seems to be some bug on mobile introduced where the page has some space to the right?

image

I also think we should have lightmode codeblocks, Ideally we dont need to do all this ourself but just use some tool that lets us use VSCode themes or something? I know `shiki` has this.

gabrielelpidio:
Shiki is the [default](https://docs.astro.build/en/reference/configuration-reference/#markdownshikiconfig) syntax highlighter in astro. Adding a theme should be as easy as:
```js
export default defineConfig({
  // ...config
  markdown: {
    shikiConfig: {
      theme: "poimandres",
    },
  },
})
```
on `astro.config.mjs`

but having multiple syntax highlighting for dark and light mode doesn't seem to be totally resolved at the time of writing: 
Check out: https://github.com/shikijs/shiki/issues/33

So we could either use "css-variables" as the theme and pass some variables or just use something like the `poimandres` which looks relatively good on both light and dark mode

For reference
image

Also think full width code blocks look much better

image

Also Updated the PR with the recoommendations mentioned above

juliusmarminge (approved):
Awesome.

kroucher (approved):
Love it, awesome work @gabrielelpidio!

PR #447, opened 2022-09-15
refactor(docs): major refactor and configure theme

juliusmarminge:
Closes #<issue>

## ✅ Checklist

---

## Changelog

- Refactored the ThemeToggleButton cause it's logic was overcomplicated
  - May take a stab at TableOfContents too cause that's a mess
- Prepare Tailwind for a standardized colorscheme (haven't put any useful colors in there yet)
- Properly applied styles with `clsx` instead of ternaries
- Removed some unused stuff, redundant null-checks etc

---

## Screenshots

_[Screenshots]_

💯


www/tailwind.config.cjs
```diff
@@ -5,6 +5,16 @@ module.exports = {
...
       colors: {
+        base: "rgb(var(--color-base) / <alpha-value>)",
+        neutral: "rgb(var(--color-neutral) / <alpha-value>)",
+        primary: "rgb(var(--color-primary) / <alpha-value>)",
+        secondary: "rgb(var(--color-secondary) / <alpha-value>)",
+        accent: "rgb(var(--color-accent) / <alpha-value>)",
+        warning: "rgb(var(--color-warning) / <alpha-value>)",
+        error: "rgb(var(--color-error) / <alpha-value>)",
+        success: "rgb(var(--color-success) / <alpha-value>)",
+        info: "rgb(var(--color-info) / <alpha-value>)",
```
juliusmarminge, line 16: this will allow us to style things using `bg-warning`, `text-error/50` etc

www/src/styles/global.css
```diff
@@ -3,6 +3,30 @@
...
+    --color-neutral: 0 0 0;
+    --color-primary: 255 155 179;
+    --color-secondary: 111 114 185;
+    --color-accent: 255 155 179;
+    --color-warning: 255 155 179;
+    --color-error: 255 155 179;
+    --color-success: 255 155 179;
+    --color-info: 255 155 179;
+  }
+
```
juliusmarminge, line 29: need to put some nice color codes in here

c-ehrlich:
Other than specific colors (which should probably be a different PR) this looks good! Toggle button implementation is so much better than before

juliusmarminge:
> Other than specific colors (which should probably be a different PR) this looks good! Toggle button implementation is so much better than before

I did some basic colors that we can increment on: https://create-t3-9m2cxstvy-t3-oss.vercel.app/themeTest

Coyenn:
@juliusmarminge Nice work!
One thing I noticed is that the color of the GitHub button changes to black on Desktop screen sizes. 
Bildschirm­foto 2022-09-15 um 21 42 52

I have fixed the issue in this PR: 
https://github.com/t3-oss/create-t3-app/pull/450

juliusmarminge:
There may very likely be some UI glitches introduced when I deleted `index.css`, I didn't notice any straightaway so I don't think it did much tbf...

gabrielelpidio:
Light mode seems to be quite affected
image

www/src/components/headSeo.astro
```diff
@@ -20,27 +20,17 @@ const imageAlt = frontmatter.image?.alt ?? OPEN_GRAPH.image.alt;
 <meta property="og:title" content={formattedContentTitle} />
 <meta property="og:type" content="article" />
 <meta property="og:url" content={canonicalUrl} />
-<meta
-  property="og:locale"
-  content={frontmatter.ogLocale ?? SITE.defaultLanguage}
```
juliusmarminge, line 25: these checks came from the old astro config which didn't have any typesafety. we dont need it anymore

www/src/components/navigation/SidebarToggle.tsx
```diff
@@ -1,42 +1,26 @@
...
 
-  useEffect(() => {
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const body = document.querySelector("body")!;
-    if (sidebarShown) {
-      body.classList.add("mobile-sidebar-toggle");
-    } else {
-      body.classList.remove("mobile-sidebar-toggle");
-    }
-  }, [sidebarShown]);
```
juliusmarminge, line 19: this class wasn't used in any css

www/src/components/navigation/moreMenu.astro
```diff
@@ -1,85 +1,75 @@
...
 export interface Props {
   editHref: string;
 }
 
 const { editHref } = Astro.props;
-const showMoreSection = CONFIG.COMMUNITY_INVITE_URL;
 ---
 
 <div class="flex flex-col w-full items-center sm:items-start pt-8">
-  {
```
juliusmarminge, line 15: again, don't need to conditionally render these since we have the typesafety in place to warn if we were to remove these config options

www/src/components/navigation/moreMenu.astro
```diff
@@ -1,85 +1,75 @@
...
+        class="flex items-center gap-2 pl-2"
+        href={CONFIG.COMMUNITY_INVITE_URL}
+        target="_blank"
+      >
+        <svg
+          aria-hidden="true"
+          focusable="false"
+          data-prefix="fas"
+          data-icon="comment-alt"
+          class="svg-inline--fa fa-comment-alt fa-w-16"
```
juliusmarminge, line 58: curious what these icons are 🤔

www/src/styles/global.css
```diff
@@ -26,15 +52,7 @@
     }
   }
 
-  .mobile-sidebar-toggle {
```
juliusmarminge, line 29: there wasn't a single element using this class

juliusmarminge:
> You might have already seen it but just in case, light mode seems to be quite affected image

whoops, you feel like fixing it?

gabrielelpidio:
I'll try and fixing while solving the conflicts with #448

juliusmarminge:
> I'll try and fix it while solving the conflicts with #448

Seems like @c-ehrlich #449 fixed it.

We do have a lot of unused space here though if you want to take a look at that:
CleanShot 2022-09-16 at 00 10 20

juliusmarminge:
Mobile navbar toggle got broken too it seems like.

Also we should not lazy-load the themetoggle since it needs to load in order for the theme to apply

nexxeln (commented):
Looks good, there's some flash when I go to the docs page from the home page.

gabrielelpidio:
#453 solves some glitches, mobile menu and tries to improve the grid situation in the blog page

juliusmarminge:
> Looks good, there's some flash when I go to the docs page from the home page.

I think its cause the themetogglebutton which handles the theme isnt loaded immediately

Coyenn:
@juliusmarminge Some components on the landing page (including the buttons) suddenly have an underline on hover which looks weird.

PR to fix this: https://github.com/t3-oss/create-t3-app/pull/455
Note: There are still some components on the landing page (especially anchor tags. In the navbar for example) which I did not remove the underline from because it looks acceptable.

juliusmarminge:
Is this mergeable?

I dont like the white here:

image

gabrielelpidio:
The top part is kinda weird, I was having a go about it and has to do with [theme-color meta tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color). Sometimes it picks it up automatically from the header(?) but it's not constant. Also, the flashing from dark to light seems to be still happening. I think I can take a look at it later in the day or tomorrow

juliusmarminge:
> I think I can take a look at the issues later in the day or tomorrow

Check now

PR #462, opened 2022-09-17
fix: apply and use theme colors on landing page and algolia

Coyenn:
The theme "light" will now be applied to the landing page. This doesn't have much of an effect but it allows us to use the newly introduced theme colors there. I've styled the algolia search box using the new theme colors.
The twitter cards also now don't look out of place anymore.

The primary font family will be switched to Inter. If you do not like this change, please leave feedback below on why you oppose the idea.

## ✅ Checklist

- [x] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-08-15).
- [x] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] I performed a functional test on my final commit

PR #468, opened 2022-09-18
fix: footer background color

c-ehrlich:
## ✅ Checklist

---

## Changelog

Footer BG was a slightly different color than the remainder of the page in light mode. Actually it doesn't need a bg at all because it can just use that of the page.

---

## Screenshots

Before

Screenshot 2022-09-18 at 12 29 56

After

Screenshot 2022-09-18 at 12 32 45

💯


juliusmarminge:
It's fixed as part of #462

PR #476, opened 2022-09-18
refactor: de-react-ify

juliusmarminge:
Closes #<issue>

## ✅ Checklist

---

## Changelog

Some components seem to be React when they don't have to be.

---

## Screenshots

_[Screenshots]_

💯


www/src/components/headCommon.astro
```diff
@@ -14,25 +14,3 @@
     element.setAttribute("tabindex", "0");
   });
 </script>
-
```
juliusmarminge, line 17: when the themetoggler is no longer react we can move this part to the toggler itself which makes more sense imo

www/src/components/navigation/themeToggleButton.astro
```diff
@@ -0,0 +1,81 @@
+<div
+  class="flex bg-t3-purple-200/50 dark:bg-slate-700 p-2 w-fit mx-auto rounded-lg gap-3"
+>
+  <label class="cursor-pointer text-t3-purple-500 dark:text-slate-900">
```
juliusmarminge, line 4: i made both icons the same color, can change the sun back to gold if you don't like it

nexxeln: I like this more tbh

nexxeln (dismissed):
Great stuff

www/src/components/navigation/themeToggleButton.astro
juliusmarminge: yea me too, keeping it then :)

juliusmarminge:
should i try porting the TOC too? then it's only the typewriter left but i think that's fine unless we just go for a gif or smth? @nexxeln

nexxeln:
> should i try porting the TOC too? then it's only the typewriter left but i think that's fine unless we just go for a gif or smth? @nexxeln

Yeah sure

juliusmarminge:
@nexxeln I also made the change so that the H1 is "rendered" in the TOC without having to be explicitely defined as H1 and H2

PR #481, opened 2022-09-18
feat(prettier): add prettier-plugin-tailwindcss to website and cli 

### LINKED ISSUE #480: feat: add prettier-plugin-tailwindcss to cli and website

As of now classes are being added to both the cli and the code and it gets pretty messy sometimes for example:
```js
//Without prettier plugin
class="absolute flex flex-col mt-1 transform transition-transform duration-300 ease-in-out z-50 bg-default shadow shadow-t3-purple-100"

//With prettier plugin
class="bg-default shadow-t3-purple-100 absolute z-50 mt-1 flex transform flex-col shadow transition-transform duration-300 ease-in-out"
```

Add plugin to website and cli generated code

- Keeping it only for website
- Not adding the plugin

### Additional information

~~Important: this wouldn't add prettier as a formatter to ct3a, instead, the generated code would be pre-formatted with `prettier-plugin-tailwindcss`~~
We could consider adding it to the cli if the user chooses to use tailwind
Would love to have your thoughts, I'll leave a Draft PR down below

gabrielelpidio:
Closes #480 

## ✅ Checklist

---

## Changelog

- add prettier-plugin-tailwindcss


gabrielelpidio:
Changeset would need to be added if we want to merge as it is. But we could decide we don't want to merge it to cli

juliusmarminge:
I like it.

cli/src/installers/tailwind.ts
```diff
@@ -8,7 +8,13 @@ export const tailwindInstaller: Installer = async ({
   runPkgManagerInstall,
 }) => {
   await runPkgManagerInstall({
-    packages: ["tailwindcss", "postcss", "autoprefixer"],
+    packages: [
+      "tailwindcss",
+      "postcss",
+      "autoprefixer",
+      "prettier",
+      "prettier-plugin-tailwindcss",
```
juliusmarminge, line 16: I wonder if we should also add the eslint plugin or is that going too far?

juliusmarminge (commented):
Looks good. You do need to update your branch though, got some old stuff in there.

I also think we should merge my PR first for as little friction as possible?

juliusmarminge (commented):
Looks like this is in a mergeable state once updated. Why is it still a draft?

gabrielelpidio:
Have been pretty busy at work. I can take a look at it to update it later in the day

gabrielelpidio:
Updated the PR. Let me know if anything seems off

juliusmarminge (commented):
Don't we need to configure the plugin in our config too, as well as linking the tailwind config?

https://github.com/tailwindlabs/prettier-plugin-tailwindcss#installation

https://github.com/tailwindlabs/prettier-plugin-tailwindcss#resolving-your-tailwind-configuration

juliusmarminge:
@nexxeln What's your take on this, I know you were opposed to include prettier before.

Another option (for the app part) would be to add an option for it so it's opt-in.
- in that case "prettier" should be the option which would install prettier, and if tailwind is also selected install the plugin?

gabrielelpidio:
> Don't we need to configure the plugin in our config too, as well as linking the tailwind config?
> 
> https://github.com/tailwindlabs/prettier-plugin-tailwindcss#installation
> 
> https://github.com/tailwindlabs/prettier-plugin-tailwindcss#resolving-your-tailwind-configuration

Doesn't seem to be needed, I tan the formatting command as well as the VSCode extension for the prettier, and everything worked automatically

juliusmarminge:
> > Don't we need to configure the plugin in our config too, as well as linking the tailwind config?
> > 
> > https://github.com/tailwindlabs/prettier-plugin-tailwindcss#installation
> > 
> > https://github.com/tailwindlabs/prettier-plugin-tailwindcss#resolving-your-tailwind-configuration
> 
> Doesn't seem to be needed, I ran the formatting command as well as the VSCode extension for prettier, and everything worked automatically 

And it formats on save? That seems to be the thing breaking the most 🤔

nexxeln:
> Don't we need to configure the plugin in our config too, as well as linking the tailwind config?
> 
> https://github.com/tailwindlabs/prettier-plugin-tailwindcss#installation
> 
> https://github.com/tailwindlabs/prettier-plugin-tailwindcss#resolving-your-tailwind-configuration

Yeah I think this needs to be done.

nexxeln:
> @nexxeln What's your take on this, I know you were opposed to include prettier before.
> 
> Another option (for the app part) would be to add an option for it so it's opt-in.
> 
> * in that case "prettier" should be the option which would install prettier, and if tailwind is also selected install the plugin?

I'm still against it. A lot of people use Headwind. And if someone wants prettier they can just install it. I think if we just have this in the website it makes sense but not the CLI.

The best thing we can do imo is to mention these extensions in the docs for tailwind.

juliusmarminge:
> I'm still against it. A lot of people use Headwind. And if someone wants prettier they can just install it.

People who prefer Headwind would just be able to not select prettier I guess 🤷🏼‍♂️

> I think if we just have this in the website it makes sense but not the CLI.

Definetely for our repo, so that the classes are sorted from the beginning of a scaffold, as well as for the docs we should have it. How about we just do that for now and pull an issue up for discussion about whether or not to include it in the CLI too?

nexxeln:
> > I'm still against it. A lot of people use Headwind. And if someone wants prettier they can just install it.
> 
> People who prefer Headwind would just be able to not select prettier I guess 🤷🏼‍♂️
> 
> > I think if we just have this in the website it makes sense but not the CLI.
> 
> Definetely for our repo, so that the classes are sorted from the beginning of a scaffold, as well as for the docs we should have it. How about we just do that for now and pull an issue up for discussion about whether or not to include it in the CLI too?

Alright

juliusmarminge:
Struggling to get the plugin to play nice with Astro? It doesn't format on save at least, and I'm unsure if it does even on `pnpm format`? I made some change and it didn't fix it 🤷🏼‍♂️

www/src/components/landingPage/about.astro
```diff
@@ -8,7 +8,7 @@ import PageSection from "./pageSection.astro";
   size={"24"}
   id="about"
 >
-  <div class="flex flex-col gap-2 md:gap-4 max-w-[800px] xl:max-w-full mx-auto">
+  <div class="flex gap-2 md:gap-4 max-w-[800px] xl:max-w-full mx-auto flex-col">
```
juliusmarminge, line 11: this change doesn't get formatted when I changed it :/

juliusmarminge:
Seems like they're waiting for a Tailwind release.

juliusmarminge (approved):

gabrielelpidio:
Yes, I must have missed that, remembering, seems that I tested the autosave only with the `.tsx` files, However, some .astro files were formatted when I ran `pnpm format`. I'll investigate further into this issue

PR #492, opened 2022-09-21
feat(ci): conditionally run prerelease workflow

### LINKED ISSUE #487: feat(ci): restrict ci prerelease workflow to only conditionally

The [prerelease](https://github.com/t3-oss/create-t3-app/blob/main/.github/workflows/prerelease.yml) runs on every push to main. This is a bit redundant for docs and other miscellaneous changes that doesn't have any impact on the CLI.

Check for relevant changes before running the workflow to reduce versions published to npm.

Run it in CI if the `area: t3-app` or `area: cli` is present, so that we could test beta before merging important PRs. A second workflow would then be needed to comment the version to use in npm, similar to cloudflare's: https://github.com/cloudflare/wrangler2/blob/main/.github/workflows/write-prerelease-comment.yml

### Additional information

Could also modify the version tag of the beta release to be a proper version tag that follows semver (like 5.13.0-beta.1) instead of just hardcoding 6.0.0 and appending the commit hash.

GRob27:
Closes #487 

## ✅ Checklist

---

## Changelog

Check that labels contain `area: t3-app`  or  `area: cli`
Change script to run on pull requests instead of push because labels are not available on push


juliusmarminge:
thanks. looks promising, hope it works.

juliusmarminge:
I came to think of another way, we can run it if this label is added, that way we can manually choose what PRs we want a beta release for since there isn't that many which actually needs it, whatya think?

GRob27:
I think this is better since it gives the possibility to opt in or out of the prerelease. Will do this modification!

juliusmarminge:
> I think this is better since it gives the possibility to opt in or out of the prerelease. Will do this modification!

You think you'll also be able to add the other bit of the issue?

GRob27:
> > I think this is better since it gives the possibility to opt in or out of the prerelease. Will do this modification!
> 
> You think you'll also be able to add the other bit of the issue?

Will look into it!

juliusmarminge:
> > > I think this is better since it gives the possibility to opt in or out of the prerelease. Will do this modification!
> 
> > 
> 
> > You think you'll also be able to add the other bit of the issue?
> 
> 
> 
> Will look into it!

Awesome, dw if you can't get it done, we can just leave it open

GRob27:
> Additional information
Could also modify the version tag of the beta release to be a proper version tag that follows semver (like 5.13.0-beta.1) instead of just hardcoding 6.0.0 and appending the commit hash.

Currently wondering on the best way to approach this.
So far my solution would look something like this, however i could not find find a good and solid way to get the new beta version (to replace  the commit hash)
`pkg.version = pkg.version + "-beta." + stdout.trim();`

juliusmarminge:
> > Additional information
> 
> Could also modify the version tag of the beta release to be a proper version tag that follows semver (like 5.13.0-beta.1) instead of just hardcoding 6.0.0 and appending the commit hash.
> 
> 
> 
> Currently wondering on the best way to approach this.
> 
> So far my solution would look something like this, however i could not find find a good and solid way to get the new beta version (to replace  the commit hash)
> 
> `pkg.version = pkg.version + "-beta." + stdout.trim();`

dw bout it I would say. I think since we went with running in PRs just adding a comment with the command to use is enough?

.github/workflows/prerelease-comment.yml
```diff
@@ -0,0 +1,52 @@
...
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed
+
+jobs:
+  comment:
+    if: |
+      github.repository_owner == 't3-oss' &&
+      ${{ github.event.workflow_run.conclusion == 'success' }}
```
juliusmarminge, line 13: will this comment on PRs even if a new version hasn't been published?

GRob27: This would run after the pre-release workflow, I assumed if we reached here it would always have a new version. We coud explicitly fail the previous step to make sure to never run the comment if there is not a new version

juliusmarminge: At the top it says runs after completed CI?

GRob27: That should not be there, fixing

juliusmarminge: Ah great 👍

GRob27: That was the previous step in my mock CI

juliusmarminge:
Let's test it!

juliusmarminge (approved):

PR #502, opened 2022-09-25
fix: nextauth github missing database field

axcellence:
Closes #<issue>

## ✅ Checklist

---

## Changelog

GitHub sign in requires `refresh_token_expires_in` present in the Account table. 
Added to the `refresh_token_expires_in` Prisma model for Account

---

## Screenshots

CleanShot 2022-09-25 at 12 11 08@2x

💯

### DIFF
--- a/.changeset/giant-bugs-argue.md
+++ b/.changeset/giant-bugs-argue.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+GitHub signin requires `refresh_token_expires_in` present in the Account table. Added to the prisma model
--- a/cli/template/addons/prisma/auth-schema.prisma
+++ b/cli/template/addons/prisma/auth-schema.prisma
@@ -26,6 +26,7 @@ model Account {
     provider          String
     providerAccountId String
     refresh_token     String? //@db.Text
+    refresh_token_expires_in Int?
     access_token      String? //@db.Text
     expires_at        Int?
     token_type        String?


c-ehrlich:
I'm not in favor of this, assuming we keep Discord as the default provider.

Reason is that if you go through the next-auth provider docs, _most_ of them require some additional work compared to the very simple Discord provider. How many of these do we want shoehorn into a single template? IMO it's fine to give one example, and have people extend it to fit whatever other provider they want to use.

That being said, I could see a case for changing the default provider to GitHub, as the average CT3A user is more likely to have a GitHub account than a Discord account.

axcellence:
> I'm not in favor of this, assuming we keep Discord as the default provider.
> 
> Reason is that if you go through the next-auth provider docs, _most_ of them require some additional work compared to the very simple Discord provider. How many of these do we want shoehorn into a single template? IMO it's fine to give one example, and have people extend it to fit whatever other provider they want to use.
> 
> That being said, I could see a case for changing the default provider to GitHub, as the average CT3A user is more likely to have a GitHub account than a Discord account.

That's a fair point

juliusmarminge:
> I'm not in favor of this, assuming we keep Discord as the default provider.
> 
> 
> 
> Reason is that if you go through the next-auth provider docs, _most_ of them require some additional work compared to the very simple Discord provider. How many of these do we want shoehorn into a single template? IMO it's fine to give one example, and have people extend it to fit whatever other provider they want to use.
> 
> 
> 
> That being said, I could see a case for changing the default provider to GitHub, as the average CT3A user is more likely to have a GitHub account than a Discord account.

Github is kinda whack imo cause you need one app for dev and one for prod (you can't have multiple callback URLs). Discord is a breeze and I think that's why we changed to have it as default

juliusmarminge:
We can annotate in the docs that you should read up on the providers you are adding I guess 🤔

PR #518, opened 2022-09-29
docs: update prisma section on why page based on #402 feedback

bdsqqq:
## ✅ Checklist

---

## Changelog

- Move TS comparison to the start;
- Added note about compatibility with several databases, linked to the connectors doc page
- Added note about the main focus being DX with the tools Prisma brings as secondary
- Minor improvements to the reading flow

---

## Screenshots

image

💯


bdsqqq:
#402 was closed and I don't know how to keep this going there.
I applied most of the feedback, but don't know how to talk about Prisma studio

nexxeln (commented):
I think leave out the "notably" and add "and the Prisma Studio which gives a tabular interface to easily manipulate data quickly."

Something along these lines?

bdsqqq:
Amazing! I added the prisma studio string you suggested with minor tweaks, thanks!

The notably is there to say "there's a bunch of features, these are usually the ones you should take a look at", I don't think the phrase would work the same way without it or an equivalent. 
But I agree it feels a bit out of place, I made the paragraph this way to keep the flow of "This is the main point, this is secondary, these are the most important in the secondary", but if we were to make it "this is the main point, these are interesting, this is the rest if you really wanna dive into it" we could drop the notably. 

What you think? I'm down to rewrite it with another flow

nexxeln:
> Amazing! I added the prisma studio string you suggested with minor tweaks, thanks!
> 
> The notably is there to say "there's a bunch of features, these are usually the ones you should take a look at", I don't think the phrase would work the same way without it or an equivalent. But I agree it feels a bit out of place, I made the paragraph this way to keep the flow of "This is the main point, this is secondary, these are the most important in the secondary", but if we were to make it "this is the main point, these are interesting, this is the rest if you really wanna dive into it" we could drop the notably.
> 
> What you think? I'm down to rewrite it with another flow

I like this but also curious of what your rewrite will look like.

juliusmarminge (approved):
ship ittt

juliusmarminge:
> > Amazing! I added the prisma studio string you suggested with minor tweaks, thanks!
> > The notably is there to say "there's a bunch of features, these are usually the ones you should take a look at", I don't think the phrase would work the same way without it or an equivalent. But I agree it feels a bit out of place, I made the paragraph this way to keep the flow of "This is the main point, this is secondary, these are the most important in the secondary", but if we were to make it "this is the main point, these are interesting, this is the rest if you really wanna dive into it" we could drop the notably.
> > What you think? I'm down to rewrite it with another flow
> 
> I like this but also curious of what your rewrite will look like.

I'll merge as is, feel free to continue if you have something even better 👍🏼

PR #523, opened 2022-09-29
feat(seo): added op-image

JARofStuff:
Closes #<issue>

## ✅ Checklist

---

## Changelog

- Added op:image
- Set up HeadSeo component to work on both layouts (with and without Frontmatter)
- Set 'site' in astro.config.mjs to be current working url (https://beta.create.t3.gg)
- Removed astro-seo from LandingPage layout
- Removed unnecessary props from landing page (they were the same as the defaults set in config.ts)


.changeset/spicy-carpets-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+Added op:image, set up headSeo for correct usage. Set config site to 'beta.create.t3.gg'
```
juliusmarminge, line 5: Not needed for docs changes.

Changesets only for the CLI

www/src/layouts/landingPage.astro
```diff
@@ -21,7 +17,8 @@ const currentPage = Astro.url.pathname;
 <html dir="ltr" lang="en-us" class="initial">
   <head>
     <HeadCommon />
-    <SEO title={title} description={description} />
```
juliusmarminge, line 24: Can remove it as dependency too

.changeset/spicy-carpets-happen.md
JARofStuff: Ok, removed the file.

www/src/layouts/landingPage.astro
JARofStuff: Removed

juliusmarminge (approved):
Great

PR #529, opened 2022-10-01
feat: tailor .env to only include relevant keys

### LINKED ISSUE #515: feat: tailor .env to only include relevant keys

5.13.1

### Describe the bug

When scaffolding an app, without next-auth there's still pieces added for it in the `.env` file provided.
```# Next Auth
NEXTAUTH_SECRET=""
NEXTAUTH_URL=http://localhost:3000


# Next Auth Discord Provider
DISCORD_CLIENT_ID=
DISCORD_CLIENT_SECRET=
```
Not sure if it's meant to be there or if it's a hassle to get out when not needed.

### To reproduce

Make an app, check `.env`.

### Additional information

_No response_

CDE90:
Closes #515 

## ✅ Checklist

---

## Changelog

Tailor the .env file to only include the relevant keys.

---

## Screenshots

💯


juliusmarminge:
I think we can do better. We don't need to keep duplicates of this and can generate one dynamically instead. Something like this should be good.

```ts
let env = ""
if (packages.prisma.inUse) {
  env += `
DATABASE_URL=file://dev.db
`
}
if (packages.nextAuth.inUse) {
  env += `
NEXTAUTH_URL=http://localhost:3000
...
`
}
...
fs.writeFileSync(PKG_ROOT, env, "UTF-8");
```

The env-keys could even be part of the installers or something so that it is easily extendible?

CDE90:
> I think we can do better. We don't need to keep duplicates of this and can generate one dynamically instead. Something like this should be good.
> 
> ```ts
> let env = ""
> if (packages.prisma.inUse) {
>   env += `
> DATABASE_URL=file://dev.db
> `
> }
> if (packages.nextAuth.inUse) {
>   env += `
> NEXTAUTH_URL=http://localhost:3000
> ...
> `
> }
> ...
> fs.writeFileSync(PKG_ROOT, env, "UTF-8");
> ```
> 
> The env-keys could even be part of the installers or something so that it is easily extendible?

I've changed it to something like this now, does it look ok?

Also I'm not entirely sure whether this should be classed as a minor change or a patch, so advice on that would be appreciated

juliusmarminge:
> Also I'm not entirely sure whether this should be classed as a minor change or a patch, so advice on that would be appreciated

patch is fine I think

PR #533, opened 2022-10-01
feat: tailor .env to only include relevant keys (#529)

juliusmarminge:
* feat: tailor .env to only include relevant keys

* chore: add changeset

* refactor: generate env file dynamically

Original PR #529

Testing beta release


juliusmarminge:
Works great @CDE90

PR #557, opened 2022-10-04
feat: preprocess nextauth url

### LINKED ISSUE #535: feat: mark NEXTAUTH_URL as optional in environment variable schema

The default behavior of next auth is to fallback to VERCEL_URL when NEXTAUTH_URL is unavailable.
https://next-auth.js.org/configuration/options#nextauth_url

Marking NEXAUTH_URL as optional will prevent the need to set NEXTAUTH_URL when deploying to Vercel.

The new dev experience debugging the schema errors would generally lead someone adding the primary application URL to NEXTAUTH_URL, which means they need to juggle env vars for different environments and lose the benefits of the provided VERCEL_URL variable.

``` ts
...
  NEXTAUTH_URL: z.string().url().optional(),
});
```

Could instead add a comment suggesting the use of .optional if the plan is to deploy to vercel. This would make it slightly less opinionated.

It also may be possible to check the presence of either VERCEL_URL or NEXTAUTH_URL, but I'm not sure how to do that myself.

### Additional information

It may also be a good idea to add a comment to the example .env file explaining that it does not need to be set on Vercel, but may need to be on other hosting platforms.

juliusmarminge:
Closes #535

## ✅ Checklist

---

## Changelog

Uses Zod preprocess to default url to VERCEL_URL if it exists

Needs testing before merging

---

## Screenshots

_[Screenshots]_

💯


juliusmarminge:
Seems like the VERCEL_URL environment variable isn't actually an URL at build-time:

CleanShot 2022-10-06 at 09 32 09@2x

Changing the validation schema to 
```ts
  NEXTAUTH_URL: z.preprocess((str) => process.env.VERCEL_URL ?? str, z.string()),
```
works, but will obviously not validate that the URL you manually pass in for DEV or other deployments is valid - any string will do.

This is the VERCEL_URL for an example deployment using the above schema:

CleanShot 2022-10-06 at 09 34 45@2x

which I think doesn't validate as an url cause its missing `http(s)://`.

I think the tradeoff is fine, in that we don't validate that it's a valid URL if it means we can auto-fill the VERCEL_URL. What's think?

c-ehrlich:
Would it be possible to use a different validator depending on the NODE_ENV?

c-ehrlich:
> > Would it be possible to use a different validator depending on the NODE_ENV?
> 
> Sure, we could do something like this to make sure that it's a valid URL, but if we're on Vercel, just a string is fine. Not sure what approach is best 🤷🏼‍♂️
> 
> https://github.com/juliusmarminge/preprocess/blob/35b936c1710676c37ddd8ac04bf4e91379431f11/src/env/schema.mjs#L12-L15
> 
> This would make it so your build fails if its not a valid url, but accept any string on vercel

that seems reasonable. maybe add a comment to explain what's going on here, don't want to scare newcomers away from zod 😅

juliusmarminge:
Yea should prob update docs too.  Will do later

c-ehrlich (approved):
looks good with the comment :)

c-ehrlich (approved):

PR #582, opened 2022-10-06
feat: remove prettier from base

juliusmarminge:
Closes #<issue>

## ✅ Checklist

---

## Changelog

No real need to install prettier in the base application - goes against the `you can npm install it` so moved it to the tailwind installer so it's only installed alongside the tailwind-plugin.

---

## Screenshots

_[Screenshots]_

💯


c-ehrlich (approved):
Makes much more sense than what we had before

PR #587, opened 2022-10-07
feat: add .env-example

### LINKED ISSUE #585: feat: make env vars optional

I pulled my t3 app on my laptop at an internet cafe (I wanted to do a little UI work). Trying to run it locally, I'm confronted with 
❌ Invalid environment variables:
 DATABASE_URL: Required
 NEXTAUTH_SECRET: Required
 NEXTAUTH_URL: Required
 DISCORD_CLIENT_ID: Required
 DISCORD_CLIENT_SECRET: Required

I created a `.env` and stuck it in my src folder with those variables with empty values

Still doesn't work.

I then added `optional` to them here
```
    DATABASE_URL: z.string().url().optional(),
    NODE_ENV: z.enum(["development", "test", "production"]),
    NEXTAUTH_SECRET: z.string().optional(),
    NEXTAUTH_URL: z.string().url().optional(),
    DISCORD_CLIENT_ID: z.string().optional(),
    DISCORD_CLIENT_SECRET: z.string().optional(),
})
```

And that fixed the problem.

I believe these env vars should be optional from the start. People who are just starting out my just want to be focusing on their frontends and not have to worry about environment issues.

You probably have a better solution than the one I proposed. 

Just make source buildable out-of-the-box if someone were to push to their remote, then pull to another machine.

### Additional information

_No response_

c-ehrlich:
Closes #585

## ✅ Checklist

---

IIRC we only generated `.env-example` a while ago, and changed it to `.env` so that the app could be immediately started after scaffolding. Will close this PR if anyone has a good reason why we shouldn't include both `.env` and `.env-example` by default.

---

## Changelog

- Creates a `.env-example` file with the same content as the normal `.env` file plus a warning about not putting secrets in there.
- Moves the `if (!envFile) return;` guard clause up a few lines to prevent pointlessly generating the envContent before not saving it to anywhere.

---

## Screenshots

image

💯

### DIFF
--- a/.changeset/tender-ducks-mate.md
+++ b/.changeset/tender-ducks-mate.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": minor
+---
+
+Add .env-example to scaffolded app
--- a/cli/src/installers/envVars.ts
+++ b/cli/src/installers/envVars.ts
@@ -25,6 +25,8 @@ export const envVariablesInstaller: Installer = ({ projectDir, packages }) => {
       break;
   }
 
+  if (!envFile) return;
+
   if (usingPrisma) {
     envContent += `
 # Prisma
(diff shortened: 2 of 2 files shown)


nexxeln (approved):

juliusmarminge:
Should be added to the docs too perhaps?

PR #590, opened 2022-10-08
fix(template): remove semicolon in _app of with-auth template

nexxeln:
## ✅ Checklist

---

## Changelog

Man I thought I had fixed this but somehow this semicolon came again here.

---


c-ehrlich (approved):
Good catch. I think it's caused by Prettier.

juliusmarminge:
Could have been fixed on main but not transferred over to next

nexxeln:
> Could have been fixed on main but not transferred over to next

Yeah probably.

PR #596, opened 2022-10-10
chore: Made <details> tags pretty 

dayvster:
Closes #<issue>

## ✅ Checklist

---

## Changelog

Styled the details tags, added an animation script included it in the docs `pageContent.tsx` made sure that it loads in async not to block the page load.

---

## Screenshots

image

image

💯


www/public/animateDetails.js
```diff
@@ -0,0 +1,117 @@
+/**
```
c-ehrlich, line 1: The functionality seems good based on the preview deployment, but I don't love the idea of inlining a .js file. That being said I've never implemented something like this. Is this how other projects do it as well?

dayvster: @c-ehrlich I suppose the JS code could simply be moved directly into `pageContent.astro` and only refactored out when/if it needs to be re-used somewhere else?

nexxeln: > @c-ehrlich I suppose the JS code could simply be moved directly into `pageContent.astro` and only refactored out when/if it needs to be re-used somewhere else?

Yeah I would like if it was in an astro file, or just a file in `src/utils` or whatever. Putting a js file in public feels weird.

dayvster: @c-ehrlich @nexxeln refactored it from `public/animateDetails.js` into `pageContent.astro` should we mark this convo as resolved?

nexxeln:
CI is failing for some reason

dayvster:
> CI is failing for some reason

Noticed, it expects the script in pageContent to be typescript and is complaining about missing types.

juliusmarminge:
> > CI is failing for some reason
> 
> 
> 
> Noticed, it expects the script in pageContent to be typescript and is complaining about missing types. 

You can add the `is:inline` directive or make it properly typed

dayvster:
> > > CI is failing for some reason
> > 
> > 
> > Noticed, it expects the script in pageContent to be typescript and is complaining about missing types.
> 
> You can add the `is:inline` directive or make it properly typed

Done and done.
 
> EDIT: but as @c-ehrlich pointed out, is all this JavaScript needed to make it animate the "fold"?

Sadly I'd say yes, I tried doing it with pure CSS/tailwind at first and it didn't quite work.

www/public/animateDetails.js
dayvster: @nexxeln @c-ehrlich refactored from `public/animateDetails.js` into `docs/pageContent.astro` as requested.

juliusmarminge:
> > > > CI is failing for some reason
> > > 
> > > 
> > > Noticed, it expects the script in pageContent to be typescript and is complaining about missing types.
> > 
> > 
> > You can add the `is:inline` directive or make it properly typed
> 
> Done and done.
> 
> > EDIT: but as @c-ehrlich pointed out, is all this JavaScript needed to make it animate the "fold"?
> 
> Sadly I'd say yes, I tried doing it with pure CSS/tailwind at first and it didn't quite work.

Alright Lighthouse Performance still 100 so all good :)

Maybe leave a FIXME comment to make that script TypeScript sometime in the future?

www/src/components/docs/pageContent.astro
```diff
@@ -101,3 +101,122 @@ const title = frontmatter.title;
     block.appendChild(copyButton);
   });
 </script>
+<script is:inline>
```
juliusmarminge, line 104:
```suggestion
<script is:inline>
// FIXME: We can make this properly typed later
```

www/src/components/docs/pageContent.astro
```diff
@@ -101,3 +101,123 @@ const title = frontmatter.title;
...
+// FIXME: We can make this properly typed later
```
juliusmarminge, line 105:
```suggestion
  //FIXME: We can make this properly typed later
```

juliusmarminge (approved):
Fancy :)

nexxeln (approved):

PR #609, opened 2022-10-14
add support for both v10 and v9 trpc

OrJDev:
Closes #<issue>

## ✅ Checklist

---

## Changelog

When using tRPC user will be requested to select trpc version (or can provide it as arg using --trpc <trpcversion>) - supported versions are 10 and 9 ofc.

---

## Screenshots

will be asked when the version is not mentioned. (--trpc <10 | 9>)
Screenshot_1

💯


juliusmarminge:
Appreciate the work you've put into this but wish you had opened an issue first so all this work wouldn't have been "for nothing".

We've had this brought up before, and before I created the `next` branch I made a PR similar to this which we decided against.

Maintaining two branches wasn't a walk in the park, it took quite a bit of extra time so I'm happy we feature locked main and only
maintain the v10 nowadays. If users still want to start of their new projects using tRPC v9 (which I wouldn't recommend), they can use `npx create-t3-app@5`.

OrJDev:
> Appreciate the work you've put into this but wish you had opened an issue first so all this work wouldn't have been "for nothing".
> 
> We've had this brought up before, and before I created the `next` branch I made a PR similar to this which we decided against.
> 
> Maintaining two branches wasn't a walk in the park, it took quite a bit of extra time so I'm happy we feature locked main and only
> maintain the v10 nowadays. If users still want to start of their new projects using tRPC v9 (which I wouldn't recommend), they can use `npx create-t3-app@5`.

Its ok, I was aware of the main branch but I was bored and thought it would be better to have just one branch instead of committing changes to both the next branch and the main branch.

c-ehrlich:
Agree with Julius - I can see this took a lot of work, and skimming the code it looks quite well implemented. But I don't think we should merge it.

We want to be opinionated, and one of those opinions is that v10 is now good enough that we don't want to openly offer anything else. If someone wants v9, they can use an older version of ct3a or set it up themselves, but I don't think we want to be the ones leading them down that path.

Again the implementation itself seems good to me, and I hope you won't be discouraged from contributing in the future :)

PR #615, opened 2022-10-16
fix: missing characters in homepage cli on android

c-ehrlich:
## ✅ Checklist

---

## Changelog

Currently, the cli simulation has missing characters on some Android devices (possibly also other OSs?) in the 'A' of the ascii text. This fixes it.

---

## Screenshots

Apologies for the photos, I just borrowed this Android device and don't have anything set up on it

Before:
IMG_6573 Medium

After:
IMG_6572 Medium

💯

### DIFF
--- a/www/src/components/headCommon.astro
+++ b/www/src/components/headCommon.astro
@@ -1,5 +1,5 @@
 <!-- Global Metadata -->
-<meta charset="utf-8" />
+<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
 <meta name="viewport" content="width=device-width" />
 <meta name="generator" content={Astro.generator} />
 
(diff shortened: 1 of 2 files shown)


www/src/components/headCommon.astro
```diff
@@ -1,5 +1,5 @@
 <!-- Global Metadata -->
-<meta charset="utf-8" />
+<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
```
nexxeln, line 2: What does this do?

nexxeln (approved):
Looks great

www/src/components/headCommon.astro
c-ehrlich: enforces the charset for super-old browsers that ignore the newer `charset` feature. on second thought maybe its safe to assume that users are not on something like ie8

will change this back, at least until there is a report of someone hitting this issue on an actual device they're using

nexxeln (approved):

PR #642, opened 2022-10-23
fix: next-auth doesn't install when using node 18

c-ehrlich:
## ✅ Checklist

---

## Changelog
- If the user is on node 18 and chooses to install during cli, `--ignore-engines true` (yarn) or `--engines-strict false` (npm or pnpm) is used
- If next-auth is selected, `.npmrc` and `.yarnrc` files with the same flags are created, regardless of what node version the user is currently on
- FIXME's linking to [the next-auth issue](https://github.com/nextauthjs/next-auth/issues/4575) so that we can hopefully get rid of this again ASAP. I'm also subscribed to the issue.
- Tested with current versions of npm, yarn classic, and pnpm

I don't love doing this and could also see the argument for just leaving next-auth not working on node 18 until they release a fix, but personally I'm leaning towards this (not great) solution for the time being. 

I'd also appreciate a Windows user confirming that this works before we merge.

💯

### DIFF
--- a/cli/src/helpers/installDependencies.ts
+++ b/cli/src/helpers/installDependencies.ts
@@ -9,7 +9,21 @@ export const installDependencies = async (projectDir: string) => {
   const pkgManager = getUserPkgManager();
   const spinner = ora(`Running ${pkgManager} install...\n`).start();
 
-  await execa(pkgManager, ["install"], { cwd: projectDir });
+  // FIXME: temp fix for next-auth with node 18
+  // see: https://github.com/nextauthjs/next-auth/issues/4575
+  if (process.versions.node.startsWith("18")) {
+    if (pkgManager === "yarn") {
+      await execa(pkgManager, ["add", "--ignore-engines", "true"], {
+        cwd: projectDir,
+      });
+    } else {
+      await execa(pkgManager, ["install", "--engine-strict", "false"], {
+        cwd: projectDir,
+      });
+    }
+  } else {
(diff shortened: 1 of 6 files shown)


juliusmarminge:
> I don't love doing this and could also see the argument for just leaving next-auth not working on node 18 until they release a fix, but personally I'm leaning towards this (not great) solution for the time being.

Since Node 18 will be LTS soon, I think we should definitely support it. Just a shame you have to workarounds to get certain libraries to follow along :(

cli/src/helpers/installDependencies.ts
```diff
@@ -9,7 +9,21 @@ export const installDependencies = async (projectDir: string) => {
   const pkgManager = getUserPkgManager();
   const spinner = ora(`Running ${pkgManager} install...\n`).start();
 
-  await execa(pkgManager, ["install"], { cwd: projectDir });
+  // FIXME: temp fix for next-auth with node 18
+  // see: https://github.com/nextauthjs/next-auth/issues/4575
+  if (process.versions.node.startsWith("18")) {
```
juliusmarminge, line 14:
```suggestion
  if (
    process.versions.node.startsWith("18") ||
    process.versions.node.startsWith("19")
  ) {
```

juliusmarminge: I'm guessing the same goes for 19 so we should check that too since that's now a thing

nexxeln (approved):
works

Issue #603, opened 2022-10-11
feat: improve docker setup


Nsttt:
### Is your feature request related to a problem? Please describe.

Currently in the docs it states that I should remove the `import { env } from "./src/env/server.mjs";` when deploying with Docker. 

I believe this work as a placeholder and there should be some sort of guide / system that allows users an easy-going experience when deploying with Docker.

Maybe some pre-defined files in the templates would be good, Or just docs are enough.

### Describe the solution you'd like to see

An easy-going setup with docker.

Allow for different configs in `dev` or `prod` and guide the user to deploy the app.

### Desribe alternate solutions

Probably split the config files into 2 one for prod one for development?

### Additional information

I'm looking into this at the time of writting this, so any progress I'll update it here.

Any opinions and feedback is much appreciated.

c-ehrlich:
I keep forgetting to write the docs on how to set up a docker dev setup with HMR etc. But obviously that alone isn't good enough - the real goal would be to be able to develop locally and deploy to docker, without having to use different branches or some other hack.

If this requires a complex setup (ie multiple config files etc) I think we should not include it in the template, but give instructions in the docs.

Curious to see what you come up with :)

Nsttt:
I've made some progress on this over the week. 

So far I've managed to dinamically import the enviroment into the `next.config.mjs`. To do this, in order to use top-level await I have changed the target in tsconfig to `es2017` **still unsure about this and I would like some feedback.**

Following, there should also be a few more changes into the docs pages explaining.
- How to setup and prod enviroment and everything to keep in mind on settings.
- How to setup and dev enviroment.
- Probably a more detailed view of working with enviromental variables ?

Anything else you feel is needed please add below. I'll make a PR with the changes soon. 😄

c-ehrlich:
What does targeting es2017 mean in this context? It's at 96% on caniuse, so would it mean incompatibility with 4% of web users?

Nsttt:
To your first question, I believe 

To your second question, to my knowledge, moving away from `es5` loses compatibility with IE11, which I don't find much of an issue and not much else. Change, allows us to use features from es6, so in the `next.config.mjs`, we can add the following snippet.

```ts
process.env.NODE_ENV === "development"
  ? await import("./src/env/server.mjs")
  : null;
```

And so, get a conditionally imported `env`.

juliusmarminge:
Resolved in #646

PR #646, opened 2022-10-23
feat: use top-level await to import server envs

Nsttt:
Related to #603 

## ✅ Checklist

---

## Changelog

- Use top-level await in the `next.config.mjs`.
- Change tsconfig target to es2017.

💯

### DIFF
--- a/cli/template/base/next.config.mjs
+++ b/cli/template/base/next.config.mjs
@@ -1,5 +1,7 @@
 // @ts-check
-import { env } from "./src/env/server.mjs";
+process.env.NODE_ENV === "development"
+  ? await import("./src/env/server.mjs")
+  : null;
 
 /**
  * Don't be scared of the generics here.
--- a/cli/template/base/next.config.mjs
+++ b/cli/template/base/next.config.mjs
@@ -1,7 +1,6 @@
 // @ts-check
-process.env.NODE_ENV === "development"
-  ? await import("./src/env/server.mjs")
-  : null;
+/* run the build with this set to skip validation */
+!process.env.SKIP_ENV_VALIDATION && (await import("./src/env/server.mjs"));
 
 /**
  * Don't be scared of the generics here.
--- a/www/src/pages/en/deployment/docker.md
+++ b/www/src/pages/en/deployment/docker.md
@@ -50,7 +50,7 @@ README.md
 
 ### 3. Create Dockerfile
 
-> In order to not pull the server enviroment variables to our container, we have to add a`SKIP_ENV_VALIDATION=1`flag to the build command.
+> Since we're not pulling the server environment variables into our container, the [environment schema validation](/en/usage/environment-variables) will fail. To prevent this, we have to add a `SKIP_ENV_VALIDATION=1` flag to the build command so that the env-schemas aren't validated at build time.
 
 <details>
     <summary>
--- a/www/src/pages/en/deployment/docker.md
+++ b/www/src/pages/en/deployment/docker.md
@@ -25,15 +25,7 @@ export default defineNextConfig({
 });
 ```
 
-### 2. Remove Env Import
-
-Remove the `env`-import from [`next.config.mjs`](https://github.com/t3-oss/create-t3-app/blob/main/cli/template/base/next.config.mjs) so it isn't pulled into the build image:
-
-```diff
-- import { env } from "./src/env/server.mjs";
-```
-
-### 3. Create dockerignore file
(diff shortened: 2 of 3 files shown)


cli/template/base/next.config.mjs
```diff
@@ -1,5 +1,7 @@
 // @ts-check
-import { env } from "./src/env/server.mjs";
+process.env.NODE_ENV === "development"
+  ? await import("./src/env/server.mjs")
+  : null;
```
juliusmarminge, line 4: Not a fan of this, if you're not using docker, you'd likely want the build to fail if the envs aren't validated. Either add this snippet to the docker docs, or use another env var that could cancel it?

juliusmarminge (changes requested):

cli/template/base/next.config.mjs
Nsttt: Very true, wasn't thinking of that. I'll think how can that be improved.

juliusmarminge: with another var I mean something like

```ts
/* run the build with this set to skip validation */
!process.env.SKIP_ENV_VALIDATION
  && await import("./src/env/server.mjs");
```

and then in the docker-compose you can build with this set

Nsttt: Yep, this seems fine. Thanks!

juliusmarminge:
Need:

- changeset (so we can trigger a release)
- mentioning in the env-vars docs on a) what it does and b) how to use it

www/src/pages/en/deployment/docker.md
```diff
@@ -98,9 +90,9 @@ COPY . .
 # ENV NEXT_TELEMETRY_DISABLED 1
 
 RUN \
- if [ -f yarn.lock ]; then yarn build; \
- elif [ -f package-lock.json ]; then npm run build; \
- elif [ -f pnpm-lock.yaml ]; then yarn global add pnpm && pnpm run build; \
+ if [ -f yarn.lock ]; then SKIP_ENV_VALIDATION=1 yarn build; \
```
juliusmarminge, line 93: Maybe add a note before the details block about this part, since it's probably of the more custom thing we do, rest is pretty standard?

cli/template/base/next.config.mjs
```diff
@@ -1,5 +1,6 @@
 // @ts-check
-import { env } from "./src/env/server.mjs";
+/* run the build with this set to skip validation */
+!process.env.SKIP_ENV_VALIDATION && (await import("./src/env/server.mjs"));
```
c-ehrlich, line 3: This is such a clever solution :)

www/src/pages/en/deployment/docker.md
c-ehrlich: Agreed, a comment would be good here

Nsttt: Done !

www/src/pages/en/deployment/docker.md
```diff
@@ -56,7 +48,9 @@ README.md
 
 </details>
 
-### 4. Create Dockerfile
+### 3. Create Dockerfile
+
+> In order to not pull the server enviroment variables to our container, we have to add a`SKIP_ENV_VALIDATION=1`flag to the build command.
```
juliusmarminge, line 53:
```suggestion
> Since we're not pulling the server environment variables into our container, the [environment schema validation](/en/usage/environment-variables) will fail. To prevent this, we have to add a `SKIP_ENV_VALIDATION=1` flag to the build command so that the env-schemas aren't validated at build time.
```

juliusmarminge (approved):

www/src/pages/en/deployment/docker.md
```diff
@@ -56,7 +48,9 @@ README.md
 
 </details>
 
-### 4. Create Dockerfile
+### 3. Create Dockerfile
+
+> Since we're not pulling the server environment variables into our container, the [environment schema validation](/en/usage/environment-variables) will fail. To prevent this, we have to add a `SKIP_ENV_VALIDATION=1` flag to the build command so that the env-schemas aren't validated at build time.
```
juliusmarminge, line 53:
```suggestion
> Since we're not pulling the server environment variables into our container, the [environment schema validation](/en/usage/env-variables) will fail. To prevent this, we have to add a `SKIP_ENV_VALIDATION=1` flag to the build command so that the env-schemas aren't validated at build time.
```

PR #707, opened 2022-11-08
fix(docs): consistent use of `NextAuth` / `NextAuth.js`

AyanavaKarmakar:
## ✅ Checklist

---

## Changelog

- Replace all `Next-Auth` with `NextAuth`
- Replace all `Next-Auth.js` with `NextAuth.js`

Why omit the `-`? PFB.

PRimage

---

💯

### DIFF
--- a/www/src/pages/en/faq.md
+++ b/www/src/pages/en/faq.md
@@ -13,7 +13,7 @@ We try to keep this project as simple as possible, so you can start with just th
 If you are not familiar with the different technologies used in this project, please refer to the respective docs. If you still are in the wind, please join our [Discord](https://t3.gg/discord) and ask for help.
 
 - [Next.js](https://nextjs.org/)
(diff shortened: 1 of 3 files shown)


AyanavaKarmakar:
> Should we not be consistence with the `.js` too? Which is more correct? The full namn does include it?

Hmm, looks like they refer to the package as `NextAuth` but everywhere else they're using `NextAuth.js`... ig we can replace all `NextAuth` with `NextAuth.js` for consistency.

c-ehrlich:
Between their github, docs, etc they seem to be inconsistent between "NextAuth.js" and "NextAuth". IMO we should pick one and stick with it. I don't have a preference which one we choose.

juliusmarminge:
Came to think of this tweet from the creator:

https://twitter.com/balazsorban44/status/1489757336017195009

c-ehrlich (approved):

PR #710, opened 2022-11-08
feat: replace current multi-file template approach with handlebars

PiotrekPKP:
## ✅ Checklist

---

## Changelog

- removed all the `*.tsx` files in `/page-studs` directory
- replaced those files with 3 `*.hbs` (handlebars) files
- removed the "name picker" for template files and replaced it with a compiler for handlebars templates

---

## Screenshots

image

💯


c-ehrlich:
Thanks for the PR, it's an interesting approach to fixing the duplication problem. That being said the problem with the duplication isn't that it's more lines of code, but that it's not ideal in terms of DX. However it feels to me like this solution also comes with its own DX issues. 

Which of these is better/easier?
1. Work on one index file, which is just a normal `.tsx`, paste changes into other index files when done
2. With the handlebars setup, either:
    - Work directly in the handlebars file and not get live preview
    - Generate a `.tsx` file from the handlebars file, work in that, paste changes back into the handlebars file when done

My gut feeling is that 1 is actually the better DX despite of course being far from perfect. But I've not worked with handlebars in this sort of context before so am open to having my mind changed and would be interested to hear what others think.

c-ehrlich (commented):
Thanks for the PR, it's an interesting approach to fixing the duplication problem. That being said the problem with the duplication isn't that it's more lines of code, but that it's not ideal in terms of DX. However it feels to me like this solution also comes with its own DX issues.

Which of these is better/easier?

    Work on one index file, which is just a normal .tsx, paste changes into other index files when done
    With the handlebars setup, either:
        Work directly in the handlebars file and not get live preview
        Generate a .tsx file from the handlebars file, work in that, paste changes back into the handlebars file when done

My gut feeling is that 1 is actually the better DX despite of course being far from perfect. But I've not worked with handlebars in this sort of context before so am open to having my mind changed and would be interested to hear what others think.

nexxeln:
I also think handlebars will just slow us down, both in terms of perf and in terms of us maintainers when updating templates since we're not used to it.

It's easier for us to update stuff in the templates when they are a single file.

PiotrekPKP:
I thought about doing this switch after reading TODOs regading this issue.

Those are my thoughts:
- when it comes to code highlighting / auto completion while writing template files, it still does not work because of react package not being installed in the project (maybe I did something wrong tho?)
- copying content from one template to another CAN lead to mistakes such as forgotten changes and things not being prepared for the template
- updating templates is not really different from the current approach - you just have one / two files instead of multiple copies and you can just use the `{{#if withOption}}` helper

I don't think that there even is something that can be 100% perfect in our case, though handlebars can open us to new options like:
- inserting variables like project name inside of the template file (eg. changing page title from "Create T3 App" to `<my project name>`)
- adding more template options to the CLI, like "empty project" (which I would personally like to have in the future)
  - that would mean generating eg. one more set of template files, this time without the `AuthShowcase` and other components

Overall I'm not a fan of current approach. The lack of auto completion and language server should not limit us that much - maintaining multiple copies and being limited to only pre-written code is kind of cutting us off from new development opportunities.

Nevertheless, if you want me to change something, I'm open to your proposals :wink:

PR #717, opened 2022-11-10
docs: update first steps

robotkutya:
8. step is missing and needed

as per
https://www.nexxel.dev/blog/ct3a-guestbook#setting-up-authentication

Closes #<issue>

## ✅ Checklist

---

## Changelog

Following the 7. steps in the beta docs section "First Steps" does not work -- you can not log in via Discord Auth Provider.

I added an 8th step based on
https://www.nexxel.dev/blog/ct3a-guestbook#setting-up-authentication
to ensure a good first experience for new adapters.

---

## Screenshots

_[Screenshots]_

💯


www/src/pages/en/usage/first-steps.md
```diff
@@ -23,7 +23,8 @@ Of course, if you prefer to use a different auth provider, you can also use one
 5. Click "Reset Secret", copy the new secret, and add it to your `.env` as `DISCORD_CLIENT_SECRET`.
 6. Click "Add Redirect" and type in `http://localhost:3000/api/auth/callback/discord`.
    - For production deployment, follow the previous steps to create another Discord Application, but this time replace `http://localhost:3000` with the URL that you are deploying to.
-7. Save Changes. You should now be able to log in.
+7. Save Changes.
+8. Set the NEXTAUTH_SECRET as some random string too. You should now be able to log in.
```
AyanavaKarmakar, line 27: Maybe merge with Step 6, alongside `production deployment`, since it is [not required](https://next-auth.js.org/configuration/options#secret) in dev but will only throw an error in production.

c-ehrlich, line 27:
```suggestion
8. Set the `NEXTAUTH_SECRET` in `.env`. In development any string will work, for production see the note in `.env` on generating a secure secret.

You should now be able to log in.
```

c-ehrlich (changes requested):
Thanks for opening this :)

www/src/pages/en/usage/first-steps.md
c-ehrlich: also is the secret even required in dev? i thought it just complains in the console if you don't have it?

AyanavaKarmakar: > If no value is specified in development (and there is no NEXTAUTH_SECRET variable either), it uses a hash for all configuration options, including OAuth Client ID / Secrets for entropy.

https://next-auth.js.org/warnings#no_secret

c-ehrlich: hmm probably better to suggest adding it

robotkutya: Can you quickly try it out by creating a new project? I just did and I can not log in without setting it to something. I am doing this on `dev`. So it doesn't just throw an error for me.

Step 6) is in the UI of discord. You can't set the `NEXTAUTH_SECRET` there. So I recommend keeping it separate.

robotkutya: @c-ehrlich Much better wording, thanks! :)

juliusmarminge: I think it's cause we have the secret set to empty string by default?

If it's undefined, Next-Auth will override it and set it to some hash based on your config.

Maybe the solution here is to either predefine it to something like `supersecret` or comment it out, alternatively add a `.min(1)` to the Zod schema?

AyanavaKarmakar: I see. How about adding a comment in `.env`, like 'Make sure not to keep the secret empty', alongside `.min(1)` in the Zod schema?

Looks like they will probably make defining the secret mandatory in the future.

> We will likely make this option mandatory, even in development, in the future.

https://next-auth.js.org/configuration/options#description-1

c-ehrlich: > I think it's cause we have the secret set to empty string by default?
> 
> If it's undefined, Next-Auth will override it and set it to some hash based on your config.
> 
> Maybe the solution here is to either predefine it to something like `supersecret` or comment it out, alternatively add a `.min(1)` to the Zod schema?

definitely agree with `min(1)` in schema

i don't love adding something like `NEXTAUTH_SECRET=changeme` to `.env` but it's probably better than having `pnpm dev` fail out of the box

robotkutya: It seems to me like we reached a consensus! :)

Added the `min(1)` check to the schema --> users will now have a meaningful error message guide them when no auth env var is set 👍 

Also added the suggested copy change to the readme. I think it reads very well with that change, no need to add a dummy string to the config. 🖖 

Anything else, guys?

robotkutya: Ah okay, adding the `min(1)` breaks the build step, I get what you mean now.

Added a dummy string `changeme`.

robotkutya: just noticed I changed it to `'changeme'` ... 🤦 

centering a div I mean changing a string is hard, okay?!
plus it's microsoft's fault, because code spaces

fixed it 🤖

juliusmarminge (approved):

c-ehrlich (approved):
looks good, lets merge :)

PR #719, opened 2022-11-11
fix: not able to log in if `NEXTAUTH_SECRET` is not set in dev

AyanavaKarmakar:
## ✅ Checklist

---

## Changelog

- Add `.min(1)` to Zod schema of `NEXTAUTH_SECRET`
- Add a comment to prevent keeping the secret empty in `.env`
- Warn the user to add the secret in cli next steps

---

## Alternate Solution

- Add `.min(1)` to Zod schema of `NEXTAUTH_SECRET`
- Predefine `NEXTAUTH_SECRET=changeme` https://github.com/t3-oss/create-t3-app/pull/717#discussion_r1020081951
- Warn the user to update the predefined secret in cli next steps

💯

### DIFF
--- a/cli/src/helpers/logNextSteps.ts
+++ b/cli/src/helpers/logNextSteps.ts
@@ -28,9 +28,5 @@ export const logNextSteps = ({
     );
   }
 
-  if (packages?.nextAuth.inUse) {
-    logger.warn("  change NEXTAUTH_SECRET in .env");
-  }
-
   logger.info(`  ${pkgManager === "npm" ? "npm run" : pkgManager} dev`);
 };
--- a/cli/src/helpers/logNextSteps.ts
+++ b/cli/src/helpers/logNextSteps.ts
@@ -29,7 +29,7 @@ export const logNextSteps = ({
   }
 
   if (packages?.nextAuth.inUse) {
-    logger.warn("  set your own NEXTAUTH_SECRET in .env");
+    logger.warn("  change NEXTAUTH_SECRET in .env");
   }
 
   logger.info(`  ${pkgManager === "npm" ? "npm run" : pkgManager} dev`);
--- a/cli/src/helpers/logNextSteps.ts
+++ b/cli/src/helpers/logNextSteps.ts
@@ -28,5 +28,9 @@ export const logNextSteps = ({
     );
   }
 
+  if (packages?.nextAuth.inUse) {
+    logger.warn("set NEXTAUTH_SECRET in .env");
+  }
+
   logger.info(`  ${pkgManager === "npm" ? "npm run" : pkgManager} dev`);
 };
--- a/cli/src/helpers/logNextSteps.ts
+++ b/cli/src/helpers/logNextSteps.ts
@@ -28,5 +28,9 @@ export const logNextSteps = ({
     );
   }
 
+  if (packages?.nextAuth.inUse) {
+    logger.warn("  change NEXTAUTH_SECRET in .env");
+  }
+
   logger.info(`  ${pkgManager === "npm" ? "npm run" : pkgManager} dev`);
 };
--- a/cli/src/helpers/logNextSteps.ts
+++ b/cli/src/helpers/logNextSteps.ts
@@ -29,7 +29,7 @@ export const logNextSteps = ({
   }
 
   if (packages?.nextAuth.inUse) {
-    logger.warn("  set NEXTAUTH_SECRET in .env");
+    logger.warn("  set your own NEXTAUTH_SECRET in .env");
   }
 
   logger.info(`  ${pkgManager === "npm" ? "npm run" : pkgManager} dev`);
--- a/cli/src/helpers/logNextSteps.ts
+++ b/cli/src/helpers/logNextSteps.ts
@@ -29,7 +29,7 @@ export const logNextSteps = ({
   }
 
   if (packages?.nextAuth.inUse) {
-    logger.warn("set NEXTAUTH_SECRET in .env");
+    logger.warn("  set NEXTAUTH_SECRET in .env");
   }
 
   logger.info(`  ${pkgManager === "npm" ? "npm run" : pkgManager} dev`);
--- a/cli/src/installers/envVars.ts
+++ b/cli/src/installers/envVars.ts
@@ -38,7 +38,7 @@ DATABASE_URL=file:./db.sqlite
 # Next Auth
 # You can generate the secret via 'openssl rand -base64 32' on Linux
 # More info: https://next-auth.js.org/configuration/options#secret
-# NEXTAUTH_SECRET=
+NEXTAUTH_SECRET=
 NEXTAUTH_URL=http://localhost:3000
 
 # Next Auth Discord Provider
--- a/cli/src/installers/envVars.ts
+++ b/cli/src/installers/envVars.ts
@@ -37,6 +37,7 @@ DATABASE_URL=file:./db.sqlite
     envContent += `
 # Next Auth
 # You can generate the secret via 'openssl rand -base64 32' on Linux
+# Make sure not to keep the string empty
 # More info: https://next-auth.js.org/configuration/options#secret
 NEXTAUTH_SECRET=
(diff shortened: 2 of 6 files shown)


AyanavaKarmakar:
Note: Primary solution fails/cancels E2E tests because of wrong env var error due to `.min(1)` and `NEXTAUTH_SECRET` being empty

cli/src/helpers/logNextSteps.ts
```diff
@@ -28,5 +28,9 @@ export const logNextSteps = ({
     );
   }
 
+  if (packages?.nextAuth.inUse) {
+    logger.warn("set NEXTAUTH_SECRET in .env");
```
juliusmarminge, line 32: Needs indentation

juliusmarminge (changes requested):

cli/src/helpers/logNextSteps.ts
AyanavaKarmakar: Fixed via https://github.com/t3-oss/create-t3-app/pull/719/commits/ab9d0fa3f1531f3b3197e1d5ea175b60108a187e

nexxeln (approved):

cli/src/installers/envVars.ts
```diff
@@ -37,8 +37,9 @@ DATABASE_URL=file:./db.sqlite
     envContent += `
 # Next Auth
 # You can generate the secret via 'openssl rand -base64 32' on Linux
+# Make sure not to keep the secret empty
```
c-ehrlich, line 40: don't think we need this if we are including a default

cli/src/helpers/logNextSteps.ts
```diff
@@ -28,5 +28,9 @@ export const logNextSteps = ({
     );
   }
 
+  if (packages?.nextAuth.inUse) {
+    logger.warn("  set NEXTAUTH_SECRET in .env");
+  }
```
c-ehrlich, line 33: I don't know if we need to add the line about the secret to CLI - we are including a default secret now, and our .env (which users need to check when deploying to Vercel etc because they need to know which env vars to specify there) already includes instructions for how to generate a production ready secure secret.

But I could be convinced otherwise if you or someone else feels strongly about this

cli/src/installers/envVars.ts
```diff
@@ -37,8 +37,9 @@ DATABASE_URL=file:./db.sqlite
...
 # More info: https://next-auth.js.org/configuration/options#secret
-NEXTAUTH_SECRET=
+NEXTAUTH_SECRET=CHANGE_ME
```
c-ehrlich, line 42: I like this much more than the secret in the other NEXTAUT_SECRET related pr that we merged, because it makes it clear that this shouldn't be used as the "real" value. Please keep your version when resolving the merge conflict.

cli/src/helpers/logNextSteps.ts
juliusmarminge: I think it can stay

cli/src/installers/envVars.ts
AyanavaKarmakar: removed (https://github.com/t3-oss/create-t3-app/pull/719/commits/0d284110f835d268c45d43fd03e8aa6c291a07b8)

cli/src/helpers/logNextSteps.ts
AyanavaKarmakar: Since we are already predefining the secret, maybe it can be reworded to "  set your own NEXTAUTH_SECRET in .env"?

juliusmarminge: Or just "change"

juliusmarminge: An alternative approach is to change the schema so it's only required in prod, and have the NEXTAUTH_SECRET commented out in the .env

juliusmarminge:
I think you gotta add a secret to the ci

AyanavaKarmakar:
> I think you gotta add a secret to the ci

I'm not really sure how to do that. Something like this?

image

juliusmarminge:
> > I think you gotta add a secret to the ci
> 
> 
> 
> I'm not really sure how to do that. Something like this?
> 
> 
> 
> image
> 
> 

Yea just some dummy string but in the e2e workflow

AyanavaKarmakar:
Finally! Thanks for the pointer, Julius.

cli/src/installers/envVars.ts
```diff
@@ -38,7 +38,7 @@ DATABASE_URL=file:./db.sqlite
 # Next Auth
 # You can generate the secret via 'openssl rand -base64 32' on Linux
 # More info: https://next-auth.js.org/configuration/options#secret
-NEXTAUTH_SECRET=supersecret
+# NEXTAUTH_SECRET=
```
c-ehrlich, line 41: `z.string()` actually passes with an empty string, so im not sure its necessary to comment it out here?

cli/template/addons/env/auth-prisma-schema.mjs
```diff
@@ -8,7 +8,10 @@ import { z } from "zod";
 export const serverSchema = z.object({
   DATABASE_URL: z.string().url(),
   NODE_ENV: z.enum(["development", "test", "production"]),
-  NEXTAUTH_SECRET: z.string().min(1),
+  NEXTAUTH_SECRET:
+    process.env.NODE_ENV === "production"
+      ? z.string().min(1)
+      : z.string().optional(),
```
c-ehrlich, line 14: does `z.string().optional()` behave differently than just `z.string()` in this context?

juliusmarminge: Should be .min(1).optional() I think.

We want it to be okay to omit the secret completely (which it is when it's commented out) but if it's defined, it can't be empty.

juliusmarminge: > does `z.string().optional()` behave differently than just `z.string()` in this context?

Yes, optional means it can be commented out (I.e undefined)

cli/template/addons/env/auth-prisma-schema.mjs
```diff
@@ -8,7 +8,10 @@ import { z } from "zod";
 export const serverSchema = z.object({
   DATABASE_URL: z.string().url(),
   NODE_ENV: z.enum(["development", "test", "production"]),
-  NEXTAUTH_SECRET: z.string().min(1),
+  NEXTAUTH_SECRET:
+    process.env.NODE_ENV === "production"
+      ? z.string().min(1).optional()
+      : z.string().optional(),
```
juliusmarminge, line 14:
```suggestion
      ? z.string().min(1)
      : z.string().min(1).optional(),
```

cli/template/addons/env/auth-schema.mjs
```diff
@@ -7,7 +7,10 @@ import { z } from "zod";
  */
 export const serverSchema = z.object({
   NODE_ENV: z.enum(["development", "test", "production"]),
-  NEXTAUTH_SECRET: z.string().min(1),
+  NEXTAUTH_SECRET:
+    process.env.NODE_ENV === "production"
+      ? z.string().min(1).optional()
+      : z.string().optional(),
```
juliusmarminge, line 13:
```suggestion
      ? z.string().min(1)
       : z.string().min(1).optional(),
```

cli/template/addons/env/auth-schema.mjs
```diff
@@ -7,7 +7,10 @@ import { z } from "zod";
  */
 export const serverSchema = z.object({
   NODE_ENV: z.enum(["development", "test", "production"]),
-  NEXTAUTH_SECRET: z.string().min(1),
+  NEXTAUTH_SECRET:
+    process.env.NODE_ENV === "production"
+      ? z.string().min(1)
+       : z.string().min(1).optional(),
```
juliusmarminge, line 13:
```suggestion
      ? z.string().min(1)
      : z.string().min(1).optional(),
```

juliusmarminge (approved):

PR #727, opened 2022-11-13
feat: render warning when using a tagged or outdated version

juliusmarminge:
Closes #<issue>

## ✅ Checklist

---

## Changelog

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/cli/src/utils/renderVersionWarning.ts
+++ b/cli/src/utils/renderVersionWarning.ts
@@ -1,12 +1,15 @@
-import { execSync } from "child_process";
+// import { execSync } from "child_process";
+import { execa } from "execa";
 import { getVersion } from "./getT3Version.js";
 import { logger } from "./logger.js";
 
-export const renderVersionWarning = () => {
+export const getNpmVersion = async () =>
+  await execa("npm", ["view", "create-t3-app", "version"]).then(
+    (v) => v.stdout,
+  );
+
+export const renderVersionWarning = (npmVersion: string) => {
   const currentVersion = getVersion();
-  const npmVersion = execSync("npm view create-t3-app version")
-    .toString()
-    .trim();
 
   if (currentVersion.includes("beta")) {
     logger.warn("  You are using a beta version of create-t3-app.");
--- a/cli/src/utils/renderVersionWarning.ts
+++ b/cli/src/utils/renderVersionWarning.ts
@@ -8,10 +8,10 @@ export const renderVersionWarning = () => {
     .toString()
     .trim();
 
-  if (currentVersion.includes("@beta")) {
+  if (currentVersion.includes("beta")) {
     logger.warn("You are using a beta version of create-t3-app.");
     logger.warn("Please report any bugs you encounter.");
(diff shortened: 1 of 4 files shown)


juliusmarminge:
CleanShot 2022-11-13 at 12 10 15@2x

juliusmarminge:
I don't like the delay but not sure if there's a way to get rid of it?
CleanShot 2022-11-13 at 12 16 06

juliusmarminge:
I think it is

cli/src/utils/renderVersionWarning.ts
```diff
@@ -0,0 +1,30 @@
+import { execSync } from "child_process";
+import { getVersion } from "./getT3Version.js";
+import { logger } from "./logger.js";
+
+export const getNpmVersion = () =>
```
c-ehrlich, line 5: you can get rid of the delay in most cases by checking `https://registry.npmjs.org/-/package/create-t3-app/dist-tags` first, which is just an http request that returns json, and falling back to the terminal command when it fails

see here how create-react-app does it: https://github.dev/facebook/create-react-app/blob/d960b9e38c062584ff6cfb1a70e1512509a966e7/packages/create-react-app/createReactApp.js#L198

juliusmarminge: Yea that seems faster.

I still think it's better to put the request before we render the title so that they appear at the same time, thus reducing "flashing/layout shift".

cli/src/utils/renderVersionWarning.ts
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ * https://github.com/facebook/create-react-app/blob/main/packages/create-react-app/LICENSE
+ */
+import https from "https";
+function checkForLatestVersion(): Promise<string> {
+  return new Promise((resolve, reject) => {
```
c-ehrlich, line 41: could probably write a simpler version of this using await and fetch, but the facebook code is most likely very stable and well tested, so i'm ok with keeping this syntax :)

c-ehrlich:
Tested the prerelease version and it works for beta. Couldn't figure out how to test it for "real" versions but feel free to resolve and merge if you're sure it works for those :)

Also currently it will bring up the message if the user somehow managed to have a _newer_ version than what's on npm, but I think this is actually a good thing.

cli/src/utils/renderVersionWarning.ts
juliusmarminge: yea I just copied their ;)

juliusmarminge:
> Also currently it will bring up the message if the user somehow managed to have a _newer_ version than what's on npm, but I think this is actually a good thing.

The only way that will happen is if they modify the version in the package.json so I think that's an edge case I'm willing pass on.

nexxeln (approved):
Ship i guess

PR #736, opened 2022-11-15
docs(tRPC): Switch to new `setData` signature

bennettdams:
## ✅ Checklist

---

## Changelog

The signature for `setData` has changed. This commit will do two things:
- Provide `undefined` for the input argument (mandatory).
- Use the updater function (`old => [...old, newPost ]`) instead of the `prevData` directly. This makes it also align with the React Query docs:
  
  image

### DIFF
--- a/www/src/pages/en/usage/trpc.md
+++ b/www/src/pages/en/usage/trpc.md
@@ -258,10 +258,13 @@ const MyComponent = () => {
     async onMutate(newPost) {
       // Cancel outgoing fetches (so they don't overwrite our optimistic update)
       await utils.post.list.cancel();
+
       // Get the data from the queryCache
       const prevData = utils.post.list.getData();
+
       // Optimistically update the data with our new post
-      utils.post.list.setData([...prevData, newPost ]);
+      utils.post.list.setData(undefined, old => [...old, newPost ]);
+
       // Return the previous data so we can revert if something goes wrong
       return { prevData };
     },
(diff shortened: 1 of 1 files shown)


www/src/pages/en/usage/trpc.md
```diff
@@ -258,10 +258,13 @@ const MyComponent = () => {
     async onMutate(newPost) {
       // Cancel outgoing fetches (so they don't overwrite our optimistic update)
       await utils.post.list.cancel();
+
       // Get the data from the queryCache
       const prevData = utils.post.list.getData();
```
c-ehrlich, line 263: Not sure about the API changes, Julius can probably give a comment on that. But if we're not using `prevData` anymore then we don't need it here.

bennettdams: I let `prevData` like it is, as one can use it to put it into the context by returning it from `onMutate`. That way, you can use the previous data to roll back (see `onError` below) on a failed mutation.

bennettdams: Oh, I just realized there's also another old `setData` call in `onError`. Will add another commit.

www/src/pages/en/usage/trpc.md
```diff
@@ -258,16 +258,19 @@ const MyComponent = () => {
...
+
       // Optimistically update the data with our new post
-      utils.post.list.setData([...prevData, newPost ]);
+      utils.post.list.setData(undefined, old => [...old, newPost ]);
```
juliusmarminge, line 266:
```suggestion
      utils.post.list.setData(undefined, (old) => [...old, newPost]);
```

PR #786, opened 2022-11-17
chore: ignore next-env.d.ts

skuridin:
## ✅ Checklist

---

## Changelog

Added `next-env.d.ts` to `.gitignore` as recommended in Next.js [documentation](https://nextjs.org/docs/basic-features/typescript).

---

## Screenshots

image

💯

### DIFF
--- a/.changeset/five-tigers-double.md
+++ b/.changeset/five-tigers-double.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+Add next-env.d.ts to .gitignore
--- a/cli/template/base/_gitignore
+++ b/cli/template/base/_gitignore
@@ -15,6 +15,7 @@
 # next.js
 /.next/
 /out/
+next-env.d.ts
 
 # production
 /build


.changeset/five-tigers-double.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+Add next-env.d.ts to .gitignore
```
juliusmarminge, line 5: Should probably also delete the file https://github.com/t3-oss/create-t3-app/blob/next/cli/template/base/next-env.d.ts

nexxeln (commented):
Pretty sure the docs are incorrect here. This file is required for TypeScript to work in Next.js and it should not be removed or edited. So if someone clones a project they would have to make this file on their own.

In the latest create-next-app as well this is added and committed in the initial commit.

juliusmarminge:
> Pretty sure the docs are incorrect here. This file is required for TypeScript to work in Next.js and it should not be removed or edited. So if someone clones a project they would have to make this file on their own.
> 
> 
> 
> In the latest create-next-app as well this is added and committed in the initial commit.
> 
> 
> 
> Please make this PR to the Next.js repo and fix their docs haha.

It's generated everytime you run dev or build

nexxeln:
> > Pretty sure the docs are incorrect here. This file is required for TypeScript to work in Next.js and it should not be removed or edited. So if someone clones a project they would have to make this file on their own.
> > In the latest create-next-app as well this is added and committed in the initial commit.
> > Please make this PR to the Next.js repo and fix their docs haha.
> 
> It's generated everytime you run dev or build

Oh I didn't know that. But then why does create-next-app put it in initial commit?

.changeset/five-tigers-double.md
skuridin: I would keep it to be consistent with `create-next-app`. It does copy this file into created app directory.

skuridin:
> Oh I didn't know that. But then why does create-next-app put it in initial commit?

I just created new `create-next-app` and while the file is present, it is not in the "Initial commit from Create Next App" git commit.

I can only assume it's copied to the new project to avoid surprises from sudden file appearance.

nexxeln:
> > Oh I didn't know that. But then why does create-next-app put it in initial commit?
> 
> I just created new `create-next-app` and while the file is present, it is not in the "Initial commit from Create Next App" git commit.
> 
> I can only assume it's copied to the new project to avoid surprises from sudden file appearance.

Ah okay that makes sense.

nexxeln:
> Then we can remove this part in the docs maybe?
> 
> https://create.t3.gg/en/folder-structure#next-envdts
> 
> cc @nexxeln

It's still a file that is being generated, so I think we should keep it

.changeset/five-tigers-double.md
nexxeln: Yeah I agree we shouldn't delete

nexxeln (approved):
LGTM

PR #794, opened 2022-11-18
fix: Mobile pop out menu hidden incorrectly

### LINKED ISSUE #777: bug: Docs- Mobile pop out menu hidden incorrectly

https://create.t3.gg/
Chrome

### Describe the bug

image

When looking at the Docs- if the browser is 767px wide, the mobile nav menu is hidden incorrectly, meaning the button in the top right to expand/hide the menu, appears to do nothing.

### To reproduce

Go to- https://create.t3.gg/
Set browser width to 767px
Attempt to click the menu button top right

### Additional information

_No response_

KarthikRaju391:
Closes #777

## ✅ Checklist

---

## Changelog

Changed the `max-width` property of the media-query from `767px` to `768px`

---

## Screenshots

image
image

💯


nexxeln:
Please remove the changeset, we only need it for CLI changes

KarthikRaju391:
So, I have to remove the changeset and push it again?

juliusmarminge (changes requested):
This is incorrect I think?

With this the menu stays open at 768 but the toggle is removed

This one:

https://user-images.githubusercontent.com/51714798/202823173-6ab3965b-7ca8-4add-b29c-bfc9f5a8e64e.mp4

Production: 

https://user-images.githubusercontent.com/51714798/202823081-61c3305a-77ce-47a8-b003-20688b04303d.mp4

KarthikRaju391:
I'll look into it @juliusmarminge

c-ehrlich:
The reason for the 767 value is that the breakpoint is at 768, so we want this behaviour until 1 pixel before that.

juliusmarminge:
> The reason for the 767 value is that the breakpoint is at 768, so we want this behaviour until 1 pixel before that.
> 
> 
> 
> Was this PR in response to a specific issue in the UI?

#777 but that issue seems very flaky

KarthikRaju391:
The issue was that at 767 px the popout wasn't being shown.
It was being shown at 766px and 768px but not at 767px

AyanavaKarmakar:
> The reason for the 767 value is that the breakpoint is at 768, so we want this behaviour until 1 pixel before that.

Looks like the menu doesn't pop up **only** @ 767 x 767 dimensions. What if we set the `max-width` to 767.5 instead of 767 / 768?

KarthikRaju391:
It is actually working correctly when `max-width` is 767.5px.

KarthikRaju391:
https://user-images.githubusercontent.com/67912316/202848877-37a97b7b-add1-4195-ba38-43837167d2d8.mp4

c-ehrlich:
> Create.T3.App.-.Brave.2022-11-19.17-08-11.mp4

what you are showing here seems like the correct behaviour - the menu exists until 767 px, and doesn't exist at 768+ px because then the "Docs" etc items appear in the navbar instead

AyanavaKarmakar:
> It is actually working correctly when `max-width` is 767.5px.

Can you please push the changes, so that we can get a live preview?

juliusmarminge:
I still can't reproduce the original issue though, it works as expected with breakpoint on 768 for me on both Safari, Chrome and Firefox.

nexxeln:
> I still can't reproduce the original issue though, it works as expected with breakpoint going from 767 to 768 for me on both Safari, Chrome and Firefox.

I can reproduce the issue on brave on windows

KarthikRaju391:
I'm out currently, I'll push the changes in an hour.

nexxeln (approved):
Works on my end

juliusmarminge:
Nothing broke on my end either, still good on Safari, Chrome and Firefox for me

juliusmarminge (approved):

PR #813, opened 2022-11-19
chore(ci): use new output api

juliusmarminge:
- uses new output api
- updates to use node 18 in ci

These warnings are now gone: 

CleanShot 2022-11-20 at 01 02 06@2x


c-ehrlich (approved):
Nice, I mentioned this a while ago and then completely forgot about it 😅

Also somehow missed that you had already upgraded to semantic-pull-request@v5 and actions/checkout@v3. Should be good now for a while :)

PR #830, opened 2022-11-22
fix(style): imports grouping, trivial

dreglad:
This is a trivial style fix. Third party library imports are grouped apart from local imports throughout the project, I guess this file shouldn't be an exception.

#### **I assume this change is too trivial to require anything from the checklist other than following the commit convention**

- [ ] **_Not required_** I performed a functional test on my final commit.

---

## Changelog

_[Not required]_

---

## Screenshots

_[Not required]_

💯

### DIFF
--- a/cli/template/addons/next-auth/get-server-auth-session.ts
+++ b/cli/template/addons/next-auth/get-server-auth-session.ts
@@ -1,5 +1,6 @@
 import { type GetServerSidePropsContext } from "next";
 import { unstable_getServerSession } from "next-auth";
+
 import { authOptions } from "../../pages/api/auth/[...nextauth]";
 
 /**


dreglad:
#### An additional thought on the topic

What about having imports sorted/grouped by the linter?

Something like [prettier-plugin-sort-imports](https://www.npmjs.com/package/@trivago/prettier-plugin-sort-imports) would do the job with minimal boilerplate.

I wouldn't disagree if some consider that grouping imports by default may be too opinionated, but I don't think anyone could reasonably argue against enforcing import order by default..  Much like `prettier-plugin-tailwindcss` already does for Tailwind classes

wyattades:
Hey all, I know there's a lot of solutions/packages out there for "sortings imports", but I think one in particular is the best fit for the prettier + typescript philosophies: https://github.com/simonhaenisch/prettier-plugin-organize-imports 

From there docs:
> A plugin that makes Prettier organize your imports (i. e. sorts, combines and removes unused ones) using the `organizeImports` feature of the TypeScript language service API. This is the same as using the "Organize Imports" action in VS Code.

In other words, developers should not have to care about how imports are grouped. The plugin itself doesn't even know. Just let the typescript compiler decide.

---

Also, `prettier-plugin-tailwindcss@0.2.0` [supposedly fixes](https://github.com/tailwindlabs/prettier-plugin-tailwindcss/issues/31#issuecomment-1327882587) compatibility with the other prettier plugins.

PR #831, opened 2022-11-22
fix: enable NEXTAUTH_SECRET by default

ruhap:
NEXTAUTH_SECRET is not enabled by default but Discord provider keys are. Next-auth needs to have NEXTAUTH_SECRET and its also checked on schema.

Closes #<issue>

## ✅ Checklist

---

## Changelog

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/cli/src/installers/envVars.ts
+++ b/cli/src/installers/envVars.ts
@@ -38,7 +38,7 @@ DATABASE_URL=file:./db.sqlite
 # Next Auth
 # You can generate the secret via 'openssl rand -base64 32' on Linux
 # More info: https://next-auth.js.org/configuration/options#secret
-# NEXTAUTH_SECRET=
+NEXTAUTH_SECRET=
 NEXTAUTH_URL=http://localhost:3000
 
 # Next Auth Discord Provider


ruhap:
This commit would make it more clear that this .env variable needs to be used. Error you get without it:

```
`[next-auth][error][SIGNIN_OAUTH_ERROR] 
https://next-auth.js.org/errors#signin_oauth_error "ikm" must be at least one byte in length {
  error: {
    message: '"ikm" must be at least one byte in length',
    stack: 'TypeError: "ikm" must be at least one byte in length\n' +`
```

c-ehrlich:
This was recently changed in https://github.com/t3-oss/create-t3-app/pull/719

If I understand correctly, the reason was specifically to make it fail - because including an empty key or the same default key in every t3 app are both bad for security.

I would be open to having my mind changed on this, but if we do something it should imo be a default key like `changeme`. 

What does everyone else think?

AyanavaKarmakar:
> I would be open to having my mind changed on this, but if we do something it should imo be a default key like changeme.

IMO, if we set a default value no one will probably bother to change it, since any string works and we will end up with numerous apps using the same secret. Which isn't good since it is used to [encrypt](https://next-auth.js.org/configuration/options#nextauth_secret) the NextAuth.js JWT, and hash email verification tokens.

juliusmarminge:
I thought if you didn't set one at all (which is the case when it's commented out) that NextAuth would generate one for you by hashing your options?

c-ehrlich:
> I thought if you didn't set one at all (which is the case when it's commented out) that NextAuth would generate one for you by hashing your options?

Yes I think that's right. It does that in dev, and fails in prod, which is good behavior imo (we don't want people to use `changeme` or whatever as their prod secret).

nexxeln:
Yeah I like the current implementation as well

PR #836, opened 2022-11-22
fix: twitter data

juliusmarminge:
created a new token and added it to vercel

setting cache control to public to further reduce requests, maybe that helps?

### DIFF
--- a/www/src/components/landingPage/tweets/tweets.astro
+++ b/www/src/components/landingPage/tweets/tweets.astro
@@ -60,11 +60,8 @@ if (import.meta.env.MODE === "CI" || !token) {
     "1570054715240763393",
   ]);
 }
-const SIX_HOURS_IN_SECONDS = 6 * 60 * 60;
-Astro.response.headers.set(
-  "Cache-Control",
-  `s-max-age=${SIX_HOURS_IN_SECONDS}`,
-);
+const THREE_HOURS = 3 * 60 * 60;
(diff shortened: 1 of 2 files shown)


www/src/components/landingPage/tweets/tweets.astro
```diff
@@ -60,11 +60,8 @@ if (import.meta.env.MODE === "CI" || !token) {
     "1570054715240763393",
   ]);
 }
-const SIX_HOURS_IN_SECONDS = 6 * 60 * 60;
-Astro.response.headers.set(
-  "Cache-Control",
-  `s-max-age=${SIX_HOURS_IN_SECONDS}`,
-);
+const THREE_HOURS = 3 * 60 * 60;
```
nexxeln, line 63: Why did we make it 3 hours from 6 hours?

juliusmarminge: Idk just felt like it ;)

Public cache so will still reduce requests quite a bit

c-ehrlich: 3 hours seems fine to me, should not get rate limited on that

juliusmarminge: I have elevated access on my twitter app which gives me 2M requests per month, should be good

c-ehrlich (approved):

PR #840, opened 2022-11-23
fix: extra whitespace on landing page

iDarkLightning:
## ✅ Checklist

---

## Changelog

Display the background pattern via a pseudo-element bg-image instead of an image tag. Fixes unnecessary whitespace on specific screen sizes
---

## Screenshots

**Current**:
image

**Changed**:
image

💯

### DIFF
--- a/www/src/layouts/landingPage.astro
+++ b/www/src/layouts/landingPage.astro
@@ -43,14 +43,9 @@ const { isNotFound } = Astro.props;
     class="min-h-screen bg-gradient-to-b from-gray-900 via-[#300171] to-slate-900"
   >
     <JumpToContent />
-    <img
-      src="/images/background-pattern.svg"
-      class={`absolute inset-0 opacity-5 w-max min-w-full object-cover ${
-        !isNotFound ? "h-max" : "h-screen"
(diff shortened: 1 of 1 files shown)


c-ehrlich:
I'm not seeing the background at all on the test deployment. But your screenshot of the fix has it. Could you see what's the difference between between your local environment and the deployment?

image
vs
image

iDarkLightning:
Weird, it seems to be trying to find the image on the path `/assets/images/background-image.svg` on the preview deployment. That's giving a 404 which results in the image not being shown. Locally, it works fine and just targets `images/background-image.svg`.

Not sure why this is happening, do you know if this is something to do with the preview deployment system here? I've never faced this issue on vercel before

c-ehrlich:
> ...would love an explanation on why that makes a difference but it seems to be working now

I believe this might be because images are served differently depending on environment  - in staging we build the static site, on your computer it's coming from the development server.

Could check this theory by reverting to the "broken" version, and building / starting the static site locally.

c-ehrlich (approved):
Looks good, thanks for the contribution :)

PR #846, opened 2022-11-23
docs: ternary theme toggle

ruridge:
## ✅ Checklist

---

Issue: #847 

## Changelog

Overriding the devices preferred color scheme and not being able to revert to the default setting is an anti-pattern and something that needs to be driven out of web design.

Options are "light", "dark", and "system".  The functionality is exactly the same as before but with the addition of allowing the user to revert their color scheme preference back to the preferred color scheme set on the device.

Some of the changes have been implemented as an inline script, this is to ensure there is no flash of un-styled content when the page first loads (selected color scheme button color).

---

### Chrome

chrome-screenshot

### Safari

safari-screenshot

### iPhoneSE

iPhoneSE-screenshot

### Video
https://user-images.githubusercontent.com/856071/203596816-fdc5158c-db7f-423f-ac00-0236e293dc66.mov

💯

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -58,6 +58,6 @@
     "prettier-plugin-tailwindcss": "^0.1.13",
     "tsup": "^6.2.3",
     "type-fest": "^3.0.0",
-    "typescript": "^4.8.4"
+    "typescript": "^4.9.3"
   }
 }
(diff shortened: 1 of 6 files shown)


c-ehrlich:
We switched to the entire theme select being a single button recently because the touch targets are frustratingly small otherwise. I could see moving to a theme dropdown that includes "System" as an option, but not a fan of this three-icon toggle.

juliusmarminge:
Yea a drop-down is probably the way here. Not a fan of a three-way-button, very hard to click especially on mobile.

PR #849, opened 2022-11-23
docs: lang select headless ui

juliusmarminge:
use headless ui for better accessibility


c-ehrlich (approved):
aria and tabselect seems to work :) lgtm

PR #863, opened 2022-11-25
docs(RFC): external link glyph

juliusmarminge:
was brought up in another pr that we should mark external links.

not sure if I like it or not myself

CleanShot 2022-11-25 at 09 17 37@2x

### DIFF
--- a/www/astro.config.mjs
+++ b/www/astro.config.mjs
@@ -18,6 +18,10 @@ export default defineConfig({
         {
           target: "_blank",
           rel: ["noreferrer noopener"],
+          content: {
+            type: "text",
+            value: "↗",
+          },
         },
       ],
       "rehype-slug",
--- a/www/src/styles/global.css
+++ b/www/src/styles/global.css
@@ -162,10 +162,6 @@
           dark:decoration-t3-purple-200 dark:hover:text-t3-purple-300;
   }
 
-  .markdown a[rel~="noreferrer"] > span {
-    @apply pl-1;
-  }
-
   .markdown code {
     @apply break-words lg:break-normal;
   }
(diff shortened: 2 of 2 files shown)


nexxeln (approved):
I like it

c-ehrlich:
I think it's alright, but:
1. i don't think this is the right icon/styling for it yet, these usually look something like this: (random google image search screenshot)
image
2. I think the reason this was brought up in first place is a different problem, which is that our color palette is missing contrast colors so it becomes kind of hard to parse the page. (but that doesn't mean we shouldn't add the glyph)

Compare Astro docs to us:


image

nexxeln:
> I think it's alright, but:
> 
> 1. i don't think this is the right icon/styling for it yet, these usually look something like this:

Does it matter though? I really like how the arrow looks, very minimal

juliusmarminge:
> i don't think this is the right icon/styling for it yet, these usually look something like this: (random google image search screenshot)

I don't think that icon is a valid unicode, I couldn't find any atleast. Not sure if you can put a proper svg

asrvd:
github has something like this too, looks nice to me, maybe add a bit of gap between text and icon though
image

juliusmarminge:
Added some padding but not the biggest fan that the underline "cuts off"

CleanShot 2022-11-27 at 11 39 53@2x

nexxeln:
> Added some padding but not the biggest fan that the underline "cuts off"
> 
> CleanShot 2022-11-27 at 11 39 53@2x

it was fine without the padding imo

juliusmarminge:
> > Added some padding but not the biggest fan that the underline "cuts off"
> > CleanShot 2022-11-27 at 11 39 53@2x
> 
> it was fine without the padding imo

Cool, I'll remove the padding then :)

PR #879, opened 2022-11-26
feat(www): add breadcrumbs & redirects

asrvd:
## ✅ Checklist

---

## Changelog

- added breadcrumbs
- added missing redirects for site
- added a new contribution guideline

---

_[Screenshots]_

image

image

image

💯

### DIFF
--- a/www/src/components/docs/breadCrumbs.tsx
+++ b/www/src/components/docs/breadCrumbs.tsx
@@ -36,7 +36,7 @@ export default function BreadCrumbs() {
             href={crumb.href}
             className="rounded-lg bg-t3-purple-500/10 p-1 dark:bg-t3-purple-500/20"
           >
-            {crumb.text}
+            {crumb.text[0]?.toUpperCase() + crumb.text.slice(1)}
           </a>
           {index < breadcrumbs.length - 1 && (
             <svg width="16" height="16" viewBox="0 0 24 24">
--- a/www/src/components/docs/breadCrumbs.tsx
+++ b/www/src/components/docs/breadCrumbs.tsx
@@ -46,7 +46,10 @@ export default function BreadCrumbs() {
 
   return (
     <div className="mb-4 flex items-center gap-2 px-2 text-sm">
-      <a href="/">
+      <a
+        href="/"
+        className="rounded-lg bg-t3-purple-500/10 p-1 dark:bg-t3-purple-500/20"
+      >
         <svg width="20" height="20" viewBox="0 0 24 24">
           <path
             fill="currentColor"
--- a/www/src/components/docs/breadCrumbs.tsx
+++ b/www/src/components/docs/breadCrumbs.tsx
@@ -1,4 +1,35 @@
+import { SIDEBAR } from "../../config";
+import { getLanguageFromURL } from "../../languages";
+
+type SlugType = "" | "usage" | "deployment";
+
 export default function BreadCrumbs() {
+  const lang = getLanguageFromURL(window.location.href);
+  const slugToEntryPath = (slug: SlugType) => {
+    switch (slug) {
+      case "":
+        return "Create T3 App";
+      case "usage":
+        return "Usage";
+      case "deployment":
+        return "Deployment";
+    }
+  };
+  const slug =
+    window.location.pathname.slice(1).split("/").length > 2
+      ? window.location.pathname.slice(1).split("/")[1]
+      : "" || "";
+  const actualEntries =
+    SIDEBAR[lang][
+      slugToEntryPath(
+        slug === undefined || slug === "" ? "" : (slug as SlugType),
+      )
+    ];
+
+  const getPathNameFromLink = (link: string) => {
+    return actualEntries?.find((entry) => entry.link === link)?.text;
+  };
+
   const breadcrumbs = window.location.href
     .split("/")
     .slice(window.location.href.split("/").length > 5 ? -2 : -1)
@@ -9,7 +40,7 @@ export default function BreadCrumbs() {
         .join("/");
       return {
         href,
-        text: crumb,
+        text: getPathNameFromLink(href.slice(href.indexOf("en"))) || crumb,
       };
     });
 
--- a/www/src/components/docs/breadCrumbs.tsx
+++ b/www/src/components/docs/breadCrumbs.tsx
@@ -1,0 +1,54 @@
+export default function BreadCrumbs() {
+  const breadcrumbs = window.location.href
+    .split("/")
+    .slice(window.location.href.split("/").length > 5 ? -2 : -1)
+    .map((crumb) => {
+      const href = window.location.href
+        .split("/")
+        .slice(0, window.location.href.split("/").indexOf(crumb) + 1)
+        .join("/");
+      return {
+        href,
+        text: crumb,
+      };
+    });
+
+  return (
+    <div className="mb-4 flex items-center gap-2 px-2 text-sm">
+      <a href="/">
+        <svg width="20" height="20" viewBox="0 0 24 24">
+          <path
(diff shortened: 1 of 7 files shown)


asrvd:
btw, shall i disable the last breadcrumb for each page as the user is already on that one and i made the component `client:only` to make window usable, is there any other better way to achieve this?

juliusmarminge:
can we at least apply some `capitalize` on the items, im guessing it's hard to get the proper capitalization with for example `typescript`

www/src/components/docs/breadCrumbs.tsx
```diff
@@ -0,0 +1,54 @@
...
+          fill="currentColor"
+          fill-rule="evenodd"
+          d="m9.005 4l8 8l-8 8L7 18l6.005-6L7 6z"
+        />
+      </svg>
+      {breadcrumbs.map((crumb, index) => (
+        <div className="flex items-center gap-2">
+          <a
+            href={crumb.href}
+            className="rounded-lg bg-t3-purple-500/10 p-1 dark:bg-t3-purple-500/20"
```
juliusmarminge, line 37: can we at least apply some `capitalize` on the items, im guessing it's hard to get the proper capitalization with for example `typescript`

juliusmarminge:
getting 404 here 

CleanShot 2022-11-26 at 19 10 41

kevinzunigacuellar:
I think this can be migrated to an Astro component using the `Astro.url.pathname`

asrvd:
> getting 404 here
> 
> CleanShot 2022-11-26 at 19 10 41 [ CleanShot 2022-11-26 at 19 10 41 ](https://user-images.githubusercontent.com/51714798/204103030-4075bf73-ec38-400d-9bf9-6411ff8212a8.gif) [ ](https://user-images.githubusercontent.com/51714798/204103030-4075bf73-ec38-400d-9bf9-6411ff8212a8.gif)

thats cus we are redirecting `/usage` but not `/en/usage`, i'll add redirects for these too then

asrvd:
> I think this can be migrated to an Astro component using the api [`Astro.url`](https://docs.astro.build/en/reference/api-reference/#astrourl)

I had made it using astro first, but was getting some weird ts errors which were fixed after changing it to tsx

www/src/components/docs/breadCrumbs.tsx
asrvd: yes, will do

juliusmarminge:
> thats cus we are redirecting /usage but not /en/usage, i'll add redirects for these too then

This won't really be viable when we get a lot of translations. Is there no better way to redirect depending on the current lang or something?

asrvd:
> > thats cus we are redirecting /usage but not /en/usage, i'll add redirects for these too then
> 
> This won't really be viable when we get a lot of translations. Is there no better way to redirect depending on the current lang or something?

middleware? 🤔

juliusmarminge:
> > > thats cus we are redirecting /usage but not /en/usage, i'll add redirects for these too then
> > 
> > This won't really be viable when we get a lot of translations. Is there no better way to redirect depending on the current lang or something?
> 
> middleware? 🤔

Might be easier to just make /<lang>/deployment a valid path?

asrvd:
> > > > thats cus we are redirecting /usage but not /en/usage, i'll add redirects for these too then
> > > 
> > > This won't really be viable when we get a lot of translations. Is there no better way to redirect depending on the current lang or something?
> > 
> > middleware? 🤔
> 
> Might be easier to just make `/<lang>/deployment` a valid path?

and usage too

asrvd:
what about this though? @juliusmarminge 

image

www/src/components/docs/breadCrumbs.tsx
```diff
@@ -0,0 +1,54 @@
...
+    .split("/")
+    .slice(window.location.href.split("/").length > 5 ? -2 : -1)
+    .map((crumb) => {
+      const href = window.location.href
+        .split("/")
+        .slice(0, window.location.href.split("/").indexOf(crumb) + 1)
+        .join("/");
+      return {
+        href,
+        text: crumb,
```
juliusmarminge, line 12: I wonder if we shouldn't grab this text from the sidebar instead? We could make some util that finds the link and maps it to the correct text. (This would also make it work for translations)

Right now there are a few weird ones like
CleanShot 2022-11-28 at 14 54 46@2x
CleanShot 2022-11-28 at 14 54 39@2x
CleanShot 2022-11-28 at 14 54 34@2x

juliusmarminge (changes requested):

asrvd:
> > what about this though? @juliusmarminge
> > image
> 
> doesn't really matter I don't think, either / or /en/introduction

no, I mean wouldnt we need to add one for each lang then?

juliusmarminge:
> > > what about this though? @juliusmarminge
> > > image
> > 
> > 
> > doesn't really matter I don't think, either / or /en/introduction
> 
> no, I mean wouldnt we need to add one for each lang then?

Oh right, well do we really need a redirect here? Doesn't hurt i guess but is https://create.t3.gg/en a thing i'd go to?

Or we can add to the translation guide to add a redirect for `/lang` -> `/lang/introduction` - or maybe we can work with a glob pattern here? `/<any 2 characters>` -> `/<same 2 characters>/introduction`

www/src/components/docs/breadCrumbs.tsx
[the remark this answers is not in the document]
asrvd: done https://github.com/t3-oss/create-t3-app/pull/879/commits/330f419581e5b3f6cdf418095a797f129056c112

www/src/components/docs/breadCrumbs.tsx
asrvd: done via https://github.com/t3-oss/create-t3-app/commit/330f419581e5b3f6cdf418095a797f129056c112

www/src/components/docs/breadCrumbs.tsx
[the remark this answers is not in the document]
asrvd: oh so like the font weight can gradually increase? 🤔

www/src/components/docs/breadCrumbs.tsx
```diff
@@ -0,0 +1,85 @@
...
+          d="m9.005 4l8 8l-8 8L7 18l6.005-6L7 6z"
+        />
+      </svg>
+      {breadcrumbs.map((crumb, index) => (
+        <div className="flex items-center gap-2">
+          <a
+            href={crumb.href}
+            className="rounded-lg bg-t3-purple-500/10 p-1 dark:bg-t3-purple-500/20"
+          >
+            {crumb.text[0]?.toUpperCase() + crumb.text.slice(1)}
```
juliusmarminge, line 70: don't need the capitalization anymore

```suggestion
            {crumb.text}
```

asrvd: we actually need it cus it fallbacks to normal `crumb` text for links it cant find in the sidebar entries like "usage" ad "deplyment"
image

juliusmarminge: oh, we currently get TRPC (and not tRPC) that's why I noticed it

asrvd: need to apply some logic here too then

juliusmarminge: Never mind it's fine for now. Will prob get fixed if we refactor it later

asrvd:
can be merged now ig

asrvd:
can be merged now ig

juliusmarminge (approved):
Nice work. Sorry for all the pushback.

PR #880, opened 2022-11-26
docs: improve language select styling

Coyenn:
Closes #<issue>

## ✅ Checklist

---

## Changelog

- Improved the language select dropdown animation
- Removed margin top from language select button that made it look asymmetrical
- Added a border to the language select dropdown in dark mode
- Increased padding of the language select dropdown to make it look less narrow (I'm not sure if this looks good or not, tell me if it doesn't and I'll remove it 👍)

From (Notice how the language select button is not properly aligned with the rest of the buttons in the header. It has a slight margin top)
Bildschirmfoto 2022-11-26 um 20 05 28

To
Bildschirmfoto 2022-11-26 um 20 05 53

### DIFF
--- a/www/src/components/navigation/LanguageSelect.tsx
+++ b/www/src/components/navigation/LanguageSelect.tsx
@@ -16,7 +16,7 @@ export default function LanguageSelect({ language }: LanguageSelectProps) {
   return (
     <div className="flex items-center gap-2">
       <Listbox value={language} onChange={handleSelect}>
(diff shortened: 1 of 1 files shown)


c-ehrlich (commented):
This looks much better already :)

Just one thing: the height still seems to be different by 1px each on top/bottom
image

c-ehrlich (approved):
Thanks, looks great now :)

Issue #871, opened 2022-11-25
bug: No .env.example in some templates


mtmeyer:
### Provide environment information

  System:
    OS: macOS 12.0.1
    CPU: (8) arm64 Apple M1
    Memory: 92.91 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.2 - ~/.nvm/versions/node/v16.14.2/bin/node
    Yarn: 1.22.17 - /opt/homebrew/bin/yarn
    npm: 8.5.0 - ~/.nvm/versions/node/v16.14.2/bin/npm

### Describe the bug

I noticed when I generate a new project, an `.env.example` file is not always included.  For example if you create a project and select either no additional packages or just trpc, you don't get the example env file.  Not sure if this is expected behaviour but i would assume there is always a potential use case for .env files regardless of which additional packages you install.  
I realise that its super easy to create an env file but the docs refer to this example file and figured it would be nice to have across all template options.

### To reproduce

Run `npx create-t3-app`, select all defaults except just select trpc in the additional packages section.

c-ehrlich:
It's the intended behavior currently because an app with only tRPC doesn't require any environment variables to get started. I can see the case for still including an empty file with just some comments, especially considering https://github.com/t3-oss/create-t3-app/issues/368

Would like to hear some other opinions on this.

Toavina23:
I think that there should be a prompt asking if the user is fine with creating an empty .env.example file

Toavina23:
> I think that there should be a prompt asking if the user is fine with creating an empty .env.example file

but this should only be prompted if no packages that require a non empty .env are selected to be intalled

c-ehrlich:
I don't think we should introduce a prompt for everything. Ship a sensible default, and if people want the opposite they can create/delete the file.

mtmeyer:
I think there's always a possibility you could need env vars regardless of the additional packages installed so I see no harm in adding an example.
If no example is included then I think it might be worth mentioning it in the docs as that's what confused me in the first place. An alternative could also be to just include an example in the docs for when someone goes looking for it?

PR #901, opened 2022-11-30
feat: always include .env and .env.example

c-ehrlich:
Closes https://github.com/t3-oss/create-t3-app/issues/871

## ✅ Checklist

---

## Changelog

- rename `NEXT_PUBLIC_BAR` to `NEXT_PUBLIC_CLIENTVAR` as it's more descriptive and we are already using this name in the docs anyway
- include a sample `.env` and `.env.example` file in all apps, even ones that don't need any env vars initially
- include schema files with those apps as well
- i'm leaning towards this feature but would be ok not merging if it's decided that we don't want this. the way i see it is that the downside is more files in a base app, and the upside is communicating more clearly our default env strategy.

---

## Screenshots

image

💯


cli/template/addons/env/base-schema.mjs
```diff
@@ -0,0 +1,29 @@
+// @ts-check
```
juliusmarminge, line 1: There is already a /env/schema.mjs in the template/base app

juliusmarminge (commented):
Don't think we need to add this new file, just make any necessary modifications here https://github.com/t3-oss/create-t3-app/blob/next/cli/template/base/src/env/schema.mjs

juliusmarminge (approved):

Issue #911, opened 2022-12-03
bug: next-auth doesn't work out of the box without prisma


c-ehrlich:
### Provide environment information

```
  System:
    OS: macOS 12.5.1
    CPU: (8) arm64 Apple M1
    Memory: 82.56 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
    Yarn: 1.22.19 - ~/Documents/dev/node_modules/.bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm
```
create-t3-app v 6.11.0

### Describe the bug

Currently, installing a t3-app with next-auth but without prisma creates a project that doesn't work out of the box and requires significant modification in order to get it working.

<img width="1552" alt="image" src="https://user-images.githubusercontent.com/8353666/205455836-0387aef6-cfcb-4bb8-b3f9-0bb22dd84165.png">

Error:
```
[next-auth][error][JWT_SESSION_ERROR] 
https://next-auth.js.org/errors#jwt_session_error Invalid Compact JWE {
  message: 'Invalid Compact JWE',
  stack: 'JWEInvalid: Invalid Compact JWE\n' +
    '    at compactDecrypt (/Users/ce-dev/Documents/dev/temp/nextauth-without-adapter/node_modules/.pnpm/jose@4.11.1/node_modules/jose/dist/node/cjs/jwe/compact/decrypt.js:16:15)\n' +
    '    at jwtDecrypt (/Users/ce-dev/Documents/dev/temp/nextauth-without-adapter/node_modules/.pnpm/jose@4.11.1/node_modules/jose/dist/node/cjs/jwt/decrypt.js:8:61)\n' +
    '    at Object.decode (/Users/ce-dev/Documents/dev/temp/nextauth-without-adapter/node_modules/.pnpm/next-auth@4.18.0_pknogjuzx4bv7zxtatcb2ahtsq/node_modules/next-auth/jwt/index.js:64:34)\n' +
    '    at async Object.session (/Users/ce-dev/Documents/dev/temp/nextauth-without-adapter/node_modules/.pnpm/next-auth@4.18.0_pknogjuzx4bv7zxtatcb2ahtsq/node_modules/next-auth/core/routes/session.js:41:28)\n' +
[...]

PR #932, opened 2022-12-06
fix: breaking error caused by unstable_getServerSession changes in next-auth 4.18

c-ehrlich:
Closes https://github.com/t3-oss/create-t3-app/issues/911

## ✅ Checklist

---

## Changelog

- spread authOptions into a new object to fix unstable_getServerSession error
- this is probably not the best possible solution so added a FIXME
- unpin next-auth
- update docs & fix outdated naming (`nextAuthOptions` instead of `authOptions`)

---

## Screenshots

image

💯

### DIFF
--- a/.changeset/lucky-kangaroos-tickle.md
+++ b/.changeset/lucky-kangaroos-tickle.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+fix next-auth error
--- a/cli/src/installers/index.ts
+++ b/cli/src/installers/index.ts
@@ -22,10 +22,8 @@ export type AvailablePackages = typeof availablePackages[number];
  */
 export const dependencyVersionMap = {
   // NextAuth.js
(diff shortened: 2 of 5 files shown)


juliusmarminge (commented):
Why does it needs to be spread?

c-ehrlich:
> Why does it needs to be spread?

See Line 142:
https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/next/index.ts#L142

As far as I can tell, this line removes the providers, so it you pass authOptions by reference it does it to the actual object

juliusmarminge (approved):

balazsorban44:
This should be fixed in `4.18.1` and could potentially be reverted! https://github.com/nextauthjs/next-auth/releases/tag/next-auth%40v4.18.1

nexxeln:
cc @c-ehrlich

PR #988, opened 2022-12-17
docs(AR): new folder structure Arabic

Sboonny:
Closes #<issue>

## ✅ Checklist

---

## Changelog

- Changed the file to `.mjx`
- Fixed heading structure.
- Removed ملف and مجلد from the heading because some were misused and describing the wrong state of the folder/file.

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/www/src/pages/ar/folder-structure.mdx
+++ b/www/src/pages/ar/folder-structure.mdx
@@ -21,8 +21,6 @@ import Diagram from "../../components/docs/folderStructureDiagram.astro";
 
 مجلد prisma يحتوي على `prisma.schema` والذي بدوره يحتوي ملف تكوين الاتصال مع قاعدة البيانات كما يحتوي ايضا علي schema التي تحدد شكل الـ tables هذا إلى جانب ملفات الـ migrations، إقراَ [كيفية إستخدام Prisma](/en/usage/prisma)
 
-<sub>(في حالة Prisma)</sub>
-
 ## `public`
 
 مُجلد Public يحتوي على الـ assets التي يقوم الخادم بتقديمها، ملف `favicon.ico` هو مثال على ذلك.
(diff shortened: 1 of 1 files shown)


www/src/pages/ar/folder-structure.mdx
```diff
@@ -6,227 +6,267 @@ lang: ar
...
 
 ملف `trpc.ts` يُسخدم كـ entrypoint لـ tRPC، إقرأ المزيد هنا [tRPC usage](usage/trpc#-utilstrpcts)
 
 <sub>(في حالة tRPC)</sub>
 
-### ملف `.env`
+</div>
+<div>
+
+### `.env`
```
Sboonny, line 215: Random question, should this be heading two? `##`

I think the heading is used here to indicate where the file is, so if it's heading 2, then it's in the root section of the app.

www/src/pages/ar/folder-structure.mdx
```diff
@@ -6,227 +6,267 @@ lang: ar
...
 
 ملف `.env` يُستخدم لتخزين `environment variables`، إقرأ المزيد [Environment Variables](usage/env-variables)
 
-### ملف `.env.example`
+</div>
+<div>
+
+### `.env.example`
```
Sboonny, line 222: Same, and the sections below it

Sboonny (commented):
Random question about the original structure of the page

juliusmarminge:
> Random question about the original structure of the page

I think we did this to make the sizing more appropriate cause there were some overflow when we followed the "actual proper heading numbers" so to say...

juliusmarminge (changes requested):
I think all the `sub` ones should go since now you pick your selections in the Form component

juliusmarminge (approved):

PR #991, opened 2022-12-18
fix: tailwind base styles are duplicated

ronanru:
## ✅ Checklist

---

## Changelog

On the docs site, all tailwind base styles were included twice in the bundled .css files. First are added by the tailwind astro integration, second by `@tailwind base` line in `global.css`.
This PR adds a config to the astro integration to fix that.

---

## Screenshots

image

image

💯

### DIFF
--- a/www/astro.config.mjs
+++ b/www/astro.config.mjs
@@ -47,7 +47,7 @@ export default defineConfig({
   },
   integrations: [
     react(),
-    tailwind(),
+    tailwind({ config: { applyBaseStyles: false } }),
     image({
       serviceEntryPoint: "@astrojs/image/sharp",
     }),


juliusmarminge:
Why is this not default lol???

c-ehrlich:
There's an explanation here, although I do agree this feels like kind of a footgun.

https://docs.astro.build/de/guides/integrations-guide/tailwind/#configapplybasestyles

Thanks for noticing and fixing this :)

c-ehrlich (approved):

juliusmarminge:
> There's an explanation here, although I do agree this feels like kind of a footgun.
> 
> 
> 
> https://docs.astro.build/de/guides/integrations-guide/tailwind/#configapplybasestyles
> 
> 
> 
> Thanks for noticing and fixing this :)

Is it cause we import our css files in the layout?

PR #995, opened 2022-12-20
docs: Fix active color overflow of the Input with the search icon in t…

darknet-00:
Fix active color overflow of the Input with the search icon in the Search component

## ✅ Checklist

---

## Changelog

This is really a small fix. The problem was that when the input is active, the active color overflows (clashes) the search icon in the Search modal component. I will provide screenshots illustrating the before and after state of the UI. I couldn't find an open issue for this and I'm not sure If I need to open one before trying to contribute at all. If that's the case, I'll be happy to follow that rule.

---

## Screenshots

Current state: 
image

After the change:
image

PNPM check:
image

### DIFF
--- a/www/src/styles/algolia/style.css
+++ b/www/src/styles/algolia/style.css
@@ -171,6 +171,9 @@ html.dark {
   margin: 0;
   padding: 0;
 }
+.DocSearch-Reset {
+  margin-inline-start: 12px;
+}
 .DocSearch-MagnifierLabel,
 .DocSearch-Reset {
   align-items: center;
@@ -231,6 +234,7 @@ html.dark {
   display: none;
 }
 .DocSearch-Dropdown {
+  margin-top: 10px;
   max-height: calc(
     var(--docsearch-modal-height) - var(--docsearch-searchbox-height) -
       var(--docsearch-spacing) - var(--docsearch-footer-height)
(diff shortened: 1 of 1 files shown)


www/src/styles/algolia/style.css
```diff
@@ -63,6 +63,7 @@ html.dark {
 }
 .DocSearch-Search-Icon {
   stroke-width: 1.6;
+  margin-right: 15px;
```
c-ehrlich, line 66:
```suggestion
  margin-inline-end: 15px;
```

This doesn't work in RTL languages as it places the margin on the wrong side. Could be fixed by using inline direction instead.

image

c-ehrlich (changes requested):
Thanks for the PR, it's an improvement already. That being said now that this is being worked on I think it could also be improved on the "X" icon side and the bottom.

image

www/src/styles/algolia/style.css
darknet-00: Thanks for the review. 

Oh yes, I didn't think about the RTL languages, thanks for your suggestion.

All of the mentioned issues should be fixed with the latest changes.

c-ehrlich (approved):
Looks great in both ltr and rtl, thanks :)

c-ehrlich:
it seems like you have maintainer write access disabled in your fork so i can't merge the latest next branch into it

darknet-00:
> it seems like you have maintainer write access disabled in your fork so i can't merge the latest next branch into it

It should be enabled now, I'm sorry about that.

Issue #975, opened 2022-12-14
bug: docs - Docker node:16-alpine requires setting `engineType = "binary"` for Prisma to work


nicolomaioli:
### Provide environment information

  System:
    OS: Linux 6.0 Fedora Linux 37 (Workstation Edition)
    CPU: (16) x64 AMD Ryzen 7 PRO 6850U with Radeon Graphics
    Memory: 23.97 GB / 30.12 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.19.0 - ~/.volta/tools/image/node/16.19.0/bin/node
    npm: 8.19.3 - ~/.volta/tools/image/node/16.19.0/bin/npm

ct3aMetadata.initVersion: 6.11.3

### Describe the bug

I've added a `Dockerfile` (copy-pasted from the docs, using `node:16-alpine` as a base image) and `docker-compose.yml` to a freshly bootstrapped app. I modified `index.tsx` to run `trpc.example.getAll.useQuery()`, and received this error message at runtime:

```
Listening on port 3000 url: http://localhost:3000
prisma:error
Invalid `prisma.example.findMany()` invocation:


Unable to load Node-API Library from /app/node_modules/.prisma/client/libquery_engine-linux-musl.so.node, Library may be corrupt
```

- I got a similar error with `node:16-buster-slim` (pointing to the Debian engine).
- Changing the base image to `node:16-buster` fixes the error.
- Adding `engineType = "binary"` to Prisma `generator client` also fixes the error under `node:16-alpine`.

### To reproduce

- Bootstrap application with `trpc` and `prisma`:

```
npm create t3-app@6.11.3
```

- Follow steps 1-3 [from the docs](https://create.t3.gg/en/deployment/docker)
- Add `trpc.example.getAll.useQuery()` to `index.tsx`
- Use the following minimal `docker-compose.yml`:

```
version: "3.9"
services:
  app:
    platform: "linux/amd64"
    build:
      context: .
      dockerfile: Dockerfile
      args:
        NEXT_PUBLIC_CLIENTVAR: "clientvar"
    working_dir: /app
    ports:
      - "3000:3000"
    image: t3-app
    environment:
      - DATABASE_URL=mysql://foo:password@db:3306/bar
  db:
    image: mysql:8
    command: --default-authentication-plugin=mysql_native_password
    restart: always
    ports:
      - 3306:3306
    environment:
      MYSQL_USER: foo
[...]

PR #1006, opened 2022-12-24
docs: recommend `node:16-alpine3.16`

nicolomaioli:
Closes #<issue>

## ✅ Checklist

---

## Changelog

- Recommend `node:16-alpine3.16` as a base image for Docker, as per issue #975
- Affects all translations

💯

### DIFF
--- a/www/src/pages/ar/deployment/docker.md
+++ b/www/src/pages/ar/deployment/docker.md
@@ -66,7 +66,7 @@ README.md
 ```docker
 ##### DEPENDENCIES
 
-FROM --platform=linux/amd64 node:16-alpine AS deps
+FROM --platform=linux/amd64 node:16-alpine3.16 AS deps
 RUN apk add --no-cache libc6-compat openssl1.1-compat
 WORKDIR /app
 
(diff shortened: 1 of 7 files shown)


c-ehrlich:
Thanks for the PR :)

I think it would also be nice to have an explanation in the docs of why we're pinning Alpine to 3.16, maybe with a link to the issue. I'm not sure what's the best approach in terms of adding this but also making it easy for translators to keep their languages up to date. This is the first time we've had that situation, so decided to ask the translators what they think. Will wait for some comments and then let you know.

image

nicolomaioli:
Sounds good! I just replied on Discord so I can see how the conversation evolves 🙏

c-ehrlich (approved):
Let's get this merged the way it is so that peoples docker deployments stop breaking, and then you can open a separate PR adding an explanation of why we are pinning alpine. This will then "invalidate" this article for the other languages.

I will hopefully set up code owners for translation next month, from then on there should be an easier workflow for this. Am also considering moving the code blocks to their own files as was discussed on Discord.

nicolomaioli:
Awesome, thanks! I'll get something going over the next couple of days, soon as the holiday frenzy settles.

Issue #1002, opened 2022-12-23
bug (docs): language selector doesn't use custom scrollbar


c-ehrlich:
### Provide environment information

.

### Describe the bug

<img width="309" alt="image" src="https://user-images.githubusercontent.com/8353666/209333804-bcbe6fbd-0bc5-4b71-a2f1-13f65fd0b2ae.png">

### To reproduce

go to the website and click the language selector

### Additional information

There is more than one possible solution for this:
1. Display the languages in a grid
2. Make the container as tall as the content (I doubt we will get enough languages that this will become a problem)
3. Use the T3 scrollbar
    - if this is chosen, please make sure that it is also on the correct (right) side in Arabic

If you decide to implement this, feel free decide yourself which is better or propose another solution :)

PR #1015, opened 2022-12-26
docs: make scrollbar show on right in RTL

AyanavaKarmakar:
## ✅ Checklist

---

## Changelog

Refer: #1002 

image

- Before

image

- After

image

💯

### DIFF
--- a/www/src/components/navigation/LanguageSelect.tsx
+++ b/www/src/components/navigation/LanguageSelect.tsx
@@ -2,6 +2,7 @@ import { KnownLanguageCode, KNOWN_LANGUAGES } from "../../config";
 import { Listbox, Transition } from "@headlessui/react";
 import clsx from "clsx";
 import { Fragment } from "react";
+import { getIsRtlFromLangCode } from "../../languages";
 
 type LanguageSelectProps = {
   language: KnownLanguageCode;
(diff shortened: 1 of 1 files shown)


juliusmarminge:
Why should it be on the right? The main scrollbar is on the left, no?

AyanavaKarmakar:
> Why should it be on the right? The main scrollbar is on the left, no?

All the scrollbars are on the right in RTL like in LTR. Shouldn't the language selector also be on the right, like @c-ehrlich said?

image

juliusmarminge:
Oh my bad then. I thought they were on the left but must have remembered wrong.

c-ehrlich:
> Oh my bad then. I thought they were on the left but must have remembered wrong.
> 
> Are we doing something custom here then? Or why doesnt the browser handle this correctly?

Scrollbars on the right side in RTL seems to be standard as far as I can tell. Here's an example from the React docs. It seems they are using the same "hack" of switching the container that has the scrollbar to LTR, and then the stuff inside it back to RTL.

image

As to why browsers don't do this if it's the "standard", no idea.

Sboonny:
image

> All the scrollbars are on the right in RTL like in LTR. Shouldn't the language selector also be on the right

I think this is caused by confusion. Scroll bars are on the right in RTL. We are used to it there because of Google and Facebook positioning them there

c-ehrlich (commented):
lets get this merged :)

c-ehrlich (approved):

PR #1018, opened 2022-12-26
refactor(docs): split config out per language

c-ehrlich:
## ✅ Checklist

---

## Changelog

- split `config.ts` into per-language files that are inside `pages/<lang>/_config/config<lang>.ts`

The purpose of this is to have them within the folder that the code owners of each language get. I'm open to a different file/naming structure etc but will need to be in `pages/<lang>` to work

💯


juliusmarminge:
Can we close this @c-ehrlich or is this still something we want to do?

c-ehrlich:
it still makes sense imo (but not very important). but at this point the pr is so old that its probably easier to remake it

Issue #1000, opened 2022-12-21
feat: make env setup work in middleware


iduuck:
### Is your feature request related to a problem? Please describe.

I am currently facing a problem, when using the environment variable setup from t3 and I am opening this issue, since the other issue about environment variables not working (#999) was closed by the author.

### Describe the solution you'd like to see

It would be necessary (for now – don't know, if this is being fixed, though) to also compose the server env vars, like we do for the client env vars.

### Describe alternate solutions

non that i can think of.

c-ehrlich:
Could you explain more precisely what you are trying to do? Accessing env vars in middleware works just as it would anywhere else, see: https://github.com/c-ehrlich/mwenv/commit/492bef23be5c3b2a063d4ef07fa96d4f07529359

iduuck:
Sorry, not speaking of trpc middleware, but of Next.js middleware.

c-ehrlich:
Just tried and can confirm that this currently doesn't work, the import makes Next middleware blow up. Accessing the environment variables in a non-typesafe way (ie `process.env.SERVERVAR` does work, but obviously isn't ideal.

What would we need to do to get it to work, and are there any downsides?

iduuck:
I think the only way to work around this, is to define the server env vars the same way, as we do with the client env vars (explicitly building our env object before verifying).

iduuck:
fixed by #1021

Issue #404, opened 2022-09-06
feat: extend process.env types as well from envSchemas


Nsttt:
### Is your feature request related to a problem? Please describe.

Currently in order to get autocompletion and type-safety there's the clientEnv and serverEnv which we can import to use our environment variables.
I don't dislike this approach but I believe they're a better developer experience if we allow to use `process.env` instead, or at least improve it through extending global types.

### Describe the solution you'd like to see

We can probably extend it just like this.

```ts
// env.d.ts
import { clientSchema, serverSchema } from "./schema.mjs";
import { z } from 'zod';

declare global {
  namespace NodeJS {
    interface ProcessEnv extends z.infer<typeof clientSchema & typeof serverSchema> {}
  }
}
```

### Desribe alternate solutions

Due to separation of concerns in NextJS server and client variables this might not be the best approach, but rather an alternative to import env variables from files. 

Not really sure how can we achieve the same with `process.env` but if someone have any idea that would be a plus.

c-ehrlich:
I like the idea of this as I think many users are probably just ignoring the typesafe env vars and using process.env instead because that's what they're used it. This way type safety is the default.

I tried your code snippet and didn't get autocomplete on environment variables. Do you have a working sample app where you implemented this?

Nsttt:
> I tried your code snippet and didn't get autocomplete on environment variables. Do you have a working sample app where you implemented this?

Sure, it worked just fine for me [here ](https://github.com/Nsttt/create-t3-env-test)with this repo. Just a brand new t3-app that has that file.

c-ehrlich:
Can confirm this works, not sure I was doing wrong earlier.

Looks great. Do you want to open a PR for it? I think it should be reviewed by as many people as possible because this is such a core part of the app, but if it works without caveats then I think it's a big win.

Nsttt:
I think it should be tested a bit more.
Totally open to make a PR about it. But for example in that specific example, we can see that `NEXTAUTH_URL` still has the `string | undefined` type. Probably due to checking it with `z.string().url()`.

I'll give it a look tomorrow a bit more and I'll open a PR or update what I find here. 😄

c-ehrlich:
Another downside I can see is that you get autocomplete on all env variables everywhere, even though they don't exist everywhere (ie non-`NEXT_PUBLIC` in the frontend). This is a nice thing that was solved by splitting it into server and client.

But the existing schema would still remain of course.

[...]

PR #1021, opened 2022-12-27
feat: construct `serverEnv`, like we do with `clientEnv` #1000

iduuck:
Closes #<issue>

## ✅ Checklist

---

## Changelog

- see #1000 

💯


juliusmarminge (changes requested):
We shouldnt need 2 separate objects right? Destruction can be in a single object shared with both server and client?

Ideally some comment saying you only need manual destruction for certain use-cases, and not require all envs to be destructured, i.e:

```ts
// schema.mjs
const destructured = {
  FOO: process.env.FOO,
  NEXT_PUBLIC_BAR: process.env.NEXT_PUBLIC_BAR,
}

const serverSchema = z.object({
  FOO: z.string(),
  BAZ: z.coerce.boolean(),
});

// server.mjs
serverSchema.parse({
  ...process.env,
  ...destructured
});
```

iduuck:
I get what you mean, @juliusmarminge. However, I think like this it could be misleading. What we could do, is using the `clientEnv` and extending the `serverEnv` with it.

Like:
```
const clientEnv = { /* ... */ };
const serverEnv = { ...clientEnv, BAZ: process.env.BAZ };
```

**Also:** We would be facing some issues with the validation on the middleware-side. When we only add the middleware environment variables one-by-one, it would be giving us schema validation errors, when using on the middleware (because – in your case – `BAZ` wouldn't be available at the runtime of the middleware).

juliusmarminge:
Fair points. I didnt think of that. I just want to make sure we provide as good DX as possible as many will think its annoying to do this many steps "just to get rid of `| undefined` 🤔

iduuck:
What `| undefined` are you referring to, the one in the declaration of `serverEnv`?

Btw: Do you know, why the build is failing with a wrong import (working for me locally).

juliusmarminge:
> What `| undefined` are you referring to, the one in the declaration of `serverEnv`?

So I guess where I'm going is if I need to do too much work to get it running in the first place, maybe I'd be fine with using `process.env` in edge-runtimes and just take the non-null assertion - or maybe we could get #404 working which would type it properly.

```ts
// in some edge function, e.g. 'src/middleware.ts'
const baz1 = process.env.BAZ;
//    ^? -> string | undefined


const baz2 = process.env.BAZ!;
//^? -> string
```

Either would be fine since the variable have been validated to be present during build time.

Maybe I'm drifting here and this isn't a problem for others? I just wanna make sure we don't make it too much work to get environment variables to work or else people might just nuke it all and use `process.env` for everything 🤔 

> Btw: Do you know, why the build is failing with a wrong import (working for me locally).

You haven't defined the `serverEnv` object in the other schema files, just the base one.

iduuck:
I personally wouldn't want to use the non-null assertion throughout my application. Also for the purpose of the env validation, that t3 offers, I am sure there are use-cases, which would be awesome to have default's for.

So, just using `process.env` isn't feasable (again, personally).

juliusmarminge:
Yea all valid points - curious how many apps use the `middleware.ts` file and if it's warranted to include this by default - the `schema.mjs` file is already kinda scary as it is 🤔 Although I guess this will get more and more relevant as edge runtime becomes more popular to use for more stuff... Maybe we should file an issue to https://github.com/vercel/next.js and see why this is the case - maybe they are open to change it?

iduuck:
I think [this issue](https://github.com/vercel/next.js/issues/30612#issuecomment-955065804) is giving us the reason.

> This is related to how _middleware.ts is webpack transformed by the middleware-plugin.ts.
Only when you access an env var directly you will get the value. You can't destructure from process.env or get the whole object. I am not sure if this behaviour is intended or should be mentioned in the docs.

I think this is due to the nature of the `DefinePlugin` of webpack. However, with turbopack, I am curious, if this is working?

iduuck:
What's the plan to go for? Asking Next.js team first and implementing the PR for now? Or are we not wanting to implement the PR as is?

juliusmarminge:
> What's the plan to go for? Asking Next.js team first and implementing the PR for now? Or are we not wanting to implement the PR as is?

I think this is the best approach even though I'm not a fan of the extra step - but that seems to be out of our hands unfortunately...

Go ahead and fix this up and we can ship it if no-one else has any objections? @c-ehrlich @nexxeln

nexxeln:
> > What's the plan to go for? Asking Next.js team first and implementing the PR for now? Or are we not wanting to implement the PR as is?
> 
> I think this is the best approach even though I'm not a fan of the extra step - but that seems to be out of our hands unfortunately...
> 
> Go ahead and fix this up and we can ship it if no-one else has any objections? @c-ehrlich @nexxeln

Sounds good to me

iduuck:
OK, guys. 
- [x] I updated the code at the docs (unfortunately, I am only speaking english of all the doc languages, so I was not able to change the text for the other languages)
- [x] Synchronized upstream
- [x] Fixed all issues from the tests

juliusmarminge (approved):
👍👍

juliusmarminge:
Seems like you've disabled maintainers edit priviledges so i cant update your branch

c-ehrlich:
> > What's the plan to go for? Asking Next.js team first and implementing the PR for now? Or are we not wanting to implement the PR as is?
> 
> I think this is the best approach even though I'm not a fan of the extra step - but that seems to be out of our hands unfortunately...
> 
> Go ahead and fix this up and we can ship it if no-one else has any objections? @c-ehrlich @nexxeln

Bit late but I agree with this

iduuck:
> Seems like you've disabled maintainers edit priviledges so i cant update your branch

How am I able to enable this? This is a repository from my company, so could be that this is the the case company-wide? Couldn't find a way to enable this?

juliusmarminge:
Should be on the right side where you request reviews etc

iduuck:
CleanShot 2023-01-08 at 19 51 22@2x
 unfortunately nothing there, would have activated this then ...

juliusmarminge:
It's there when you create the PR. Should be a way to update it after creation too??

CleanShot 2023-01-08 at 19 55 14

PR #1032, opened 2022-12-30
feat: make the reference to create t3 app consistent 

Sboonny:
Closes #<issue>

## ✅ Checklist

---

## Changelog

Sometimes `create-t3-app` was wrapped in code elements, and sometimes it didn't. For consistency, I added the code elements around it, whenever it is in a description section.

---

## Screenshots

in `/en/folder-structure#tsconfigjson` path

before 
image

After
image

💯


www/src/pages/en/usage/env-variables.md
```diff
@@ -5,7 +5,7 @@ layout: ../../../layouts/docs.astro
 lang: en
 ---
 
-Create-T3-App uses [Zod](https://github.com/colinhacks/zod) for validating your environment variables at runtime _and_ buildtime by providing some additional files in the `env`-directory:
+`create-t3-app` uses [Zod](https://github.com/colinhacks/zod) for validating your environment variables at runtime _and_ buildtime by providing some additional files in the `env`-directory:
```
Sboonny, line 8: I am conflicted about this one 🤔, maybe it's better to leave it as is.

```suggestion
Create-T3-App uses [Zod](https://github.com/colinhacks/zod) for validating your environment variables at runtime _and_ buildtime by providing some additional files in the `env`-directory:
```

c-ehrlich:
I know this is currently done inconsistently in the docs, so maybe we can use this thread to discuss what makes sense. Here's some general thoughts:
- It should be "Create T3 App" or `create-t3-app`. Not "Create-T3-App" or anything else.
- I don't think either of the two is _always_ the right version, so it makes sense to use both based on which feels better in any given paragraph
- There are probably some where currently the best choice isn't being used.

Sboonny:
> There are probably some where currently the best choice isn't being used.

🤔 Perhaps the links and the headings.

Sboonny:
Oh btw, before I start changing stuff.

> maybe we can use this thread to discuss what makes sense.

What about this, if `create-t3-app` refers to the CLI, we use `create-t3-app`. If it's refering to the stack, we use Create T3 App.

juliusmarminge:
> Oh btw, before I start changing stuff.
> 
> 
> 
> > maybe we can use this thread to discuss what makes sense.
> 
> 
> 
> What about this, if `create-t3-app` refers to the CLI, we use `create-t3-app`. If it's refering to the stack, we use Create T3 App.

Makes sense to me.

c-ehrlich:
> Oh btw, before I start changing stuff.
> 
> > maybe we can use this thread to discuss what makes sense.
> 
> What about this, if `create-t3-app` refers to the CLI, we use `create-t3-app`. If it's refering to the stack, we use Create T3 App.

I agree with this in general - the project itself is "Create T3 App", the cli/npm package is `create-t3-app`. I expect you'll find that there are some cases where it's difficult to separate the two, but if you want to go over the docs and give your opinions on this I think we'll end up with something that feels more consistent than what we have now :)

www/src/pages/en/usage/next-auth.md
```diff
@@ -131,7 +131,7 @@ const userRouter = router({
...
+Getting NextAuth.js to work with Prisma requires a lot of [initial setup](https://next-auth.js.org/adapters/models/). Create T3 App handles all of this for you, and if you select both Prisma and NextAuth.js, you'll get a fully working authentication system with all the required models preconfigured. We ship your scaffolded app with a preconfigured Discord OAuth provider, which we chose because it is one of the easiest to get started with - just provide your tokens in the `.env` and you're good to go. However, you can easily add more providers by following the [NextAuth.js docs](https://next-auth.js.org/providers/). Note that certain providers require extra fields to be added to certain models. We recommend you read the documentation for the provider you would like to use to make sure you have all the required fields.
```
Sboonny, line 134: I am conflicted about this one, because I feel that it could be `create-t3-app`. 🤔 I am not native English speaker.

```suggestion
Getting NextAuth.js to work with Prisma requires a lot of [initial setup](https://next-auth.js.org/adapters/models/). `create-t3-app` handles all of this for you, and if you select both Prisma and NextAuth.js, you'll get a fully working authentication system with all the required models preconfigured. We ship your scaffolded app with a preconfigured Discord OAuth provider, which we chose because it is one of the easiest to get started with - just provide your tokens in the `.env` and you're good to go. However, you can easily add more providers by following the [NextAuth.js docs](https://next-auth.js.org/providers/). Note that certain providers require extra fields to be added to certain models. We recommend you read the documentation for the provider you would like to use to make sure you have all the required fields.
```

Sboonny (commented):
I will be honest, I was lazy and decided, if it "creates" or "sets up", then it's the CLI. 😄 

I am open to suggestions.

juliusmarminge (approved):

PR #1036, opened 2022-12-31
feat: use typescript-eslint's recommended-requiring-type-checking

### LINKED ISSUE #1035: feat: use typescript-eslint's recommended-requiring-type-checking

https://github.com/t3-oss/create-t3-app/blob/88ce6e50289281e009f1eea2223dbef4cb86df61/cli/template/base/.eslintrc.json right now extends from `plugin:@typescript-eslint/recommended`. But per [typescript-eslint.io's docs on typed linting](https://typescript-eslint.io/linting/typed-linting) there's actually a second recommended ruleset that can be enabled: `plugin:@typescript-eslint/recommended-requiring-type-checking` (RRTC).

The RRTC config isn't enabled by default because typed linting comes with some performance drawbacks. But it includes many of the particularly powerful rules in typescript-eslint. See https://typescript-eslint.io/rules for a full list.

https://github.com/trpc/trpc/pull/3482 is a pretty good reference point:

* Add `"plugin:@typescript-eslint/recommended-requiring-type-checking"` to the `extends` in the ESLint config file
* Have `tsconfig.eslint.json` extend from `tsconfig.json`, to pull in relevant compiler options
* Disable rules that are either undesirable or have way too many existing non-auto-fixable complaints
* Fix up any remaining complaints

Going without the fancier, stronger rules, I suppose? 🤔 

### Additional information

See #317 for typescript-eslint's introduction.

JoshuaKGoldberg:
Closes #1035

## ✅ Checklist

---

## Changelog

Augments the internal/CLI and user-generated ESLint configs to extend from [`plugin:@typescript-eslint/recommended-requiring-type-checking`](https://typescript-eslint.io/linting/configs#recommended-requiring-type-checking).

---

## Screenshots

...n/a? 

💯


www/src/utils/fetchGithub.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-assignment */
```
c-ehrlich, line 1: We could get rid of these disabled rules by
1. Casting the GitHub token as `string` (why is this not the default?)
2. Using Zod to validate the response. Is that worth it? cc @nexxeln @juliusmarminge

www/src/utils/pagination.ts
```diff
@@ -1,8 +1,8 @@
-import { SIDEBAR, Sidebar } from "../config";
+import { SIDEBAR, Sidebar, SidebarItem, SidebarItemLink } from "../config";
 
-export function paginate(lang: keyof Sidebar, path: string) {
-  const routes = Object.values(SIDEBAR[lang]).flat();
-  const index = routes.map((item) => item.link as string).indexOf(path);
+export function paginate(lang: keyof Sidebar, path: SidebarItemLink) {
+  const routes = Object.values(SIDEBAR[lang]).flat() as SidebarItem[];
```
c-ehrlich, line 4: Why does using `Object.values` cast the array to `any` here? Maybe there is some way to fix the type at the source so we can avoid `as`?

cli/src/utils/renderVersionWarning.ts
```diff
@@ -47,7 +52,7 @@ function checkForLatestVersion(): Promise<string> {
             let body = "";
             res.on("data", (data) => (body += data));
             res.on("end", () => {
-              resolve(JSON.parse(body).latest);
+              resolve((JSON.parse(body) as DistTagsBody).latest);
```
c-ehrlich, line 55: Could use Zod here also?

cli/template/base/.eslintrc.json
```diff
@@ -1,4 +1,22 @@
...
+    {
+      "extends": [
+        "plugin:@typescript-eslint/recommended-requiring-type-checking"
+      ],
+      "files": ["*.ts", "*.tsx"],
+      "parserOptions": {
+        "project": "tsconfig.json"
+      },
+      "rules": {
+        // Consider enabling these rules if you'd like a bit more strict linting
```
c-ehrlich, line 12: I would be ok with having all of these on by default, but I guess we need to decide how strict we want our defaults to be. What does everyone else think?

c-ehrlich (commented):
Looks great, thanks for putting in all this work! After reading back my comments, some of them don't really relate to the PR itself I guess, but this seems like as good a place as any to bring them up.

I'd basically be fine to approve this as is, but want to leave it open for a few days to allow everyone else to read it and discuss.

www/src/utils/fetchGithub.ts
nexxeln: casting is fine imo

cli/template/base/.eslintrc.json
bai: I'd be ok having stricter defaults since this is a new project template. Disabling rules would be ok for me if I have to later on.

cli/src/utils/addPackageDependency.ts
```diff
@@ -22,10 +22,8 @@ export const addPackageDependency = (opts: {
     const version = dependencyVersionMap[pkgName];
 
     if (devMode) {
-      //eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
juliusmarminge, line 25: eeehrm I ran `lint --fix` and it removed these, linter passes but editor complains 🤔 
CleanShot 2023-01-03 at 12 38 24@2x

www/src/utils/fetchGithub.ts
juliusmarminge: fixed em'

cli/package.json
```diff
@@ -31,7 +31,7 @@
     "dev": "tsup --watch",
     "clean": "rm -rf dist .turbo node_modules",
     "start": "node dist/index.js",
-    "lint": "eslint src --cache --cache-strategy content",
```
juliusmarminge, line 34: this cache doesn't do anything in CI since the file's not persisted. We're better off with turbo's remote cache

cli/template/extras/src/pages/api/trpc/[trpc].ts
```diff
@@ -11,7 +11,9 @@ export default createNextApiHandler({
   onError:
     env.NODE_ENV === "development"
       ? ({ path, error }) => {
-          console.error(`❌ tRPC failed on ${path}: ${error}`);
+          console.error(
+            `❌ tRPC failed on ${path ?? "<no-path>"}: ${error.message}`,
```
Comment (juliusmarminge) [inline cli/template/extras/src/pages/api/trpc/[trpc].ts:15]: fixes @typescript-eslint/restrict-template-expressions

cli/template/extras/src/pages/index/with-auth-trpc-tw.tsx
```diff
@@ -74,7 +74,7 @@ const AuthShowcase: React.FC = () => {
       </p>
       <button
         className="rounded-full bg-white/10 px-10 py-3 font-semibold text-white no-underline transition hover:bg-white/20"
-        onClick={sessionData ? () => signOut() : () => signIn()}
+        onClick={sessionData ? () => void signOut() : () => void signIn()}
```
juliusmarminge, line 77: fixes @typescript-eslint/no-missused-promises

cli/template/extras/src/server/api/trpc/base.ts
```diff
@@ -32,7 +32,7 @@ type CreateContextOptions = Record<string, never>;
  * - trpc's `createSSGHelpers` where we don't have req/res
  * @see https://create.t3.gg/en/usage/trpc#-servertrpccontextts
  */
-const createInnerTRPCContext = async (_opts: CreateContextOptions) => {
+const createInnerTRPCContext = (_opts: CreateContextOptions) => {
```
juliusmarminge, line 35: fixes @typescript-eslint/require-await

juliusmarminge:
@JoshuaKGoldberg I think I fixed most of the stuff. There's some weird config stuff left though (see https://github.com/t3-oss/create-t3-app/pull/1036#discussion_r1060505136). If you want to take a look at that, it would be highly appreciated

iduuck:
Just to put my two cents onto this PR. I am **not** against this PR, but we should be aware, that when we are using the `--fix` command in a generated project, it can happen, that an `type` is going to be replaced with an `interface`, which is being wrong as per https://github.com/trpc/trpc/issues/3453

www/src/utils/pagination.ts
JoshuaKGoldberg: The actual type of `Object.values` (which you can skip to with go-to-definition) is:

```ts
{
    /**
     * Returns an array of values of the enumerable properties of an object
     * @param o Object that contains the properties and methods. This can be an object that you created or an existing Document Object Model (DOM) object.
     */
    values<T>(o: { [s: string]: T } | ArrayLike<T>): T[];

    /**
     * Returns an array of values of the enumerable properties of an object
     * @param o Object that contains the properties and methods. This can be an object that you created or an existing Document Object Model (DOM) object.
     */
    values(o: {}): any[];
}
```

In other words, `Object.values` has two type system understandings:

1. If the provided `o` has an index signature with values of type `T`, then the returned type is `T[]`
2. `any`

The type of `SIDEBAR[lang]` is:

```ts
{
    "Create T3 App"?: SidebarItem<"ar">[];
    Usage?: SidebarItem<"ar">[];
    Deployment?: SidebarItem<"ar">[];
} | {
    "Create T3 App"?: SidebarItem<"en">[];
    Usage?: SidebarItem<...>[];
    Deployment?: SidebarItem<...>[];
} | ... 4 more ... | {
    ...;
}
```

There's no index signature there. So we fall back to `any`. 😞 

https://github.com/microsoft/TypeScript/issues/38520 has some fun context around why things are the way they are.

cli/src/utils/renderVersionWarning.ts
JoshuaKGoldberg: That would make sense! Basically anywhere that `JSON.parse`s 😄

cli/src/utils/addPackageDependency.ts
JoshuaKGoldberg: Hmm... I don't know why this is happening, but my first guess something spooky with the circular dependency:

1. `addPackageDependency.ts` imports `dependencyVersionMap` from `installers/index.ts`
2. `installers/index.ts` imports `nextAuthInstaller` from `nextAuth.ts`
3. `nextAuth.ts` imports `addPackageDependency` from `addPackageDependency.ts`

Circular dependencies are a no-no in general and especially with TypeScript tooling. I moved the map to its own file (`713c5c3`) and... still get the `any`s. But at least they're on the command-line now too!

cli/template/extras/src/pages/index/with-auth-trpc-tw.tsx
JoshuaKGoldberg: Are you sure it's ok for those promises to be floating? Is it possible for them to reject in a way the app should handle?

cli/src/utils/addPackageDependency.ts
JoshuaKGoldberg: Yeah I'm stuck here. TypeScript's type checker is still giving typescript-eslint an `any` type even after clearing caches / renaming a bunch of stuff around. 

On the surface this issue is behaving similarly to https://github.com/typescript-eslint/typescript-eslint/issues/4689 -> https://github.com/microsoft/TypeScript/issues/48313. So I'd suggest waiting until TypeScript 5 is released, and checking if the `eslint-disable`s added in 73ed575 are still needed. If they are, then this deserves its own bug - probably on typescript-eslint.

JoshuaKGoldberg:
> that an `type` is going to be replaced with an `interface`

That's from the default rule configurations: specifically how [`@typescript-eslint/consistent-type-definitions`](https://typescript-eslint.io/rules/consistent-type-definitions) is configured. That rule isn't enabled in the `recommended` or `recommended-requiring-type-checking` presets. So this PR shouldn't change that. 🙂

bai (approved):

cli/template/extras/src/pages/index/with-auth-trpc-tw.tsx
juliusmarminge: pretty sure this is fine - I usually do this.

I think the only scenario which you shouldn't do this is if you're using credentials

juliusmarminge:
@JoshuaKGoldberg I'm getting tons of yellow squigglies in VSCode but no reported errors in the CLI:
CleanShot 2023-01-07 at 00 51 20

CleanShot 2023-01-07 at 00 51 27

CleanShot 2023-01-07 at 00 51 32@2x

juliusmarminge:
This is a super weird one... says unsafe assignment of type any but it says the variable is a string... 🤯 
CleanShot 2023-01-07 at 00 59 45

JoshuaKGoldberg:
@juliusmarminge yes, exactly! What you're seeing is a bug: TypeScript says one thing (e.g. `string`) in both the `tsc` CLI and in the IDE, but when typescript-eslint asks it for, the type is reported as `any`. There's a good chance this will be fixed in TypeScript v5. But I'm not sure.

In the meantime, I'll just disable the rules with a sad comment.

juliusmarminge (approved):
Awesome job. Thanks a lot!

cli/template/extras/src/server/api/trpc/with-auth-prisma.ts
```diff
@@ -53,7 +53,7 @@ export const createTRPCContext = async (opts: CreateNextContextOptions) => {
   // Get the session from the server using the unstable_getServerSession wrapper function
   const session = await getServerAuthSession({ req, res });
```
MartelliEnrico, line 54: Since this invocation is async, we can't actually remove the async modifier from the method.

MartelliEnrico: You're right, sorry about that. But the problem still remains since [this change](https://github.com/t3-oss/create-t3-app/pull/1060/files#diff-c4d6c739a033f78862d3c37155af6024deacbf7e77c7e8aa47217d9d30acaa78R68) removed the Awaited<ReturnType<...>> from the trpc context call, even in the auth version.

juliusmarminge: Yea that's new from trpc 10.8.1 that you can pass either an object, or a sync/async function to it and trpc will unwrap the type automatically

https://github.com/trpc/trpc/pull/3495

PR #1044, opened 2023-01-03
fix: session not available on client when passed from server

### LINKED ISSUE #1039: bug: NextAuth session is undefined when passed as a prop from getServerSideProps
#### System:
- OS: Linux 5.10 Ubuntu 20.04.5 LTS (Focal Fossa)
- CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor
- Memory: 6.50 GB / 7.74 GB
- Container: Yes
- Shell: 5.0.17 - /bin/bash
#### Binaries:
- Node: 18.0.0 - ~/.nvm/versions/node/v18.0.0/bin/node
- npm: 8.6.0 - ~/.nvm/versions/node/v18.0.0/bin/npm

*Note: I'm using Windows Subsystem for Linux*

### Describe the bug

```tsx
const NewRecipe: NextPage<
  InferGetServerSidePropsType<typeof getServerSideProps>
> = ({ session }) => {
  const router = useRouter();

  console.log(session); // this logs undefined to the console

  return (
    <div className='h-[calc(100vh_-_var(--navbar-height))] border-t border-beige bg-black p-6'>
      <Button handleClick={() => router.back()} variant='secondary'>
        <span>back</span>
      </Button>
    </div>
  );
};

export const getServerSideProps: GetServerSideProps<{
  session: Session;
}> = async ctx => {
  const session = await getServerAuthSession(ctx);

  if (!session) {
    return {
      redirect: {
        destination: '/',
        permanent: false,
      },
    };
  }

  console.log(session); // this logs session data to the console

  return {
    props: {
      session,
    },
  };
};
```

### console.log result from getServerSideProps

image

### console.log result from NewRecipe page

image

### To reproduce

```tsx
import {
  type InferGetServerSidePropsType,
  type GetServerSideProps,
  type NextPage,
} from 'next';
import { type Session } from 'next-auth';

const Page: NextPage<InferGetServerSidePropsType<typeof getServerSideProps>> = ({ session }) => {
  console.log(session);
  return <div>content</div>;
}

export const getServerSideProps: GetServerSideProps<{ session: Session | null }> = async ctx => {
  const session = await getServerAuthSession(ctx);

  console.log(session);

  return {
    props: {
      session
    }
  }
}

```

### Additional information

I am using the stack for the first time and I have no idea if my approach is right. But the issue gets fixed when I do this in ```_app.tsx```

```tsx
const MyApp: AppType<{ session: Session | null}> =({
  Component,
  pageProps
}) => {
  const { session } = pageProps;

  return ( 
    <SessionProvider session={session}>
      <Component {...pageProps} />
    </SessionProvider>
  )
}
```

instead of this

```tsx
const MyApp: AppType<{ session: Session | null}> =({
  Component,
  pageProps: {session, ...pageProps}
}) => {
  return ( 
    <SessionProvider session={session}>
      <Component {...pageProps} />
    </SessionProvider>
  )
}
```

juliusmarminge:
Closes #1039

## ✅ Checklist

---

## Changelog

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/cli/template/extras/src/pages/_app/with-auth-trpc.tsx
+++ b/cli/template/extras/src/pages/_app/with-auth-trpc.tsx
@@ -8,10 +8,10 @@ import "../styles/globals.css";
 
 const MyApp: AppType<{ session: Session | null }> = ({
   Component,
-  pageProps: { session, ...pageProps },
+  pageProps,
 }) => {
   return (
-    <SessionProvider session={session}>
+    <SessionProvider session={pageProps.session}>
       <Component {...pageProps} />
     </SessionProvider>
   );
--- a/cli/template/extras/src/pages/_app/with-auth.tsx
+++ b/cli/template/extras/src/pages/_app/with-auth.tsx
@@ -6,10 +6,10 @@ import "../styles/globals.css";
 
 const MyApp: AppType<{ session: Session | null }> = ({
   Component,
-  pageProps: { session, ...pageProps },
+  pageProps,
 }) => {
   return (
-    <SessionProvider session={session}>
+    <SessionProvider session={pageProps.session}>
       <Component {...pageProps} />
     </SessionProvider>
   );


c-ehrlich (approved):
Just tested and it seems to work, but why? Seems like it would do the same thing

juliusmarminge:
> Just tested and it seems to work, but why? Seems like it would do the same thing

No clue...

juliusmarminge:
Maybe @balazsorban44 can shed some light on this?

juliusmarminge:
Pulled an issue upstream: https://github.com/nextauthjs/next-auth/issues/6265

balazsorban44:
See my comment here https://github.com/nextauthjs/next-auth/issues/6265#issuecomment-1369789688

juliusmarminge:
> See my comment here [nextauthjs/next-auth#6265 (comment)](https://github.com/nextauthjs/next-auth/issues/6265#issuecomment-1369789688)

Alright, thanks for the response!

Although I am in favour of the "hook-pattern" it still seems a bit cursed that we remove a prop which will break type inference etc (although the inference of gSSP is already quite busted...)

Updated this PR to be a doc addition: https://create-t3-app-git-julius-auth-session-t3-oss.vercel.app/en/usage/next-auth#retrieving-session-server-side

matibox:
> Maybe we could create custom type, that would use InferGetServerSideProps and Omit session value if it exists.

What do you think about this @juliusmarminge? This way it is typesafe. We could export the type from somewhere.

```tsx

import {
  type InferGetServerSidePropsType,
  type GetServerSideProps,
  type NextPage,
} from 'next';
import { Session } from 'next-auth';
import { getServerAuthSession } from '../server/common/get-server-auth-session';

type InferGSSPWithSession<T extends GetServerSideProps> =
  InferGetServerSidePropsType<T> extends { session: Session | null }
    ? Omit<InferGetServerSidePropsType<T>, 'session'>
    : InferGetServerSidePropsType<T>;

const Example: NextPage<InferGSSPWithSession<typeof getServerSideProps>> = ({
  session,
  notSession,
}) => {
  return <div></div>;
};

export const getServerSideProps: GetServerSideProps<{
  session: Session | null;
  notSession: string;
}> = async ctx => {
  const session = await getServerAuthSession(ctx);

  return {
    props: {
      session,
      notSession: 'abcd',
    },
  };
};

export default Example;
```

image

c-ehrlich:
I don't love this solution as it's proprietary so people will not find it in the next-auth docs. This also means that unless they go searching for it in the template, there's no way for them to know that it exists. IMO our goal should be to not stray from conventions established by the packages we use unless absolutely necessary. "How to use NextAuth in T3" should be the same question as "How to use NextAuth in a Next.js app".

(I appreciate that our getServerSession wrapper function kind of goes against this, but we're not really changing conventions there, just making it easier for a project to adapt if there are changes to this unstable API)

juliusmarminge:
Lets roll with this for now i guess?

PR #1064, opened 2023-01-09
feat: add type helper for useQuery options

alonzuman:
Closes #<issue>

## ✅ Checklist

---

## Changelog

Added a ReactQueryOptions type helper ([as seen in the trpc docs](https://trpc.io/docs/infer-types#infer-react-query-options-based-on-your-router)) to the api.ts function to make it easier to create custom hooks

e.g
```typescript
type HelloOptions = ReactQueryOptions['example']['hello']
const useHelloQuery = (options: HelloOptions) => trpc.example.hello.useQuery(undefined, options)

// when using the hook
const helloQuery = useHelloQuery({...}) // options are now typesafe
```

---

💯


juliusmarminge:
Is this used by every app so that it should be included by default?

alonzuman:
> Is this used by every app so that it should be included by default?

I find myself using it whenever creating a custom hook, which is quite often yes (whenever reusing a api.sone.query use)

@juliusmarminge 

Would love to hear more opinions :)

juliusmarminge:
I dont think we need this - I think calling trpc inline is the most common approach at least for the start. If you wanna break up this helper is in the trpc docs which you can look up as you need it.

Issue #1078, opened 2023-01-14
feat: using custom config files


alumen2101:
### Is your feature request related to a problem? Please describe.

everytime i initialize a T3 app, i have to change `tsconfig.json`, `tailwind.config.cjs` to my preference.

### Describe the solution you'd like to see

in CLI, using a flag and directory path that contains all the custom config files, so that it would use those files instead.
``` yarn create t3-app --flag [PATH]```

### Describe alternate solutions

N/A

**Comment by nexxeln** [comment]
I really like this

**Comment by juliusmarminge** [comment]
I like this too, I think it'd be nice to allow something like

```sh
$ p create t3-app@latest --overrides ../path/to/my/tailwind.config.cjs --overrides ../path/to/tsconfig.json
```

and we'd infer the override based on the filename passed in 🤔

nexxeln:
We can even have a config file in `~/.config/ct3a.toml` or something and people would have to set that up just once pointing to their custom config files and then when using the CLI they won't have to pass any flags.

In addition to this we can have a flag like `--no-overrides` where you can get the default template without your custom files.

What do people prefer? Relative path or config file?

juliusmarminge:
> What do people prefer? Relative path or config file?

I dont see why not support both. A parseOverrides function which would find stuff in a deterministic order and pass the parsed config to all the installers, so then you could even mix between a global config and per-project overrides.

Should also outline that we wouldn't take any responsibility whether your app will launch after applying all the overrides 🤣

juliusmarminge:
Adding the discussion label as I think we should give this some thought before jumping into implementation

alumen2101:
> Should also outline that we wouldn't take any responsibility whether your app will launch after applying all the overrides 

is it possible to check passed in files in order to check them if they are valid or not?

juliusmarminge:
> is it possible to check passed in files in order to check them if they are valid or not?

I don't think that's our responsibility since this would be an advanced usage feature

NathanPip:
Hey, still pretty new to the T3 ecosystem but I whipped up a simple implementation of this :-) https://github.com/t3-oss/create-t3-app/pull/1091

NathanPip:
Alright I've given this a little bit of thought. How does this sound?
Like @nexxeln suggested we have a toml file configured like so
```
[override]
override_dir = {path_to_dir_with_override_files}
overrides = [{path_to_file_or_dir}, {path_to_file_or_dir}]

[...]

PR #1091, opened 2023-01-16
feat: override default config files

NathanPip:
simple config override options for cli

regex match tailwind file extension

add lint rule to remove invalid end of line linting error

Closes #<issue>

## ✅ Checklist

---

## Changelog

A simple implementation of config overrides in the cli using --flags.
--twConfig [path] overrides default tailwind config if valid path is provided and file is present
--tsConfig [path] overrides default typescript config if valid path is provided and file is present

## Screenshots

_[Screenshots]_
image

💯


juliusmarminge:
[As mentioned on the original issue](https://github.com/t3-oss/create-t3-app/issues/1078#issuecomment-1382815908), I think it should just be a single flag and then we can match the name of the file.

Ideally we'd support overriding any file too

Issue #1083, opened 2023-01-15
feat: add `Next-ValidEnv` for typesafe environment variables


JacobADevore:
### Is your feature request related to a problem? Please describe.

I created a package called 'Next-ValidEnv', which allows the streamlining of creating typesafe environment variables in Next.js.

https://github.com/JacobADevore/next-validenv

### Describe the solution you'd like to see

Remove the 'env' folder and follow the 'Next-ValidEnv' installation.

### Describe alternate solutions

If create-t3-app wants to rely on something other than a package, I also made it extremely easy to implement this standard manually. (benefit of depending on the package is for monorepos, there is no copy-pasting + logic is abstracted away)

juliusmarminge:
See https://github.com/JacobADevore/next-validenv/pull/2 why this breaks on our setup

c-ehrlich:
I'd also say just in general that I'm hesitant to add external packages to the template, even when they're maintained by contributors - each package is a liability, and we've had enough examples of minor versions breaking stuff in the last few months as is. Doesn't mean we should never do it, just that we should be very careful about it.

juliusmarminge:
Given the way this conversation is going on discord I think we can further discuss any progress in #404

Issue #1069, opened 2023-01-10
feat: Deprecate Manual Declarations of `clientEnv` and `serverEnv`


JacobADevore:
### Is your feature request related to a problem? Please describe.

The current `env` folder structure requires manual declarations of client environment variables which is a duplicate of the client schema. An abstraction should be made to eliminate these manual declarations.

### Describe the solution you'd like to see

On client.mjs automatically create the client environment variables object by looping through the client schema.

`/** @type {{ [key: string]: string | undefined; }} */
let clientEnv = {};`

`Object.keys(clientSchema.shape).forEach(
  (key) => (clientEnv[key] = process.env[key])
);`

### Describe alternate solutions

Leaving as is...

c-ehrlich:
<img width="1624" alt="image" src="https://user-images.githubusercontent.com/8353666/211699693-f250656f-2184-40ab-955d-32b488d94b21.png">

inference is working but initial validation seems broken - i think it's related to this line: `/** @type {{ [key: string]: string | undefined; }} */ let clientEnv = {};`.

i would love it if we can get this to work, would be a huge upgrade!

juliusmarminge:
> inference is working but initial validation seems broken 

You need `NEXT_PUBLIC_BAR: z.coerce.number()` since it will always be a string when read from process.env

JacobADevore:
> Can you access the env through variable indices?
> 
> I.e does
> 
> ```ts
> const key = Math.random() > .5 ? "FOO" : "BAR";
> const value = process.env[key]
> //    ^? ???
> ```
> 
> grab the correct value or will it always be undefined?
> 
> If it works that'd be great 👍

You can, and it does get the correct values.

c-ehrlich:
> > inference is working but initial validation seems broken
> 
> You need `NEXT_PUBLIC_BAR: z.coerce.number()` since it will always be a string when read from process.env

Wow I didn't know that, guess I've never used a number as an env variable before. Will think about a good way to add this to the docs.

<img width="528" alt="image" src="https://user-images.githubusercontent.com/8353666/211715952-898a1b0c-ae26-45f4-a245-83d68c14a8e4.png">

It seems to me like everything works. @JacobADevore please feel free to open a PR for this :) Unless there's some problem I'm missing, this is a really nice dx win.

JacobADevore:
#1071

MartelliEnrico:
[...]

Issue #992, opened 2022-12-19
bug: can't build docker image with NextAuth.js


vojtechmares:
### Provide environment information

  System:
    OS: macOS 13.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 111.53 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 19.3.0 - /opt/homebrew/bin/node
    npm: 9.2.0 - /opt/homebrew/bin/npm
pnpm: 7.18.2

ct3aMetadata.initVersion: 6.11.2

### Describe the bug

Build of my Next app fails due to missing env variables (I am using the SKIP_ENV_VALIDATION=1 env var) and `Error: Failed to collect page data for /sign/in`

### To reproduce

1. clone https://github.com/vojtechmares/training-platform
2. checkout to branch `ci-docker-build`
3. run `docker build -t platform .` or `SKIP_ENV_VALIDATION=1 pnpm run build`

### Additional information

I believe that the `SKIP_ENV_VALIDATION` env var is not being passed to the app.

If I run the `pnpn run build` outside docker without `/.env` (`/` is project root) file, I get the same error. Altho I was unable to turn off the validation even when I commented out the import in `/next.config.mjs`.

I believe that the environment schema is checked when page data collection starts, since I was able to get this error for different pages.

Unfortunately I was unable to get around this issue.

I am using Next.js version `13.0.6`

juliusmarminge:
Your error isn't from the environment variable validation.

Not sure why it doesn't recognize the custom /sign/in route though...

vojtechmares:
From what I googled, it looks like that Next is trying to run `getServerSideProps` during build time (page collection data?), which results in error.

Unfortunately, I was unable to find out why is this happening and how to fix that. I checked the example from Next documentation for authentication (with NextAuth.js), if I am not missing something and I believe I am not.

I find a discussion at [vercel/next.js](https://github.com/vercel/next.js) (see https://github.com/vercel/next.js/discussions/35534), yet I did not find a solution to this issue.

vojtechmares:
I did some more investigations and googling and I think I am confident in the cause. Just no idea how to fix that.

[...]

PR #1092, opened 2023-01-17
feat: single env file

juliusmarminge:
Closes among others #1083 #1069 #1021

cc @iduuck 

## ✅ Checklist

---

## Changelog

Change to use a single env file as entry point. No need to import from server/client - we use a proxy to make sure the right errors are thrown if server vars are accessed on the client.

---

## Screenshots

_[Screenshots]_

For example, if you try accessing `env.DATABASE_URL` on the client:

image


cli/template/extras/src/env/with-auth.mjs
```diff
@@ -0,0 +1,92 @@
...
+/** @type z.infer<merged>
+ *  @ts-ignore - can't type this properly in jsdoc */
+export const env = new Proxy(parsed.data, {
+  get(target, prop) {
+    if (typeof prop !== "string") return undefined;
+    // Throw a descriptive error if a server-side env var is accessed on the client
+    // Otherwise it would just be returning `undefined` and be annoying to debug
+    if (!isServer && !prop.startsWith("NEXT_PUBLIC_"))
+      throw new Error(
+        "❌ Attempted to access serverside environment variable on the client",
```
iduuck, line 86: server-side, or?

cli/template/extras/src/env/with-prisma.mjs
```diff
@@ -0,0 +1,77 @@
...
+
+const isServer = typeof window === "undefined";
+
+const merged = server.merge(client);
+const parsed = isServer
+  ? merged.safeParse(processEnv) // on server we can validate all env vars
+  : client.safeParse(processEnv); // on client we can only validate the ones that are exposed
+
+if (parsed.success === false) {
+  console.error(
```
iduuck, line 55: imo, we should "streamline" this (below we are using "new Error", and here we are using "console.error"..

iduuck:
Beside my two minor comments, I am sure that we could additionally use the type overwrite of `process.env`, if we want to. Though, we don't have the server vs. client type safety this way.

juliusmarminge:
> Beside my two minor comments, I am sure that we could additionally use the type overwrite of `process.env`, if we want to. Though, we don't have the server vs. client type safety this way.

I dont like the process.env typing since you miss stuff like default values for stuff your host provides, transforms/preprocess and other powerful primitives Zod provides

cli/template/extras/src/env/with-prisma.mjs
juliusmarminge: Actually we do throw here, we just log a nice formatted message before.

CleanShot 2023-01-20 at 11 25 33

c-ehrlich:
Which use cases do we need to test to make sure that this works as desired?
- dev
- vercel
- docker (passing in via docker-compose)
- edge?
- other?

juliusmarminge:
> Which use cases do we need to test to make sure that this works as desired?
> 
> * dev
> * vercel
> * docker (passing in via docker-compose)
> * edge?
> * other?

Got one here: https://github.com/juliusmarminge/env-test

Tested on

- vercel edge: https://env-test-lyart.vercel.app/api/edge
- vercel lambda: https://env-test-lyart.vercel.app/api/lambda
- local dev (both edge + lambda)

Go to https://env-test-lyart.vercel.app/server and it will crash since we try accessing server var

iduuck:
I think we need to check for the following use cases:

- middleware (see #1000)
- vercel
- docker
- edge functions
- lambda
- cloudflare workers (?)

cli/template/extras/src/env/with-prisma.mjs
iduuck: I may have missed this, but perhaps also there a nice formatted message, so the redbox from Next.js is going to tell us straight which are wrong, and we do not need to go into console?

iduuck:
One more thought from my side: In the production environment, we might not want to expose which variables are wrong (see your crashing server url):
Screenshot 2023-01-21 at 12 46 20

juliusmarminge:
> One more thought from my side: In the production environment, we might not want to expose which variables are wrong (see your crashing server url):
> Screenshot 2023-01-21 at 12 46 20

Yea I was hoping this would error on build tbf when collecting the page data 🤔

c-ehrlich:
This implementation is still crashing during docker build if you call next-auth stuff in GSSP (see https://github.com/t3-oss/create-t3-app/issues/992#issuecomment-1404043116)

I got a working solution, but there's probably a cleaner way of writing it: https://github.com/c-ehrlich/docker-next-auth/blob/main/src/env.mjs

bai:
Thanks for doing this, a very welcome change ❤️

KaiSpencer:
Webstorm infers `env` as `any` but VSCode correctly infers the type from the merged zod schema. Any ideas on this? :D

iduuck:
What does tsc give you here in the project? Seems like WebStorm isn't using the right jsdoc types.

c-ehrlich:
Added changes from https://github.com/t3-oss/create-t3-app/pull/1092#issuecomment-1404044831
- works in docker
- allows calling next-auth in gssp without failing in build

Tested in local dev and docker

joachimjusth (approved):
Ok for french translation update

www/src/pages/ar/usage/env-variables.md
```diff
@@ -74,7 +74,7 @@ const validated = schema.parse(process.env);
 إذا أردت إستخدام الـ env vars فيمكنك إستيراد` env/client.mjs` و `env/server.mjs` في المكان الذي تريد
```
Sboonny, line 74:
```suggestion
إذا أردت إستخدام الـ env vars فيمكنك إستيراد` env.mjs` واستعمالهم طبيعيا. إذا إستيرادت الملف في Client وحولت استعمال قيم الserver-side، ستنذر بوجود خطأ في run-time.
```

cli/template/base/src/env.mjs
```diff
@@ -31,45 +31,57 @@ const processEnv = {
...
+  const formatErrors = (
+    /** @type {z.ZodFormattedError<Map<string,string>,string>} */
+    errors,
+  ) =>
+    Object.entries(errors)
+      .map(([name, value]) => {
+        if (value && "_errors" in value)
+          return `${name}: ${value._errors.join(", ")}\n`;
+      })
+      .filter(Boolean);
```
juliusmarminge, line 49: i think we can remove this, the default `ZodError.flatten()` looks good

c-ehrlich: image
looks nice

juliusmarminge:
Good on my side. Ship? @c-ehrlich ?

c-ehrlich:
> Good on my side. Ship? @c-ehrlich

yea good for me also

c-ehrlich (approved):

Sn0wye (commented):
LGTM!

PR #1099, opened 2023-01-19
fix(doc): a styling issue of the inline code blocks

fulopkovacs:
Closes #<issue>

## ✅ Checklist

---

## Changelog

Before this fix, the inline code blocks were wrapped in a `div`, so they were placed on new lines. These line breaks made the documentation harder to read, so I applied the `inline-block` TW classes to them (could've been `inline` too, but I thought that maybe the `block`-aspect of the positioning is important, otherwise a `span`-element would've been used).

---

## Screenshots

**Before**
image

**After**
image

_[Screenshots]_

💯


juliusmarminge (approved):
Thanks

PR #1102, opened 2023-01-19
fix: code wrapper not applying styles in firefox

c-ehrlich:
Closes #<issue>

## ✅ Checklist

- [x] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [x] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] I performed a functional test on my final commit

---

## Changelog

the `:has()` selector is not yet enabled in FireFox by default, see: https://caniuse.com/css-has

so instead this PR applies a class to the code wrapper div and styles that. tested in both ltr and rtl.

---

## Screenshots

image

💯

PR #1105, opened 2023-01-20
fix(doc): Fix the broken styling of the code blocks

### LINKED ISSUE #1100: bug: Selector for inline code elements does not work

 System:
    OS: macOS 12.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 3.82 GB / 32.00 GB
    Shell: 3.6.0 - /opt/homebrew/bin/fish
  Binaries:
    Node: 19.2.0 - ~/.local/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 8.19.2 - /opt/homebrew/bin/npm

### Describe the bug

The following lines are supposed to return a list of the `code` nodes that are not the children of a `pre` node (`www/src/components/docs/pageContent.astro`):

```typescript
    const notACodeBlock = ":not(pre) code";
    const codeElements = document.querySelectorAll(notACodeBlock);
```

The problem is that `:not(pre) code` seems to select every `code` element on the page.

### Reproduction repo

https://create.t3.gg/en/usage/prisma

### To reproduce

 Try to run this code snippet in the dev console on the https://create.t3.gg/en/usage/prisma page:
```typescript
Array.from(document.querySelectorAll(':not(pre) code')).filter(e => e.parentNode.parentNode.nodeName === 'PRE')
```
I get 2 `code` elements inside of `pre` elements as a result:
image

### Additional information

I don't know if this problem can be solved with CSS. Personally, I'd do something like this:

```typescript
   const codeElements = Array.from(document.querySelectorAll("code")).filter((codeElement) => codeElement.parentNode?.nodeName !== "PRE");
```

Happy to send a PR if this solution is acceptable! ☺️

fulopkovacs:
Fixes #1100 (which is currently closed, but not entirely fixed: https://github.com/t3-oss/create-t3-app/issues/1100#issuecomment-1398209883)

## ✅ Checklist

---

## Changelog

#1102 accidentally broke the styling of the code blocks (`code` elements inside of `pre`), because of a broken selector.

This part in `www/src/components/docs/pageContent.astro` selects **all** code blocks (including the children of `pre` elements) and applies the `.code-wrapper` class to them, hence the light background and double border:

```typescript
    const notACodeBlock = ":not(pre) code";
    const codeElements = document.querySelectorAll(notACodeBlock);
```

#1102 is a good fix to the issue it addressed btw, the was already broken. ☺️ 

---

## Screenshots

**Before**
image

**After**
image

https://create.t3.gg/en/usage/prisma#seeding-your-database

💯

### DIFF
--- a/www/src/components/docs/pageContent.astro
+++ b/www/src/components/docs/pageContent.astro
@@ -221,12 +221,13 @@ const isRtl = getIsRtlFromLangCode((lang || "en") as KnownLanguageCode);
 </script>
 <script>
   const main = () => {
-    const notACodeBlock = ":not(pre) code";
-    const codeElements = document.querySelectorAll(notACodeBlock);
+    const codeElements = Array.from(document.querySelectorAll("code")).filter(
+      (codeElement) => codeElement.parentNode?.nodeName !== "PRE",
+    );
 
     codeElements.forEach((codeElement) => {
       const codeWrapper = document.createElement("div");
-      codeWrapper.classList.add("inline-block", "code-wrapper");
+      codeWrapper.classList.add("code-wrapper");
       codeElement.parentNode?.insertBefore(codeWrapper, codeElement);
 
       codeWrapper.appendChild(codeElement);


c-ehrlich (approved):
Nice catch and thanks for the fix. Still surprised by how difficult it is to support two different writing directions 😅

PR #1109, opened 2023-01-21
refactor: replaced VERCEL with VERCEL_URL in serverSchema for consistency

### LINKED ISSUE #1108: refactor: change process.env.VERCEL to process.env.VERCEL_URL in serverSchema

Zod validation object for serverSchema makes use of both **VERCEL** and **VERCEL_URL** environment variables for checking whether the user is deploying their app on **vercel** to preprocess the default value for **NEXTAUTH_URL** environment variable. Two files named **with-auth-prisma.mjs** and **with-auth.mjs** make use of this. Path to the files is this:-  _cli/template/extras/src/env/schema/_

This is what it currently looks like:-

current

The environment variable **VERCEL** should be replaced by **VERCEL_URL** so that the **vercel** environment check works only against **VERCEL_URL** thus making serverSchema look more consistent.

This is what it should look like:-

new

Nope

### Additional information

I have the changes ready for pull request just need your confirmation.

xoldyckk:
Closes #1108

## ✅ Checklist

- [✅ ] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [ ✅] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)

---

## Changelog

Replaced all instances of environment variable named **VERCEL** to **VERCEL_URL** used in **zod** validation object for **serverSchema**.

---

## Screenshots

Before:-

old

After:-

new

💯

### DIFF
--- a/cli/template/extras/src/env/schema/with-auth-prisma.mjs
+++ b/cli/template/extras/src/env/schema/with-auth-prisma.mjs
@@ -17,7 +17,7 @@ export const serverSchema = z.object({
     // Since NextAuth.js automatically uses the VERCEL_URL if present.
     (str) => process.env.VERCEL_URL ?? str,
     // VERCEL_URL doesn't include `https` so it cant be validated as a URL
-    process.env.VERCEL ? z.string() : z.string().url(),
+    process.env.VERCEL_URL ? z.string() : z.string().url(),
   ),
   DISCORD_CLIENT_ID: z.string(),
   DISCORD_CLIENT_SECRET: z.string(),
--- a/cli/template/extras/src/env/schema/with-auth.mjs
+++ b/cli/template/extras/src/env/schema/with-auth.mjs
@@ -16,7 +16,7 @@ export const serverSchema = z.object({
     // Since NextAuth.js automatically uses the VERCEL_URL if present.
     (str) => process.env.VERCEL_URL ?? str,
     // VERCEL_URL doesn't include `https` so it cant be validated as a URL
-    process.env.VERCEL ? z.string() : z.string().url(),
+    process.env.VERCEL_URL ? z.string() : z.string().url(),
   ),
   DISCORD_CLIENT_ID: z.string(),
   DISCORD_CLIENT_SECRET: z.string(),


c-ehrlich (commented):
`VERCEL` is a Vercel System Environment Variable used to indicate that the app is running on Vercel, see: https://vercel.com/docs/concepts/projects/environment-variables#system-environment-variables

I suppose `VERCEL_URL` would also be nullish if not running on Vercel so it would do the same thing, but I think I prefer using the `VERCEL` variable which explicitly exists for this purpose.

Anyone else have an opinion on this?

juliusmarminge:
> `VERCEL` is a Vercel System Environment Variable used to indicate that the app is running on Vercel, see: https://vercel.com/docs/concepts/projects/environment-variables#system-environment-variables
> 
> 
> 
> I suppose `VERCEL_URL` would also be nullish if not running on Vercel so it would do the same thing, but I think I prefer using the `VERCEL` variable which explicitly exists for this purpose.
> 
> 
> 
> Anyone else have an opinion on this?

Yea its like i said [here](https://github.com/t3-oss/create-t3-app/issues/1108#issuecomment-1398959396) that it wont make any difference.

xoldyckk:
@c-ehrlich @juliusmarminge I know that this is just a refactor which won't at all affect how the tool currently works. But as I mentioned it's done to be consistent with what you're checking against. The comments you've written in the file don't mention anything about the `VERCEL` environment variable at all. Thus, making it an extra step for the person reading the code to make sense of why two different environment variables are being used and what the `VERCEL` environment variable means in that context.

Anyway it's up to you to decide. I just thought this was a good refactor.

juliusmarminge:
I think the better thing to do here is to modify the schema and do some transforming so that it always starts with `http(s?)://` no matter if it's from vercel or not...

c-ehrlich:
Closing this. I could see either changing docs/comments or doing something similar to what @juliusmarminge is suggesting, but the changes that are being suggested here go against the semantic usage of Vercel's environment variables.

PR #1118, opened 2023-01-24
docs: Fix path

radurentea:
## ✅ Checklist

---

## Changelog

- Keeping paths relative to root folder
---

💯

### DIFF
--- a/www/src/pages/en/usage/prisma.md
+++ b/www/src/pages/en/usage/prisma.md
@@ -9,7 +9,7 @@ Prisma is an ORM for TypeScript, that allows you to define your database schema 
 
 ## Prisma Client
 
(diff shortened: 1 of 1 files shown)


c-ehrlich:
We use paths without src in a lot of places, for example https://github.com/t3-oss/create-t3-app/blob/next/www/src/pages/en/usage/trpc.md

Whatever we decide should be consistent across the whole docs, so I guess the question is so we want the explicit src nowhere or everywhere?

juliusmarminge (approved):

juliusmarminge:
> Whatever we decide should be consistent across the whole docs, so I guess the question is so we want the explicit src nowhere or everywhere?

Problem for a later day ;)

PR #1123, opened 2023-01-26
fix: improve readability of gh stars counter

Azzeccagarbugli:
## ✅ Checklist

---

## Changelog

The current PR aims to use punctuation and format the GitHub stars counter in order to reduce ambiguity and improve readability and comprehension as reported in the [_Web Accessibility Initiative (WAI)_](https://www.w3.org/WAI/)

---

## Screenshots

| **Before** |Screenshot 2023-01-26 at 01 23 34|
|---|:---:|
| **After** |Screenshot 2023-01-26 at 01 23 23|

💯

### DIFF
--- a/www/src/components/navigation/githubIcon.astro
+++ b/www/src/components/navigation/githubIcon.astro
@@ -27,7 +27,7 @@ const isRtl = getIsRtlFromUrl(pathname);
   <div
     id="github-star"
     class={clsx(
-      "relative ltr:mr-[.33em] rtl:ml-[.33em] hidden md:flex items-center bg-clip-padding rounded-lg no-underline border-t3-purple-200/50 group-hover:border-t3-purple-200/75 bg-t3-purple-200/50 dark:bg-t3-purple-200/10 border dark:border-t3-purple-200/20 dark:group-hover:bg-t3-purple-200/10 group-hover:bg-t3-purple-200/75 dark:group-hover:border-t3-purple-200/50 transition-colors duration-300 p-1 text-slate-800 dark:text-slate-100",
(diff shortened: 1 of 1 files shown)


www/src/components/navigation/githubIcon.astro
```diff
@@ -47,7 +47,7 @@ const isRtl = getIsRtlFromUrl(pathname);
         d="M316.9 18C311.6 7 300.4 0 288.1 0s-23.4 7-28.8 18L195 150.3 51.4 171.5c-12 1.8-22 10.2-25.7 21.7s-.7 24.2 7.9 32.7L137.8 329 113.2 474.7c-2 12 3 24.2 12.9 31.3s23 8 33.8 2.3l128.3-68.5 128.3 68.5c10.8 5.7 23.9 4.9 33.8-2.3s14.9-19.3 12.9-31.3L438.5 329 542.7 225.9c8.6-8.5 11.7-21.2 7.9-32.7s-13.7-19.9-25.7-21.7L381.2 150.3 316.9 18z"
       ></path>
     </svg>
-    {githubStars}
+    {Intl.NumberFormat().format(githubStars)}
```
juliusmarminge, line 50: Do we want it localized or should we hardcode locale to en-US so that there would always be the `XX,YYY` format? I personally like the `,` even though Swedish locale doesnt have it

Azzeccagarbugli: I like the `,` separator as well but I really didn't want to _force_ the `en-US` behavior so I just went with the localized implementation. 

The article that I linked above about the [Web Accessibility Initiative (WAI)](https://www.w3.org/WAI/) says that each culture adopts a different way to read and write numbers, so I would keep the `Intl` without _imposing_ any custom separator. 

Furthermore, out of curiosity, I found out that the tRPC homepage adopts exactly the [same implementation](https://github.com/trpc/trpc/blob/4f8db604a55578a3e496fa2a44f86c5c1b2a305b/www/og-image/pages/api/landing.tsx#L18)

Let me know what you think 😄

juliusmarminge (approved):

www/src/components/navigation/githubIcon.astro
juliusmarminge: Probably for the better. I got my computer set to US anyways so I'll still see the `,` 😎

PR #1129, opened 2023-01-27
chore: cleanup auth.ts

chernodub:
From time to time, I see eslint yelling because of the redundant await here whenever I start a new project. Not sure if it follows your preferences, but I'd be happy to see the neater version of this function in your codebase. Feel free to omit the changes if it does not follow your codestyle. Thanks a lot for the amazing work you do!

## ✅ Checklist

---

## Changelog

- Simplified/improved the formatting in `auth.ts`

ps: sorry for not opening the issue, decided it would be reasonable to just leave a PR right away

### DIFF
--- a/cli/template/extras/src/server/auth.ts
+++ b/cli/template/extras/src/server/auth.ts
@@ -13,9 +13,7 @@ import { authOptions } from "../pages/api/auth/[...nextauth]";
  * @see https://next-auth.js.org/configuration/nextjs
  */
 
-export const getServerAuthSession = async (ctx: {
+export const getServerAuthSession = (ctx: {
   req: GetServerSidePropsContext["req"];
   res: GetServerSidePropsContext["res"];
-}) => {
-  return await unstable_getServerSession(ctx.req, ctx.res, authOptions);
-};
+}) => unstable_getServerSession(ctx.req, ctx.res, authOptions);


juliusmarminge:
unstable_getServerSession is async though so it should be awaited?

CleanShot 2023-01-28 at 11 39 57@2x

chernodub:
Thanks, @juliusmarminge. But, `async`/`await` is just syntactic sugar over `Promise`s, we use it whenever we need to make the code sync-like. Here we don't have any other statements besides the return. So, in this scenario, where we immediately return the result, we don't really need the `await`, right?

Basically, what I'm saying is:

```ts
async function doAsyncStuff() {
  return await Promise.resolve();
}

// these are equal, 
// we don't need to wait for the `Promise.resolve` to end, since we return it right away

function doAsyncStuff() {
  return Promise.resolve();
}
```

Am I not seeing something?

juliusmarminge:
> Thanks, @juliusmarminge. But, `async`/`await` is just syntactic sugar over `Promise`s, we use it whenever we need to make the code sync-like. Here we don't have any other statements besides the return. So, in this scenario, where we immediately return the result, we don't really need the `await`, right?
> 
> Basically, what I'm saying is:
> 
> ```ts
> async function doAsyncStuff() {
>   return await Promise.resolve();
> }
> 
> // these are equal, 
> // we don't need to wait for the `Promise.resolve` to end, since we return it right away
> 
> function doAsyncStuff() {
>   return Promise.resolve();
> }
> ```
> 
> Am I not seeing something?

No you're right - and the result of calling the function will still be await-able. But I'm just not seeing why this would result in a lint error?

juliusmarminge:
Superseded by #1136

Did some major refactoring of the whole auth.ts file. Included your change in there too

chernodub:
Thanks a lot, @juliusmarminge. By the way here's the rule I was talking about - https://eslint.org/docs/latest/rules/no-return-await

PR #1132, opened 2023-01-28
fix: correct language for gh stars count

chalop:
## ✅ Checklist

---

## Changelog

Shows the language's equivalent formatting for the GitHub stars count on localized pages -- enhances the work from [#1123](https://github.com/t3-oss/create-t3-app/pull/1123). 

The language detection code is implemented from [previous work](https://github.com/t3-oss/create-t3-app/blob/next/www/src/languages.ts#L6).

---

### before
ar-before

### after
ar-after


c-ehrlich:
Is this better? @a4addel @Sboonny

a4addel:
Yes, but with a small note in Arabic we Use __**","**__  as a seperator not **"،"**

Sboonny:
As a programmer, I advise against it, because you will be asked to change the Markdown format from `.1` and `.2` to `.۱` and `.٢`, and `persion` and `urdu` have different number system.

As a translator, I recommend using the word that describe numbers like how the newspaper articles' translators do.

aka `42,069` is 42 thousands, so it will be `٤٢ الف`. 

----

Persion, urdu and arabic translators in freeCodeCamp didn't care about this, so I advise that you shouldn't as well

chalop:
We could use the compact notation instead, this implements the feedback from @Sboonny. 
As a bonus, since the star count isn't realtime there is no need to be super specific with the number for the star count, so the compact version works for that case as well.

Examples below: 

### Arabic
ar

### English
en

### French
fr

### Norwegian
no

### Polish
pl

### Portuguese
pt

### Russian
ru

### Chinese
zh-hans

www/src/components/navigation/githubIcon.astro
```diff
@@ -47,7 +47,11 @@ const isRtl = getIsRtlFromUrl(pathname);
         d="M316.9 18C311.6 7 300.4 0 288.1 0s-23.4 7-28.8 18L195 150.3 51.4 171.5c-12 1.8-22 10.2-25.7 21.7s-.7 24.2 7.9 32.7L137.8 329 113.2 474.7c-2 12 3 24.2 12.9 31.3s23 8 33.8 2.3l128.3-68.5 128.3 68.5c10.8 5.7 23.9 4.9 33.8-2.3s14.9-19.3 12.9-31.3L438.5 329 542.7 225.9c8.6-8.5 11.7-21.2 7.9-32.7s-13.7-19.9-25.7-21.7L381.2 150.3 316.9 18z"
       ></path>
     </svg>
-    {Intl.NumberFormat().format(githubStars)}
+    {
+      Intl.NumberFormat(getLanguageFromURL(pathname), {
+        notation: "compact",
+      }).format(githubStars)
+    }
```
juliusmarminge, line 54: I think it's better to do this in the javascript part at the top rather than in the markdown?

chalop: I would agree with this, but to keep the md code clean. In terms of outcome, it would be the same right?

chalop: Something like: 
```js
const formattedStars = new Intl.NumberFormat(getLanguageFromURL(pathname), {
  notation: "compact",
}).format(githubStars);
```
and just call formattedStars on the markdown

chalop: Judging by other references on the repo, seems like the convention is to keep js/md separate as much a possible.

I pushed the changes 👍

c-ehrlich (approved):
looks and works great :)

PR #1134, opened 2023-01-28
feat: added package installation progress

### LINKED ISSUE #1049: feat: Install Progress

When using the installer provided by the tool, there is no way to show what is is busy doing. This leads me to not understanding how long it still needs, how big it is or what it is busy doing.

Just a simple progress bar with the packages install progress or even just the output from the installer, pnpm, npm, yarn ect.

Pass in a flag at the start that will display it at the bottom or open up a terminal with the command already typed in

### Additional information

_No response_

PiotrekPKP:
Closes #1049

## ✅ Checklist

---

## Changelog

- added package installation progress to the cli working for all 3 supported package managers
- `npm` - output is inherited from `stderr`
- `pnpm` and `yarn` - output is redirected from `stdout`

---

### `npm`
image

### `pnpm`
image

### `yarn`
image

💯


juliusmarminge:
Can we make the pnpm output prettier and not pushed so far to the right?

CleanShot 2023-01-28 at 11 52 39

cli/src/helpers/installDependencies.ts
```diff
@@ -1,27 +1,76 @@
...
+    case "pnpm":
+      const pnpmSpinner = ora("Running pnpm install...").start();
+      const pnpmSubprocess = execa(pkgManager, ["install"], {
+        cwd: projectDir,
+        stdout: "pipe",
+      });
+
+      await new Promise<void>((res, rej) => {
+        pnpmSubprocess.stdout?.on("data", (data: Buffer) => {
+          pnpmSpinner.text = data.toString();
```
juliusmarminge, line 38: Something like that but maybe bit smarter matching idk?

```suggestion
pnpmSpinner.text = "Progress" + data.toString().split("Progress")[1];
```

PiotrekPKP:
> Can we make the pnpm output prettier and not pushed so far to the right?
> 
> CleanShot 2023-01-28 at 11 52 39

`pnpm` uses this kind of output by default:
image

However, I think we can try to either:
- parse the progress bar manually and show it somewhat differently
- cut the string to the `|` sign and only show the text with progress info

What do you think is the better option? Of course the second one is easier haha

cli/src/helpers/installDependencies.ts
PiotrekPKP: Yeah, that's what I've just commented haha

juliusmarminge: Astro's was clean and only reused a single line too, how does that work here?

PiotrekPKP: Tbf I'm not sure how their code works, I did the same thing they did. In Astro's codebase they also just use the data from the stdout :thinking:

juliusmarminge: > Tbf I'm not sure how their code works, I did the same thing they did. In Astro's codebase they also just use the data from the stdout :thinking: 

Maybe split by `\n`, take the last snd then spöit by the `|`?

PiotrekPKP:
It looks like that now:
image

cli/src/helpers/installDependencies.ts
PiotrekPKP: I used the split by `|` for now, looks as it should I think

cli/src/helpers/installDependencies.ts
```diff
@@ -1,27 +1,78 @@
...
+    case "pnpm":
+      const pnpmSpinner = ora("Running pnpm install...").start();
+      const pnpmSubprocess = execa(pkgManager, ["install"], {
+        cwd: projectDir,
+        stdout: "pipe",
+      });
+
+      await new Promise<void>((res, rej) => {
+        pnpmSubprocess.stdout?.on("data", (data: Buffer) => {
+          if (data.toString().includes("Progress")) {
```
juliusmarminge, line 38:
```suggestion
          const text = data.toString();
          if (text.includes("Progress") && text.includes("|")) {
```

Seems like not every pnpm version uses the `|` so we should check for that too:

CleanShot 2023-01-28 at 12 19 32

juliusmarminge: https://user-images.githubusercontent.com/51714798/215263845-fd219e4a-1988-44ab-8298-80fba602d0ab.mp4

I for example doesn't get any of the progress output here

PiotrekPKP: I'm still checking for the progress text for older versions, no reason to show lines like `virtual store is at:` imo. Thanks for the info tho, I've updated the pr :+1:

juliusmarminge: CleanShot 2023-01-28 at 12 26 10

nice

juliusmarminge (approved):
Awesome work!

PR #1136, opened 2023-01-28
refactor: colocate auth stuff and bump some stuff

juliusmarminge:
Closes #<issue>

## ✅ Checklist

---

## Changelog

Colocates all the auth setup in `~/server/auth.ts` instead of importing the options from `~/pages/api...`.

Also bumps some versions to remove the `unstable_` prefix from the server session getter. Still wrapping it cause it's annoying having to import options & getter in every place you use it

Also took the liberty to update the prisma initialization to the new way that they recommend.

---

## Screenshots

Auth:
image

Prisma:
CleanShot 2023-01-28 at 17 35 05

💯


cli/template/extras/src/server/auth/base.ts
```diff
@@ -0,0 +1,73 @@
...
+  //   // role: UserRole;
+  // }
+}
+
+/**
+ * Options for NextAuth.js used to configure
+ * adapters, providers, callbacks, etc.
+ * @see https://next-auth.js.org/configuration/options
+ **/
+export const authOptions: NextAuthOptions = {
```
c-ehrlich, line 36: I like having it here a lot :)

c-ehrlich (approved):
I like these changes a lot, but I think @nexxeln should read as well before we merge :)

nexxeln (approved):

PR #1164, opened 2023-02-04
chore: update linting config

Zamiell:
## ✅ Checklist

---

## Changelog

I standardized JSDoc comments using a lint rule. When doing this:
- I took the liberty to rewrite some of the comments in more proper English.
- I removed some unnecessary hyphens. I don't think using hyphens is idiomatic in the ecosystem, e.g. the React codebase doesn't use them.

In the next PR after this is approved, I will extend the new rules to apply to the templates.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,6 +1,10 @@
 module.exports = {
   parser: "@typescript-eslint/parser", // Specifies the ESLint parser
-  plugins: ["turbo"],
+  plugins: [
+    // Provides extra useful rules:
+    // https://github.com/IsaacScript/isaacscript/tree/main/packages/eslint-plugin-isaacscript
+    "isaacscript",
+  ],
   extends: [
     "plugin:@typescript-eslint/recommended",
     "plugin:prettier/recommended",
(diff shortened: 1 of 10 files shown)


.eslintrc.cjs
```diff
@@ -1,6 +1,10 @@
 module.exports = {
   parser: "@typescript-eslint/parser", // Specifies the ESLint parser
-  plugins: ["turbo"],
```
Zamiell, line 3: The fact that this was here was a bug, as the turbo rule wasn't even being used by the codebase.

.vscode/settings.json
```diff
@@ -4,11 +4,15 @@
   },
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
+  "editor.rulers": [100],
+  "editor.tabSize": 2,
```
Zamiell, line 8: Prettier enforces a tab size of 2, and this codebase uses Prettier as the standard, so this needs to be reflected in the VSCode config as well so that it matches the project.

.vscode/settings.json
```diff
@@ -4,11 +4,15 @@
   },
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
+  "editor.rulers": [100],
```
Zamiell, line 7: I chose a value of 100 to match the existing TypeScript ecosystem. (This value needs to reflect the arbitrary value chosen for the project.)

juliusmarminge (approved):

PR #1170, opened 2023-02-06
chore: update db.ts to make the use of .env consistent

nicokunz:
## ✅ Checklist

- [x ] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [x ] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [] I performed a functional test on my final commit

---

## Changelog

I made a small change to make the use of environment variables in the db.ts consistent within the file itself. In the previous version it was using both the imported env variable and a direct access to process.env. This should not change any behaviour.

### DIFF
--- a/cli/template/extras/src/server/db.ts
+++ b/cli/template/extras/src/server/db.ts
@@ -11,4 +11,4 @@ export const prisma =
       env.NODE_ENV === "development" ? ["query", "error", "warn"] : ["error"],
   });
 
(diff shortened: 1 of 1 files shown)


juliusmarminge (approved):
Thanks

PR #1181, opened 2023-02-07
chore: add lint rule + fix astro lint

Zamiell:
Closes #<issue>

## ✅ Checklist

---

## Changelog

- I made it so that every file in the repository was linted, instead of a subset. This is a more correct/robust setup for ESLint.
- Most of the changes are just noise from the Prettier-style formatting changes from the new ESLint rule, although they should be manually reviewed as I did some touch-ups.
- Linting in Astro files did not work inside of VSCode because this repo was missing the config that is supposed to be there according to [the Astro plugin documentation](https://github.com/ota-meshi/eslint-plugin-astro#computer-editor-integrations). I fixed this.

Further work needs to be done, I think a lot of the current linting config can be cleaned up more than it currently is.


.vscode/settings.json
```diff
@@ -7,8 +7,24 @@
   "editor.rulers": [100],
   "editor.tabSize": 2,
 
-  "eslint.rules.customizations": [{ "rule": "*", "severity": "warn" }],
+  // Inform VSCode that it should use the version of TypeScript that is installed under this project
```
c-ehrlich, line 10: is it ok to have comments in json? it lights up red in the github pr review ui

c-ehrlich (commented):
What is the intended outcome of these changes?

I checked out the pr and installed packages, then broke an [eslint rule that's included in the recommendations](https://ota-meshi.github.io/eslint-plugin-astro/rules/no-conflict-set-directives/), but did not get an error in VSCode.

image

.vscode/settings.json
Zamiell: the file in question is JSONC, not JSON.
i'll add something to the PR that fixes the visual bug, one moment

Zamiell: see https://github.com/t3-oss/create-t3-app/pull/1181/commits/aa558135fe6d87e93d24917ce50fe9903f659d36

Zamiell: > I checked out the pr and installed packages, then broke an [eslint rule that's included in the recommendations](https://ota-meshi.github.io/eslint-plugin-astro/rules/no-conflict-set-directives/), but did not get an error in VSCode.

I get the expected error just fine:

image

try recloning the repo from scratch maybe and ensure that you have all of the proper extensions installed?

Zamiell: another thing to try is for any personal settings.json settings that conflict with ones in the workspace settings.json

PR #1183, opened 2023-02-07
fix: "create t3-app" -> "create-t3-app"

kellenproctor:
Yeah, the code blocks on the website were missing the dash between "create" and "t3", so when you copied the command and ran it in your CLI, nothing happened. So I just added the dashes everywhere I could find.

## ✅ Checklist

- ✅ I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- ✅ The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- ✅ I performed a functional test on my final commit

---

## Changelog

changed "create t3-app" to "create-t3-app"

---

## Screenshots

None

💯

### DIFF
--- a/.github/workflows/prerelease-comment.yml
+++ b/.github/workflows/prerelease-comment.yml
@@ -48,7 +48,7 @@ jobs:
             A new create-t3-app prerelease is available for testing. You can install this latest build in your project with:
 
             ```sh
-            pnpm create t3-app@${{ env.BETA_PACKAGE_VERSION }}
+            pnpm create-t3-app@${{ env.BETA_PACKAGE_VERSION }}
             ```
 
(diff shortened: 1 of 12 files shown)


bai:
Hmmm, AFAIK this is following the convention and the existing approach works fine for me:

Screenshot 2023-02-07 at 06 58 20

I'm unsure this won't break things?

kellenproctor:
You're right, that was dumb of me. For some reason it fails using npx, but works fine for npm, yarn, and pnpm. I should have run those first.

I'll close this. Apologies!

bai:
No need to apologize at all, thanks for raising this ❤️💖

PR #1189, opened 2023-02-09
fix: remove ts-ignores on env.mjs

p13lgst:
## ✅ Checklist

---

## Changelog

- Fix JSDoc types on `env.mjs` removing ts-ignores

💯


cli/template/base/src/env.mjs
```diff
@@ -32,16 +32,21 @@ const processEnv = {
...
   const isServer = typeof window === "undefined";
 
-  const parsed = isServer
-    ? merged.safeParse(processEnv) // on server we can validate all env vars
-    : client.safeParse(processEnv); // on client we can only validate the ones that are exposed
+  const parsed = /** @type {MergedSafeParseReturn} */ (
+    isServer
+      ? merged.safeParse(processEnv) // on server we can validate all env vars
+      : client.safeParse(processEnv) // on client we can only validate the ones that are exposed
+  );
```
juliusmarminge, line 49: do you need the type in the middle like that? doesn't this work?

```suggestion
  /** @type {MergedSafeParseReturn} */
  const parsed = isServer
    ? merged.safeParse(processEnv) // on server we can validate all env vars
    : client.safeParse(processEnv); // on client we can only validate the ones that are exposed
```

juliusmarminge (commented):
Interesting. I like it!

We need to do it for all the other files too though

cli/template/base/src/env.mjs
p13lgst: It doesn't work like that, we need this syntax for the type casting: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#casts

juliusmarminge: I see

p13lgst: Screenshot from 2023-02-09 17-43-15

The type casting works kinda like the `as` keyword in TS

bai:
This is badass ✨ Thanks for contributing!

c-ehrlich (approved):
looks good to me how it is now, but would like a 2nd approval from @juliusmarminge before merging

juliusmarminge (approved):

PR #1217, opened 2023-02-18
fix: make type-imports autofixable & general project cleanup

### LINKED ISSUE #1163: bug: changing eslint config doesn't bust the linting cache
### To reproduce

- Modify the ESLint config.
- Type `pnpm run lint`.
- Observe that the cache is used.

### LINKED ISSUE #1173: bug: erroneous line in tsconfig.json
This line should be removed from [the base monorepo tsconfig](https://github.com/t3-oss/create-t3-app/blob/next/tsconfig.json):

```
  "include": ["cli", "www"],
```

From what I can tell, this line does nothing because It gets overwritten by the child tsconfigs that inherit it. Thus, having it present at all is confusing and should be considered a bug.

From what I know about TypeScript monorepos, I am pretty sure that the base monorepo tsconfig file should never have an `include` statement in it.

### LINKED ISSUE #1182: bug: prettier/prettier rule is used
Is there a particular reason that the `prettier/prettier` ESLint rule is used?
This is not recommended by Prettier and is an anti-pattern AFAIK: https://prettier.io/docs/en/integrating-with-linters.html

If there isn't a particular reason, I can submit a PR that fixes this.

juliusmarminge:
Closes #1173
Closes #1163 (disabled cache job - not optimal but not the end of the world, can make a separate package if we really want it cached)
Closes #1182
Closes #1181

## ✅ Checklist

---

## Changelog

Ended up being a bit of everything pr 😮‍💨 

- fixes some linting, formatting, stylistic changes

https://user-images.githubusercontent.com/51714798/219900129-f9a99702-4f4e-4db3-b95d-7fe48d89f292.mp4

---

## Screenshots

_[Screenshots]_

💯


c-ehrlich:
we currently use `import type { foo } from "bar"` in some places and `import { type foo } from "bar"` in others

is it worth making that consistent?

juliusmarminge:
> we currently use `import type { foo } from "bar"` in some places and `import { type foo } from "bar"` in others
> 
> is it worth making that consistent?

probably - this lint rule is gonna prefer the inline one but it doesn't complain about the other way. maybe there is a way to do that 🤔

juliusmarminge:
> we currently use `import type { foo } from "bar"` in some places and `import { type foo } from "bar"` in others
> 
> is it worth making that consistent?

Done using this rule: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/consistent-type-specifier-style.md

package.json
```diff
@@ -36,8 +36,8 @@
     "dev:www": "turbo --filter=www dev",
     "clean": "turbo clean && rm -rf node_modules",
     "lint": "turbo lint",
-    "format": "turbo format && prettier --write *.{cjs,md,json}",
-    "format:check": "turbo format:check && prettier --check *.{cjs,md,json}",
+    "format": "turbo format && prettier --write *.{cjs,md,json} --ignore-unknown --no-error-on-unmatched-pattern",
+    "format:check": "turbo format:check && prettier --check *.{cjs,md,json} --ignore-unknown --no-error-on-unmatched-pattern",
```
juliusmarminge, line 40: Makes it not fail on the symlink:

CleanShot 2023-02-19 at 09 50 25

.github/workflows/translations.yml
```diff
@@ -48,7 +48,7 @@ jobs:
             const formattedString = changedLanguages
               .map(
                 (lang) =>
-                  `${lang.toUpperCase()}: ${CODEOWNERS[lang].map((user) => `@${user}`).join(", ")}`,
+                  `${lang.toUpperCase()}: ${CODEOWNERS[lang]?.map((user) => `@${user}`).join(", ")}`,
```
juliusmarminge, line 51: so that languages without codeowners (currently spanish) doesn't error here

Issue #1079, opened 2023-01-14
feat: add two-way server <=> client binding to the demo app


jdevries3133:
### Is your feature request related to a problem? Please describe.

This two-way binding pattern is very common. I am new to react-query, this project is actually what is bringing me into this ecosystem. As I'm trying to familiarize myself, this was one of the first thing I explored, and I think that the t3 boilerplate would be stronger if it included some kind of example like this.

### Describe the solution you'd like to see

I have a diff that you can see here which demonstrates exactly what I'm talking about: https://github.com/jdevries3133/t3_two_way_binding/pull/1

The mutation invalidates the query so that the page "live updates" when the button is clicked.

I do think it's strange that the boilerplate includes the "example" resource, complete with db schema and some backend operations, but it just doesn't quite expose it in the client!

### Describe alternate solutions

I think an ideal solution would actually have a bit of delay on the backend, and maybe even a random (say 30%) chance of an error to make a cool demo of the full feature-set that we're offering here, but these are bells and whistles that I don't necessarily need to see; just a thought.

I also think that we might just want to keep the demo app lean and get people to RTFM, and I'd accept that answer too!

c-ehrlich:
Thanks for the issue and the sample diff. 

I agree that the `example` resource feels a bit weird currently depending on what is scaffolded. It's intended to exists as basic Prisma documentation (but is kind of worthless when an app is initialized with next-auth because that comes with its own schema). 

Reasons to do it:
- Gives the `example` resource more of a purpose
- I agree that it works great in terms of documentation/education
- It would very clearly be a demo of tRPC functionality - I suspect some people don't even realize what the "Hello from tRPC" text is doing, an interactive demo would be more clear about this

Reasons not to do it:
- More code in the starter template (we've been working hard to keep it as slim as possible, especially index.tsx). This is undesireable for two reasons:
    1. I suspect most people just rip out the example router and index.tsx after scaffolding
    2. The more code we include the less likely people are to read any of it, so adding more stuff actually decreases the educational value of the existing stuff.
- It's primarily about showing how tRPC works, but it will only exist in apps that include Prisma.
    - could just mutate an array in apps without Prisma, but that feels like it actively hurts the educational nature as our main deploy target is serverless where something like that doesn't work
[...]

PR #1225, opened 2023-02-20
feat: crud example in trpc+prisma templates

c-ehrlich:
Closes https://github.com/t3-oss/create-t3-app/issues/1079

## ✅ Checklist

---

## Changelog

- adds a crud example (create + getAll) to projects that are initiated with trpc and prisma (plus any other combination of packages)
- some minor index refactoring to make it easier to edit multiple of them - everything that's specific to certain packages now lives in separate components
- we now have 10 index.tsx files. i'm so sorry

original version of this was built by @jdevries3133 - i ended up writing this from scratch because working from current next was easier, but would be good to include as a collaborator on the pr

---

## Screenshots

image

💯


c-ehrlich:
Played around with this, ends up requiring 12 (?) index.tsx files and similar situation for a few other files. I don't think it's worth the additional complexity for now. If someone has a better idea and wants to give this another try, feel free :)

PR #1234, opened 2023-02-23
feat: add ts-reset to create-t3-app monorepo (not t3-app)

drewradcliff:
## ✅ Checklist

---

## Changelog

Add ts-reset to make typescript better
https://github.com/total-typescript/ts-reset

💯


bai:
My 2c: I feel like adding this to ct3a is an overkill: adding a dependency vs reward - removing one `as unknown`.

PiotrekPKP:
> My 2c: I feel like adding this to ct3a is an overkill: adding a dependency vs reward - removing one `as unknown`.

It's not removing just the one unknown for this one use case - it's changing the entire json.parse type and other types improving the ts experience (https://github.com/total-typescript/ts-reset#rules)

juliusmarminge (commented):
This isn't adding a reset on the generated app so I think I'm for it.

One thing though which could shoot us in the foot a bit - since we'll get the "resetted" types when developing we might make mistakes when it comes to the generated app. 

Say for example the template file would use a `Array.prototype.filter(Boolean)`, we would see the typed narrowed but for the end user the type wouldn't be narrowed since the reset wouldn't apply once the CLI has generated their app.

I'd say the "risk" of this happening is very low, and will be catched in CI but might still cause unexpected behaviors for contributors seeing one thing in their editor and another thing in CI...

c-ehrlich:
I'm ok with it in the CLI.

In the template I'm strongly against it, at least for now - it's only a few days old and at version 0.x

PR #1237, opened 2023-02-24
fix: Import .env file for sensitive data

mmuller88:
Closes #<issue>

## ✅ Checklist

---

## Changelog

---

## Screenshots

_[Screenshots]_

💯


c-ehrlich:
Do we want this?

In my experience passing env vars to docker compose the way we are doing in the existing docs is more common in production apps than juggling env files. And including both in the docker-compose certainly doesn't make sense. Either way this is only intended to be a starting point that works with a stock app, which people can then customize to suit their needs.

PR #1240, opened 2023-02-25
fix: update types for global prisma instance

### LINKED ISSUE #1239: bug: Global prisma instance type is slightly incorrect

N/A

### Describe the bug

(I don't think this is a bug/feature but rather a nit. However I was forced to select one when creating this issue.)

The current code for the prisma global instance is slightly misleading:
1. It is actually undefined until initialised but the types don't represent that
2. We should probably use nullish coalescing `??` instead of `||` 

The current snippet is this:
```
const globalForPrisma = globalThis as unknown as { prisma: PrismaClient };

export const prisma =
  globalForPrisma.prisma ||
  new PrismaClient({
    log:
      env.NODE_ENV === "development" ? ["query", "error", "warn"] : ["error"],
  });

if (env.NODE_ENV !== "production") globalForPrisma.prisma = prisma;
```

However I propose that it should be:
```
const globalForPrisma = globalThis as unknown as {
  prisma: PrismaClient | undefined;
};

export const prisma =
  globalForPrisma.prisma ??
  new PrismaClient({
    log:
      env.NODE_ENV === "development" ? ["query", "error", "warn"] : ["error"],
  });

if (env.NODE_ENV !== "production") globalForPrisma.prisma = prisma;
```

The code as it currently stands does indeed work. But I see no harm in making it 'more accurate' 🤷 

### Reproduction repo

N/A

### To reproduce

N/A

### Additional information

_No response_

JingLi1998:
Closes #1239 

## ✅ Checklist

---

## Changelog

Updates the global prisma instance types.

---

💯

### DIFF
--- a/.changeset/gorgeous-jars-vanish.md
+++ b/.changeset/gorgeous-jars-vanish.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+Update the type for global prisma instance
--- a/cli/template/extras/src/server/db.ts
+++ b/cli/template/extras/src/server/db.ts
@@ -2,10 +2,12 @@ import { PrismaClient } from "@prisma/client";
 
 import { env } from "~/env.mjs";
 
-const globalForPrisma = globalThis as unknown as { prisma: PrismaClient };
+const globalForPrisma = globalThis as unknown as {
+  prisma: PrismaClient | undefined;
+};
 
 export const prisma =
-  globalForPrisma.prisma ||
+  globalForPrisma.prisma ??
   new PrismaClient({
     log:
       env.NODE_ENV === "development" ? ["query", "error", "warn"] : ["error"],


c-ehrlich:
It does intuitively make sense that this is what the type should be. But I'm not sure if this is worth deviating from [Prisma's recommendations](https://www.prisma.io/docs/guides/database/troubleshooting-orm/help-articles/nextjs-prisma-client-dev-practices) for as there are no known issues with the existing setup.

That being said this might be worth making a PR to Prisma's own docs for. If they are up to change it, then we would of course follow along.

https://github.com/prisma/docs/blob/main/content/300-guides/050-database/880-troubleshooting-orm/100-help-articles/400-nextjs-prisma-client-dev-practices.mdx

juliusmarminge:
Agree, if you get prisma to update their docs we'll follow but I don't see the point in going on our own for something that works. Also,

- the globalForPrisma const only lives in that file, and is not exported.
- immediately after it's initialized

PR #1259, opened 2023-03-02
chore: Remove gitignored next-env.d.ts

bai:
Closes #<issue>

## ✅ Checklist

---

## Changelog

Remove gitignored next-env.d.ts per https://nextjs.org/docs/basic-features/typescript:

> A file named next-env.d.ts will be created at the root of your project. This file ensures Next.js types are picked up by the TypeScript compiler. You should not remove it or edit it as it can change at any time. This file should not be committed and should be ignored by version control (e.g. inside your .gitignore file).

### DIFF
--- a/.changeset/famous-geese-chew.md
+++ b/.changeset/famous-geese-chew.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+Remove gitignored next-env.d.ts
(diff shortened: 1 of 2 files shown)


juliusmarminge:
create-next-app includes it by default in the same way we do so I think we should keep it as is

PR #1263, opened 2023-03-06
fix: search for tsconfig relative to eslintrc

mkreuzmayr:
## ✅ Checklist

---

## Changelog

I created a t3-app in an existing monorepo and I expect eslint to use the config file that is in the project and not in the turborepo root. The current `.eslintrc.cjs` always searches for the `tsconfig.json` in the root of the project. I resolved the path to do a relative seach, which is what most people expect if there lies a `tsconfig.json` next to the `eslintrc.cjs`.

💯

### DIFF
--- a/.changeset/fresh-cars-tease.md
+++ b/.changeset/fresh-cars-tease.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+search for tsconfig relative to eslintrc
--- a/cli/template/base/_eslintrc.cjs
+++ b/cli/template/base/_eslintrc.cjs
@@ -1,3 +1,6 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const path = require("path");
+
 /** @type {import("eslint").Linter.Config} */
 const config = {
   overrides: [
@@ -7,13 +10,13 @@ const config = {
       ],
       files: ["*.ts", "*.tsx"],
       parserOptions: {
-        project: "tsconfig.json",
+        project: path.join(__dirname, "tsconfig.json"),
       },
     },
   ],
   parser: "@typescript-eslint/parser",
   parserOptions: {
-    project: "./tsconfig.json",
+    project: path.join(__dirname, "tsconfig.json"),
   },
   plugins: ["@typescript-eslint"],
   extends: ["next/core-web-vitals", "plugin:@typescript-eslint/recommended"],


Zamiell:
i can confirm that i do this pattern in my monorepos as well.

juliusmarminge:
don't y'all put this stuff in the root config? i usually do this: https://github.com/t3-oss/create-t3-turbo/blob/bb46ce25a43fb6379f69988ea1f7e14f6e032ebd/.eslintrc.js#L8-L13

Zamiell:
I don't think using the `tsconfigRootDir` option obviates the problem. My monorepo has that specified, yet I still have to use `path.join`.

Furthermore, I don't quite understand your config. Why is the root ESLint extending from each project? That seems backwards. The child ESLint configs should be inheriting from the parent, like how normal TypeScript code would work.

Zamiell:
Also, as a side note, the file should have a cjs extension, not a js extension.

mkreuzmayr:
I agree with @Zamiell 

@juliusmarminge Using a single eslint config that spans your whole monorepo is a bad practice, especially if it grows in complexity. If you only have two similar projects in your repo (e.g. all of them are using React) it may be fine.

For monorepos that contain different projects in them (some React, Svelte and NodeJS) where every project needs a dedicated eslint config, your approach is not viable. In addition, it also defeats the whole project structure if configuration files are outside the project scope in which they are used.

mkreuzmayr:
> Also, as a side note, the file should have a cjs extension, not a js extension.

This should not be a problem as long as `type` is not set to `module` in the `package.json`.

Zamiell:
> This should not be a problem as long as type is not set to module in the package.json.

It's not a problem insofar that it causes any lint errors, but I think it is best practice in 2023 to name your ESLint config file `.eslintrc.cjs` regardless of what underlying package.json settings your have, because it makes it very explicit what the format of the file is, and allows you to refactor your package.json settings in the future without introducing any bugs into your codebase.

juliusmarminge:
The typescript parser options doesnt need to be separate, so I like just specifying the tsconfig paths in root.

As for the rules etc those can extend some common config package and also add new rules on their own (I think I left this as a todo for our monorepo).

Not quite sure what the benefits are of duplicating the tsconfig path in each config?

Zamiell:
> The typescript parser options doesnt need to be separate, so I like just specifying the tsconfig paths in root.

They need to be separate, because each package has a separate tsconfig. You wouldn't want to be linting `foo` with the `tsconfig.json` compiler settings of `bar` - that would cause spurious errors in your codebase and other problems. As I explained earlier, the problem becomes more pronounced when you have a node package and a browser package in the same repo.

mkreuzmayr:
It isn´t about tsconfig´s here, it´s about eslintrc´s. Your root `.eslintrc.cjs` defines the linting rules for A and B. Most projects need seperate eslint configs with different rulesets. But nevertheless, if this way works for you, it´s fine, I do not want to force you into doing it differently.

mkreuzmayr:
Thanks for merging. 🚀

PR #1232, opened 2023-02-21
fix: bind configs to variables before exporting to make sure typeerrors are reported

juliusmarminge:
Closes #<issue>

## ✅ Checklist

- [ ] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [ ] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] I performed a functional test on my final commit

---

## Changelog

Errors weren't reported when they were exported immediately:

CleanShot 2023-02-21 at 10 30 56@2x

Binding it to a variable before exporting fixes it

---

## Screenshots

_[Screenshots]_

💯

PR #1265, opened 2023-03-07
fix: unnecessary variable

Zamiell:
## ✅ Checklist

---

## Changelog

i removed unnecessary variables.
afaict [it is idiomatic in eslint/prettier configs](https://typescript-eslint.io/getting-started) to use the format in my PR (rather than the arbitrary format that was there before my PR)


c-ehrlich:
These variables are not unnecessary - see https://github.com/t3-oss/create-t3-app/pull/1232

PR #1273, opened 2023-03-11
feat: add errorformatter

juliusmarminge:
Closes #<issue>

## ✅ Checklist

---

## Changelog

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/www/src/pages/en/usage/trpc.md
+++ b/www/src/pages/en/usage/trpc.md
@@ -95,7 +95,7 @@ You'll immediately notice how good the autocompletion and typesafety is. As soon
 
 ## Inferring errors
 
-By default, `create-t3-app` sets up an [error formatter](https://trpc.io/docs/error-formatting) that let's you infer your Zod Errors if you get validation errors on the backend.
+By default, `create-t3-app` sets up an [error formatter](https://trpc.io/docs/error-formatting) that lets you infer your Zod Errors if you get validation errors on the backend.
 
 Example usage:
 
--- a/www/src/pages/en/usage/trpc.md
+++ b/www/src/pages/en/usage/trpc.md
@@ -93,6 +93,36 @@ const UserPage = () => {
 
 You'll immediately notice how good the autocompletion and typesafety is. As soon as you write `api.`, your routers will show up in autocomplete, and when you select a router, its procedures will show up as well. You'll also get a TypeScript error if your input doesn't match the validator that you defined on the backend.
 
+## Inferring errors
+
+By default, `create-t3-app` sets up an [error formatter](https://trpc.io/docs/error-formatting) that let's you infer your Zod Errors if you get validation errors on the backend.
+
+Example usage:
+
+```tsx
+function MyComponent() {
+  const { mutate, error } = api.post.create.useMutation();
+
+  return (
+    <form onSubmit={(e) => {
+      e.preventDefault();
+      const formData = new FormData(e.currentTarget);
+      mutate({ title: formData.get('title') });
+    }}>
+      <input name="title" />
+      {error?.data?.zodError?.fieldErrors.title && (
+        {/** `mutate` returned with an error on the `title` */}
+        <span className="mb-8 text-red-500">
+          {error.data.zodError.fieldErrors.title}
+        </span>
+      )}
+
+      ...
+    </form>
+  );
+}
+```
+
 ## Files
 
 tRPC requires quite a lot of boilerplate that `create-t3-app` sets up for you. Let's go over the files that are generated:
--- a/.changeset/eighty-masks-invite.md
+++ b/.changeset/eighty-masks-invite.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": minor
+---
+
+feat: add errorformatter for zod errors
--- a/cli/package.json
+++ b/cli/package.json
@@ -71,6 +71,7 @@
     "superjson": "^1.12.2",
     "tsup": "^6.6.3",
     "type-fest": "^3.6.0",
-    "typescript": "^4.9.5"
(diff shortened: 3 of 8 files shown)


nexxeln (commented):
Love it, we should add docs for this too before merging

juliusmarminge:
> Love it, we should add docs for this too before merging

Good shout! What do you think would be the best docs for this?

---

## Inferring errors

By default, tRPC sets up an [error formatter]() that let's you infer your Zod Errors if you get validation errors on the backend.

Example usage:

```
function MyComponent() {
  const mutation = api.post.create.useMutation();

  return (
    <form>
      <input name="title" />
      {mutation.error?.data?.zodError?.fieldErrors.title && (
        <span className="mb-8 text-red-500">
          {mutation.error.data.zodError.fieldErrors.title}
        </span>
      )}
    </form>
  );
}
```

---

Or do you have a better idea?

nexxeln:
Looks good to me

c-ehrlich:
> > Love it, we should add docs for this too before merging
> 
> Good shout! What do you think would be the best docs for this?
> ## Inferring errors
> 
> By default, tRPC sets up an error formatter that let's you infer your Zod Errors if you get validation errors on the backend.
> 
> Example usage:
> 
> ```tsx
> function MyComponent() {
>   const mutation = api.post.create.useMutation();
> 
>   return (
>     <form>
>       <input name="title" />
>       {mutation.error?.data?.zodError?.fieldErrors.title ? (
>         <span className="mb-8 text-red-500">
>           {mutation.error.data.zodError.fieldErrors.title}
>         </span>
>       ) : null}
> 
>       ...
>     </form>
>   );
> }
> ```
> 
> Or do you have a better idea?

docs text should be "lets" instead of "let's"
everything else looks really good.

realized while reading this pr that maybe some of the text in trpc.ts could be made a bit easier to understand overall, but thats for another pr

www/src/pages/en/usage/trpc.md
```diff
@@ -93,6 +93,36 @@ const UserPage = () => {
 
 You'll immediately notice how good the autocompletion and typesafety is. As soon as you write `api.`, your routers will show up in autocomplete, and when you select a router, its procedures will show up as well. You'll also get a TypeScript error if your input doesn't match the validator that you defined on the backend.
 
+## Inferring errors
+
+By default, `create-t3-app` sets up an [error formatter](https://trpc.io/docs/error-formatting) that let's you infer your Zod Errors if you get validation errors on the backend.
```
c-ehrlich, line 98:
```suggestion
By default, `create-t3-app` sets up an [error formatter](https://trpc.io/docs/error-formatting) that lets you infer your Zod Errors if you get validation errors on the backend.
```

nexxeln:
I think we can merge this now

JClackett:
Has anyone run into the scenario where you are updating a resource and so need a data payload but also an id to actually find the resource update:

so the function would look something like this:
```ts
update: protectedProcedure.input(z.object({ id: z.string(), data: customSchemaForResource })).mutation(..
```

now when using the errorFormatter/flatten thing from zod, any errors within the data schema validator get flattened into a "data" key, and so on the frontend you dont know which field caused the error.

What are peoples practices for handling this?

juliusmarminge:
> Has anyone run into the scenario where you are updating a resource and so need a data payload but also an id to actually find the resource update, the function would look something like this:
> 
> ```ts
> update: protectedProcedure.input(z.object({ id: z.string(), data: customSchemaForResource })).mutation(..
> ```
> 
> now when using the errorFormatter/flatten thing from zod, any errors within the data schema validator get flattened into a "data" key, and so on the frontend you dont know which field caused the error.
> 
> What are peoples practices for handling this?

What do you mean? The current formatter would return this if you send a number instead of a string to the default query on a fresh app:

```json
{
  "formErrors": [],
  "fieldErrors": {
    "text": [
      "Expected string, received number"
    ]
  }
}
```

so you have your fieldErrors with the fields that failed

JClackett:
Just curious wow would you structure your trpc input when updating a "todo", for example? as you'll need an ID of some sort and then the actual data to update, do you just put the ID alongside all the fields that are being updated?

juliusmarminge:
Yea probably, i had some procedure recently which posted a review on a product which was something like 

```ts
create: protectedProcedure
  .input(z.object({
    productId: z.string(),
    text: z.string().min(10),
    rating: z.number().min(0).max(5),
  })
  .mutation(...)
```

juliusmarminge:
Ah I see...

Something like this:

```ts
export const exampleRouter = createTRPCRouter({
  hello: publicProcedure
    .input(
      z.object({
        id: z.string(),
        data: z.object({
          text: z.string().min(3),
        }),
      })
    )
    .query(({ input }) => {
      return {
        greeting: `Hello ${input.data.text}`,
      };
    }),
});
```

would return an error like this if some field within the data object is invalid:

```json
{
  "formErrors": [],
  "fieldErrors": {
    "data": [
      "String must contain at least 3 character(s)"
    ]
  }
}
```

Hmm... not sure if you can write a better error formatter to handle that form of input structure

JClackett:
Yeah exactly, I edited my previous comment showing a workaround ill go for as I really dont wana write my own formatter :P

juliusmarminge:
The `.format()` (instead of `.flatten()`) method on zodErrors give you something like this:

```json
{
  "_errors": [],
  "id": {
    "_errors": [
      "Expected string, received number"
    ]
  },
  "data": {
    "_errors": [],
    "text": {
      "_errors": [
        "String must contain at least 3 character(s)"
      ]
    }
  }
}
```

which you could probably work with to make it work for your usecase

PR #1295, opened 2023-03-24
feat: Replace Git Bash check with Mintty check

### LINKED ISSUE #1281: bug: Terminal is detected incorrectly on windows terminal running Git-Bash

  System:

    OS: Windows 10 10.0.19045
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 9.75 GB / 15.88 GB

  Binaries:

    Node: 16.14.2 - D:\programlar\nodejs\node.EXE
    Yarn: 1.22.15 - D:\programlar\nodejs\yarn.CMD
    npm: 9.4.2 - D:\programlar\nodejs\npm.CMD

### Describe the bug

When i run `npm create t3-app@latest` from my windows terminal running git-bash, i can only create the default project. Since windows terminal is interactive i expect to be able to see options and choose accordingly.

### Reproduction repo

https://github.com/t3-oss/create-t3-app

### To reproduce

Run `npm create t3-app@latest` on a windows terminal running git-bash.

### Additional information

This is due to the check of process.env.SHELL containing bash in windows terminal. If a check for terminal instead of shell could be implemented, bug should disappear.

iAverages:
Closes #1281 

## ✅ Checklist

---

## Changelog

Replaced the Git Bash check if a check for Mintty. Git Bash works perfectly fine when using another terminal. Mintty is the issue here and it causes lots of issues with other CLI programs. 

Not too sure if we want two messages here, I think having a notice specific to Git Bash is needed as it is mainly Git Bash users who are having this issue (Since Git Bash uses Mintty by default). 

---

## Screenshots

WSL VSC (Should not be affected by the change)
image

Git Bash + Mintty
image

Git Bash + Windows Terminal
image

💯

### DIFF
--- a/cli/src/cli/index.ts
+++ b/cli/src/cli/index.ts
@@ -157,12 +157,11 @@ export const runCli = async () => {
 
   // Explained below why this is in a try/catch block
   try {
-    if (
-      process.env.SHELL?.toLowerCase().includes("git") &&
-      process.env.SHELL?.includes("bash")
-    ) {
-      logger.warn(`  WARNING: It looks like you are using Git Bash which is non-interactive. Please run create-t3-app with another
+    if (process.env.TERM_PROGRAM?.toLowerCase().includes("mintty")) {
+      logger.warn(`  WARNING: It looks like you are using Mintty which is non-interactive. Please run create-t3-app with another
   terminal such as Windows Terminal or PowerShell if you want to use the interactive CLI.`);
+      logger.warn(`  WARNING: If you are using Git Bash, please use Git Bash from another terminal, such as Windows Terminal 
+      or the terminal build into Visual Studio Code.`);
 
       const error = new Error("Non-interactive environment");
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/cli/src/cli/index.ts
+++ b/cli/src/cli/index.ts
@@ -158,10 +158,9 @@ export const runCli = async () => {
   // Explained below why this is in a try/catch block
   try {
     if (process.env.TERM_PROGRAM?.toLowerCase().includes("mintty")) {
-      logger.warn(`  WARNING: It looks like you are using Mintty which is non-interactive. Please run create-t3-app with another
-  terminal such as Windows Terminal or PowerShell if you want to use the interactive CLI.`);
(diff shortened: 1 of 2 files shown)


juliusmarminge:
Can you test this on windows @nexxeln

iAverages:
Yea I don’t like the two warnings however I feel like we would get lots of questions about Git Bash not working (with mintty). since mintty is installed by default with Git Bash. I can remove the second warning and potentially change the first message though not sure what the best option is.

cli/src/cli/index.ts
```diff
@@ -157,12 +157,11 @@ export const runCli = async () => {
 
   // Explained below why this is in a try/catch block
   try {
-    if (
-      process.env.SHELL?.toLowerCase().includes("git") &&
-      process.env.SHELL?.includes("bash")
-    ) {
-      logger.warn(`  WARNING: It looks like you are using Git Bash which is non-interactive. Please run create-t3-app with another
+    if (process.env.TERM_PROGRAM?.toLowerCase().includes("mintty")) {
+      logger.warn(`  WARNING: It looks like you are using Mintty which is non-interactive. Please run create-t3-app with another
```
c-ehrlich, line 161: Could this just be one warning, like `WARNING: It looks like you are using MinTTY, which is non-interactive. This is most likely because you are using Git Bash. In that case, please use Git Bash from another terminal, such as Windows Terminal or the terminal built into Visual Studio Code.`

iAverages: Sure. That message is much better. I’ll make that change once I get home in a few hours.

juliusmarminge: `WARNING: It looks like you are using MinTTY, which is non-interactive. This is most likely because you are using Git Bash. If that's that case, please use Git Bash from another terminal, such as Windows Terminal. Alternatively, you can provide the arguments from the CLI directly: https://create.t3.gg/en/installation#experimental-usage` to skip the prompts.`

Maybe?

nexxeln: I like julius's solution because it provied a direct solution as well.

iAverages: Changed the message to the one Julius provided.

juliusmarminge (approved):

PR #1307, opened 2023-03-27
fix: add indications for Type Error caused by id property on user object

yoshua70:
Closes #<issue>

## ✅ Checklist

- [x ] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [x ] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)

---

## Changelog

Add comment to indicate that the user should check the shape of the returned object when authenticating with NextAuth providers.
For example the Google OAuth provider does not return an `id` property on the user object causing this line of code `session.user.id = user.id`  to throw a Type Error.

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/cli/template/extras/src/server/auth/base.ts
+++ b/cli/template/extras/src/server/auth/base.ts
@@ -37,7 +37,10 @@ export const authOptions: NextAuthOptions = {
   callbacks: {
     session({ session, user }) {
       if (session.user) {
-        session.user.id = user.id;
+        // Check the shape of the data returned by the providers you use
+        // before adding other properties on the session.
+        // For example Google OAuth provider does not return an id for the user.
+        // session.user.id = user.id;
         // session.user.role = user.role; <-- put other properties on the session here
       }
       return session;
--- a/cli/template/extras/src/server/auth/base.ts
+++ b/cli/template/extras/src/server/auth/base.ts
@@ -40,7 +40,7 @@ export const authOptions: NextAuthOptions = {
         // Check the shape of the data returned by the providers you use
         // before adding other properties on the session.
         // For example Google OAuth provider does not return an id for the user.
-        // session.user.id = user.id;
+        session.user.id = user.id;
         // session.user.role = user.role; <-- put other properties on the session here
       }
       return session;
--- a/.changeset/silver-ears-double.md
+++ b/.changeset/silver-ears-double.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+Fix type error when using google oauth provider due to id property not existing


cli/template/extras/src/server/auth/base.ts
```diff
@@ -37,6 +37,9 @@ export const authOptions: NextAuthOptions = {
   callbacks: {
     session({ session, user }) {
       if (session.user) {
+        // Check the shape of the data returned by the providers you use
+        // before adding other properties on the session.
+        // For example Google OAuth provider does not return an id for the user.
```
juliusmarminge, line 42: are you sure about this?

pretty sure i've gotten an id back when using google oauth?

are you using jwt perhaps?

Dodobrat: I am also getting this error on a fresh project setup with the Discord Provider. The project can't build. 
```
Property 'id' does not exist on type '{ name?: string | null | undefined; email?: string | null | undefined; image?: string | null | undefined; }'.
```
Had to do a check for the key as well : `session.user && "id" in session.user`

yoshua70: Yeah, like @Dodobrat I had to manually check the existence of the `id` field on the `user` object. Don't know why

c-ehrlich: > I am also getting this error on a fresh project setup with the Discord Provider. The project can't build.
> 
> ```
> Property 'id' does not exist on type '{ name?: string | null | undefined; email?: string | null | undefined; image?: string | null | undefined; }'.
> ```
> 
> Had to do a check for the key as well : `session.user && "id" in session.user`
> 
> EDIT: Got it working by creating a new object, by spreading the existing session:
> 
> ```
> session({ session, user }) {
>   return {
>     ...session,
>     user: {
>       ...session.user,
>       id: user.id,
>     },
>   };
> }
> ```

I can't reproduce this. A new t3 app @ latest with all packages and the default discord auth setup builds fine

could you upload a minimal reproduction? ideally with the scaffolded app as the first commit, and any changes you make as another commit

mikahFSmusic: This was an issue with the latest version of next-auth, see issues page for more details. Downgrading next-auth to 4.20.0 helped me with the issue.

juliusmarminge: That can't be the same issue OG faced though, the update came out yesterday and he had issues before that

Dodobrat: Hey @c-ehrlich , maybe the cli was updated or something but now I don't get the type error. 
-- Also forgot to mention that I got it after bumping up the dependencies. --

**Left**: New t3@latest project
**Right**: My project
 
Screenshot 2023-04-03 at 11 18 13

These are the the more important differences in my opinion. 
No point of sharing a new repo if it actually works. The error looks to be from `next-auth`, maybe not the same as @mikahFSmusic mentioned but the version I am on (Right side) is the issue

c-ehrlich: > Hey @c-ehrlich , maybe the cli was updated or something but now I don't get the type error as of this morning. -- Also forgot to mention that I got it after bumping up the dependencies. --
> 
> **Left**: New t3@latest project **Right**: My project
> 
> Screenshot 2023-04-03 at 11 18 13
> 
> These are the the more important differences in my opinion. No point of sharing a new repo if it actually works. The error looks to be from `next-auth`, maybe not the same as @mikahFSmusic mentioned but the version I am on (Right side) is the issue

Your problem was due to breaking changes in next-auth 4.21.0. You can see in the new app that we pinned it to 4.20.1 temporarily.

juliusmarminge:
Also just verified that google does indeed send the user id back:

https://user-images.githubusercontent.com/51714798/229636424-69729acc-fd17-45a6-9dbc-c1e775ee2131.mp4

So I'm going to close this PR. If OG still can't get this working, please open an issue with a reproduction repo and some context of the necessary google oauth setup

PR #1323, opened 2023-04-01
feat: introduce github api typings

PiotrekPKP:
## ✅ Checklist

---

## Changelog

Uses zod to validate github api responses.

---

💯

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -209,6 +209,9 @@ importers:
       treeify:
         specifier: ^1.1.0
         version: 1.1.0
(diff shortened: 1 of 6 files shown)


nexxeln (changes requested):
Thanks for this!

One of the best parts about zod is that you can choose the fields you want from an api response and ignore the rest. Having a huge zod schema doesn't make sense, we should only have the fields we are using.

PiotrekPKP:
Done!

nexxeln (approved):
looks good to me!

Issue #1328, opened 2023-04-03
bug:  NextAuth module augmentation not working. 


mikahFSmusic:
### Provide environment information

 System:
    OS: macOS 11.2.3
    CPU: (8) arm64 Apple M1
    Memory: 147.81 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 19.2.0 - ~/.nvm/versions/node/v19.2.0/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 9.6.3 - ~/.nvm/versions/node/v19.2.0/bin/npm
    Watchman: 2023.01.30.00 - /opt/homebrew/bin/watchman

### Describe the bug

**NOTE: was able to get things working by downgrading next-auth to 4.20.0. It may be worth locking the package to this version in create-t3-app until the issue is fixed or a new solution is added to the docs. 

Once I'm past the prisma db push step, I'm running into the following.

in the auth.ts file, I get an error: 

```
export const authOptions: NextAuthOptions = {
  callbacks: {
    session({ session, user }) {
      if (session.user) {
        session.user.id = user.id; // Error happens here, on session.user.id 
      }
      return session;
    },
  },
```

`Property 'id' does not exist on type '{ name?: string | null | undefined; email?: string | null | undefined; image?: string | null | undefined; }'.ts(2339)`

This is a basic setup, using prisma with mongodb atlas (not that its necessarily relevant to this issue). I've tried numerous things, including adding the module augmentation to a `next-auth.d.ts` file inside of a types folder specified in tsconfig as a typeRoot. I've restarted my typescript server in between changes to make sure, but nothing has worked so far. 

### Reproduction repo

https://github.com/mikahFSmusic/worldbuilders-toolkit

### To reproduce

1. Clone the repo
2. Run `npm run build`
3. You should get the same error. 

Nothing special going on in the code here, just a default setup using create-t3-app.

LegendNihal:
I found the main problem here. When you hover over user on the right side one `session.user.id = user.id;` you can see user is adapterUser which is only available if strategy is set to database. Since we are using Prisma and Prisma doesn't run on Edge runtime, we had to set strategy to jwt. Thats why `user.id` is undefined.

PR #1329, opened 2023-04-03
fix: temporarily pin next-auth version

c-ehrlich:
Closes https://github.com/t3-oss/create-t3-app/issues/1328

## ✅ Checklist

---

## Changelog

- pins next-auth to 4.20.1
- ive also subscribed to https://github.com/nextauthjs/next-auth/issues/7132 so we can hopefully undo this asap

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/.changeset/gold-toes-compete.md
+++ b/.changeset/gold-toes-compete.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+temporarily pin next-auth version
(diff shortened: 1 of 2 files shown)


t3dotgg (approved):
I am sorry

balazsorban44:
For awareness, I gave an update here: https://github.com/nextauthjs/next-auth/issues/7132#issuecomment-1493924161

I am interested to know what you all think about it!

LegendNihal:
I found the main problem here. When you hover over user on the right side one `session.user.id = user.id;` you can see user is adapterUser which is only available if strategy is set to database. Since we are using Prisma and Prisma doesn't run on Edge runtime, we had to set strategy to jwt. Thats why `user.id` is undefined.

Issue #1332, opened 2023-04-03
feat: Opinionated import order


rortan134:
### Is your feature request related to a problem? Please describe.

Is there any plans on creating an opinionated prettier import auto-sorter for the t3 base template?

### Describe the solution you'd like to see

Something like [@trivago/prettier-plugin-sort-imports](https://www.npmjs.com/package/@trivago/prettier-plugin-sort-imports) or [@ianvs/prettier-plugin-sort-imports](https://www.npmjs.com/package/@ianvs/prettier-plugin-sort-imports) would be very nice to have already set-up.

![Code_656t1Kovbi](https://user-images.githubusercontent.com/37253958/229556130-42f99612-4b80-4dd8-a7ff-603b02786e8f.gif)

### Describe alternate solutions

Import order structure example:

```js
importOrder: [
        "^react",
        "^next/*",
        "<THIRD_PARTY_MODULES>",
        "",
        "^@/components/(.*)$",
        "^@/utils/(.*)$",
        "",
        "^@/lib/(.*)$",
        "",
        "^[./]", // local files
        "^@public/*", // root/public
    ],
  ```

c-ehrlich:
I think I'm in favor of this. It's something I have in all of my non trivial projects. We would need to make it work with the custom base path, but that should be doable.

I guess the argument against is "user can add it", and we tend to stay away from opinionated linting/formatting by default. Curious to see what others think.

juliusmarminge:
We wouldn't need to do a custom sort order, the default from ianvs is enough and if devs want their own they can add themselves. Just having the plugin and some basic rules (like no duplicates etc) goes a long way for consistency and keeping git diffs small

FinnDore:
Oh that's true tbf

ronanru:
I like [prettier-plugin-organize-imports](https://www.npmjs.com/package/prettier-plugin-organize-imports). It uses TypeScript LSP to sort imports.

iAverages:
I'd definitely like to see this added :)

its-monotype:
[eslint-plugin-simple-import-sort](https://github.com/lydell/eslint-plugin-simple-import-sort)?

FinnDore:
Is someone doing this, i see a couple of prs but not much movement. i can bash a pr out today if people want?

FinnDore:
So something i forgor was that the default cta app does not incude prettier. so do we only want to include this if they chose tailwind?

FinnDore:
not technically closed yet.

PR #1334, opened 2023-04-04
feat: add prettier functionality for prisma and eslint

squink13:
Closes #<issue>

## ✅ Checklist

---

## Changelog

- Updated prisma and @prisma/client to 4.12.0
- Added prettier-plugin-prisma package to allow formatting of prisma schema files
- Added eslint-config-prettier to handle conflicts with eslint and prettier whenever prettier is installed as per [Next.js](https://nextjs.org/docs/basic-features/eslint#prettier) and [Prettier](https://prettier.io/docs/en/integrating-with-linters.html) recommendations
---

## Screenshots

_[Screenshots]_

image
create-t3-app with all options selected.

You can find the directory [here](https://github.com/squink13/create-t3-app/tree/test-build)

💯


juliusmarminge:
I like the idea here and we should also do #1332.

I think @nexxeln objects though

cli/package.json
```diff
@@ -61,11 +61,13 @@
     "@types/gradient-string": "^1.1.2",
     "@types/inquirer": "^9.0.3",
     "@types/node": "^18.15.5",
+    "eslint-config-prettier": "^8.8.0",
     "next": "^13.2.4",
     "next-auth": "^4.20.1",
-    "prettier": "^2.8.6",
+    "prettier": "^2.8.7",
+    "prettier-plugin-prisma": "^4.12.0",
```
juliusmarminge, line 68: why would we need a prettier plugin here? the prisma cli has a `format` command and their VSCode extension is probably a better recommendation than a 3rd party prettier plugin?

cli/src/installers/prettier.ts
```diff
@@ -0,0 +1,36 @@
+import fs from "fs-extra";
+import path from "path";
+import { PKG_ROOT } from "~/consts.js";
+import { type Installer } from "~/installers/index.js";
+import { addPackageDependency } from "~/utils/addPackageDependency.js";
+
+export const prettierInstaller: Installer = ({ projectDir, packages }) => {
```
juliusmarminge, line 7: i wonder if it'd be a good shout to combine this installer with eslint?

so the default would be no eslint n'or prettier, and you could choose an `eslint+prettier` installer?

c-ehrlich:
I'm mixed about this. The more linting we add, the weirder it gets that we don't ship a "complete"/"opinionated" config. But I do lean in favour of these specific ones making sense. 

Where do we want to stand on this stuff going forward? Where is the line of being too opinionated? Do we run a risk of people not realising that they're meant to add their own prettier and eslint settings? Should we pivot towards providing a reasonable "full loadout", and people can rip it out if they want?

squink13:
I think for a better out of the box experience adding the eslint config package at the very least would be a small improvement. It gets around potential conflicts with linting and formatting. 

The prisma plugin is probably more on the bloat side, its not really necessary like @juliusmarminge mentioned.

cli/src/installers/prettier.ts
squink13: If you're looking at #1332 as well i think moving eslint & prettier to its own option would be nice. Not everyone wants opinionated formatting or linting for every project. It would also mimick the 'create-next-app' eslint option

bai:
If prisma plugin allows us to catch prisma formatting errors on CI I don't mind having it included.

My 2c is that ct3a should ship whatever the maintainers think is the best opinionated stack. Data point of one, but as an End User of ct3a I don't want to bother with linters/formatters/etc - I want to work on my app asap and I will certainly not be bothering with adding/removing linters.

DinoZoidDev (commented):
I believe this is a better solution if you want to abide by Prisma's official formatter with the added benefits of auto-completed relationships.

```json
"editor.defaultFormatter": "esbenp.prettier-vscode",
"[prisma]": {
    "editor.defaultFormatter": "Prisma.prisma"
},
  ```

juliusmarminge:
superseded by #1392

PR #1352, opened 2023-04-08
refactor: add destructuring

Tungetyt:
Closes #<issue>

## ✅ Checklist

- [✅ ] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [ ✅] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ✅] I performed a functional test on my final commit

---

## Changelog
Refactor code by removing unnecessary destructuring and adding session destructuring.

---


cli/template/extras/src/server/api/trpc/with-auth.ts
```diff
@@ -46,10 +46,8 @@ const createInnerTRPCContext = (opts: CreateContextOptions) => {
  * @see https://trpc.io/docs/context
  */
 export const createTRPCContext = async (opts: CreateNextContextOptions) => {
```
nexxeln, line 48: In my opinion, we should destructure wherever possible because its one less variable to think about, one less variable to name. So I think destructuring in the first change is fine. Here I would prefer if we destructure `opts` into `req` and `res` in the function params itself.

nexxeln (changes requested):

cli/template/extras/src/server/api/trpc/with-auth.ts
Tungetyt: In case of destructuring here, we would have to think about two variables `req` and `res` instead of one `opts`.

However, passing destructured `req` and `res` as parameters can make the code more explicit, as it clearly shows that the getServerAuthSession function requires both of them. 

Therefore, I'll go ahead and destructure `req` and `res` directly in the function's parameter list.

Tungetyt: Fixed.

Tungetyt: Fixed

nexxeln: > In case of destructuring here, we would have to think about two variables `req` and `res` instead of one `opts`.
no not really, those are properties coming from the opts object. they aren't variables we declared

nexxeln (approved):
looks good!

PR #1357, opened 2023-04-12
fix: incorrect behavior of SKIP_ENV_VALIDATION

mickyngub:
The current logic for checking SKIP_ENV_VALIDATION will pass if SKIP_ENV_VALIDATION is assigned to any value other than an empty string (even `false`). Ideally, SKIP_ENV_VALIDATION should only pass when its assigned value is `true` and nothing else.

Closes #

## ✅ Checklist

---

## Changelog

fix the logic for determining whether it should validate environment variables or not

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/.changeset/blue-walls-admire.md
+++ b/.changeset/blue-walls-admire.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+fix: incorrect behavior of SKIP_ENV_VALIDATION
--- a/cli/template/base/next.config.mjs
+++ b/cli/template/base/next.config.mjs
@@ -2,7 +2,7 @@
  * Run `build` or `dev` with `SKIP_ENV_VALIDATION` to skip env validation.
  * This is especially useful for Docker builds.
  */
-!process.env.SKIP_ENV_VALIDATION && (await import("./src/env.mjs"));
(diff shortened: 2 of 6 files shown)


juliusmarminge:
I think the best way to do coercion here is to accept

| input | resolved |
|--------|--------|
| <undefined>| false|
| "" | false |
| "false" | false |
| "0" | false |

any other values resolves to true

c-ehrlich:
Agreed, the ideal behaviour would be that truthy values evaluate to true

mickyngub:
Thank you for your input. Given that type of environment variables string, only `process.env.SKIP_ENV_VALIDATION=`, `process.env.SKIP_ENV_VALIDATION=""` or doesn't specify it at all (undefined) will be falsy, any other values will be evaluated to truthy (including "false", and "0"). 

In response to @c-ehrlich, I disagree that all truthy values should evaluate to true, since "false" or "0" would be evaluated to true if that were the case. 

Regarding @juliusmarminge's point, while undefined and "" evaluate to false by default, "false", "0" evaluate to true, we could address this by specifically checking if the value is "false" or "0" and returns false, however, I believe this approach might result in a more verbose implementation.

PR #1368, opened 2023-04-19
docs: Mount DATABASE_URL as secret instead of ARGS

upalatucci:
Hi everyone.

Using DATABASE_URL as args and pushing the image in a public registry can lead to exposing database sensitive information.

An attacker can easily inspect the docker image and see username and password if they are part of the DATABASE_URL.

So instead of doing that, I suggest mounting DATABASE_URL as a secret if we need to render pages on build time. Using ENVS is great but in my case, I had to retrieve some db info on build time for SSG.

Mount secret for the build and use ENV to run the `server.js` 

`database_url.txt` should be a file with just the database url with username and password.
We can also add some text to clarify what is this file or just add a command to take the DATABASE_URL value from the `.env` file and put it into a file

DOC:
https://docs.docker.com/compose/compose-file/compose-file-v3/#secrets
https://docs.docker.com/engine/reference/builder/#arg

## ✅ Checklist

---

## Changelog

Documentation about dockerfile

---

💯


juliusmarminge:
cc @c-ehrlich

www/src/pages/en/deployment/docker.md
```diff
@@ -84,15 +84,16 @@ RUN \
...
 ARG NEXT_PUBLIC_CLIENTVAR
 WORKDIR /app
 COPY --from=deps /app/node_modules ./node_modules
 COPY . .
 
 # ENV NEXT_TELEMETRY_DISABLED 1
 
-RUN \
+RUN --mount=type=secret,id=DATABASE_URL \
+ export DATABASE_URL=$(cat /run/secrets/DATABASE_URL) && \
```
juliusmarminge, line 96: This doesnt seem very nice if you have to add stuff here everytime?

Iirc 
- Environment are secure and not bundled in the image but only injected when you start the container.

- Args are unsafe and bundled in the image when it builds

Am I wrong?

upalatucci: Yes right!

upalatucci: btw here how i use that with github actions: 
https://github.com/upalatucci/biblioteca-nextjs/blob/main/.github/workflows/deploy.yml#L33

There is also a secret-file argument for that action but don't have experience with that

upalatucci: In addition to that, I think this file (.env or a separate file) should be included in the .dockerignore as in Dockerfile we can find the command `COPY . .` . With that command we would copy the file with a secret in the dockerfile and this is another vulnerability

c-ehrlich:
Is this a safety concern only in situations where an image is uploaded to a public registry?

If yes, I think we need to weigh this edge case vs the slightly worse dx for everyone else. Maybe just a comment in the documentation would be enough in this case.

If no (ie there is also a a safety concern in other situations) then it seems fairly clear that we should make this change.

upalatucci:
@juliusmarminge do you think i have to close this pr? Does not seems to be necessary anymore.

PR #1370, opened 2023-04-20
feat: Enforce `NEXT_PUBLIC_` env key on client with TS

### LINKED ISSUE #1369: feat: Migrate `env.mjs` to `.ts`

It currently doesn't stop us from defining non-`NEXT_PUBLIC_` key as part of the client env.

If we migrate the `.env.mjs` to `.env.ts`, we can utilize the `satisfies` keyword to enforce the ``NEXT_PUBLIC_${string}`` as the client env key.
 
However, I'm actually pretty curious why `.mjs` for `env.mjs`? It would be good if someone can answer this🙏

I personally have the working solution in my personal project, and I am more than happy to help.

-

### Additional information

_No response_

chungweileong94:
Closes #1369

## ✅ Checklist

---

## Changelog

Enforce `NEXT_PUBLIC_` env key on client, using other non-`NEXT_PUBLIC_` key on client will give you a type error.

---

## Screenshots

image

💯

### DIFF
--- a/cli/template/base/src/env.mjs
+++ b/cli/template/base/src/env.mjs
@@ -1,3 +1,5 @@
+// @ts-check
+
 import { z } from "zod";
 
 /**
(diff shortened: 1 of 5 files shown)


cli/template/base/src/env.mjs
```diff
@@ -1,3 +1,5 @@
+// @ts-check
```
juliusmarminge, line 1: No need for these as we use `checkJs` in tsconfig

bai, line 1: Do we need ts-check in JS files, given that [we have checkJs set to true in tsconfig](https://github.com/t3-oss/create-t3-app/blob/next/tsconfig.json#L11)?

chungweileong94: Yeah, I was thinking the same, but it wasn't working on VSCode for some reason. it's working now after restarting the instance.
Thx for pointing that out!

chungweileong94: Yeah, you are right. it was some issue with my VSCode earlier, thx for pointing that out!

juliusmarminge (approved):

PR #1375, opened 2023-04-22
feat: Add support for using bun package manager (bunx / bun install)

### LINKED ISSUE #1366: feat: support initializing with Bun

  System:
    OS: macOS 13.4
    CPU: (8) arm64 Apple M1
    Memory: 102.36 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.18.0 - ~/Library/Caches/fnm_multishells/78864_1681854261277/bin/node
    npm: 8.19.2 - ~/Library/Caches/fnm_multishells/78864_1681854261277/bin/npm
    Watchman: 2023.01.02.00 - /opt/homebrew/bin/watchman

### Describe the bug

If running create-t3-app with bunx it defaults to installing depends with `npm` rather than `bun install`

### To reproduce

`bunx create-t3-app`

Should prompt for "install depends with bun" instead of npm

### Additional information

I have a patch ready if you're ok with this change, the performance diff is massive...

CleanShot 2023-04-19 at 00 13 20

Zeko369:
Closes #1366

## ✅ Checklist

---

## Changelog

- [x] Add case to detect if the CLI was ran via `bunx` and use `bun install` in that case

---

## Screenshots

`bun install` goes brrrrrrrrr 🚀🚀🚀🚀

Speeeeed

💯


juliusmarminge:
Should we add CI for this?

FleetAdmiralJakob:
Any updates on this?

.github/workflows/e2e.yml
Zeko369, line 1: Hey @juliusmarminge, happy to see this PR finally getting done.. sorry I completely forgot about it. What do you think about also adding CI to run the create+install here since that's the more important part of my changes here

(Since next build isn't supported by `bun --bun` yet running `bun run build` will actually just spawn a node process and run it the same way (as i.e. `pnpm run build` or `yarn build`, running `bun install` on the other hand will be different than pnpm run dev)

Zeko369: `bun --bun run build` will override any node scripts internally to use bun and will actually run webpack with bun, `bun run build` will just use bun to run the `node_modules/.bin/webpack` script (which will be faster than npm run build, but only the first 200ms of resolving and spawning the node process for the actual script)

juliusmarminge: How do you suggest we test this?

juliusmarminge:
Did you delete your fork or something? I can't seem to push anymore...

Zeko369:
No 🤔, you should have permissions? Do you want to me to manually invite you to make sure?

.github/workflows/e2e.yml
Zeko369: Nevermind... was reading the wrong line in your commit you actually call it with bun 🤦‍♂️

Didn't see this line, just saw the one after doing `bun build` 🤦‍♂️🤦‍♂️🤦‍♂️
`- run: cd cli && bun run dist/index.js ../../ci-bun --default`

This is what happens when you try to reply to a PR after studying Calc2 for the past couple of hours... 😅

The only thing might be to check if bun was actually used to install (since it falls back to npm in the current version if I'm not mistaken with), so something like this between the run steps?

```
- name: Check File Creation
  run: 'if [ ! -f "bun.lockb" ]; then echo "File does not exist!"; exit 1; fi'
``` 

Lemme try this actually 🧪

juliusmarminge:
Nvm i was in the wrong directory lmaoo

.github/workflows/e2e.yml
```diff
@@ -69,3 +69,66 @@ jobs:
...
+          bun-version: latest
+
+      - run: cd cli && bun run dist/index.js ../../ci-bun --default
+
+      - name: We should have a Bun lockfile
+        run: |
+          if [ ! -f "../../ci-bun/bun.lockb" ]; then
+            echo "Bun lockfile not found"
+            exit 1
+          fi
```
juliusmarminge, line 125: @Zeko369 I added this to make sure we get a bun lockfile which should indicate the CLI uses bun to install the dependencies, so that when we then run `bun run build` we should have all the dependencies and such installed using Bun and the build script should succeed

.github/workflows/e2e.yml
```diff
@@ -69,3 +69,66 @@ jobs:
...
+
+      # FIXME: This doesn't actually run the build script using bun, since Next.js doesn't support it yet.
+      # But you should still be able to use `bun` as a package manager for any Next.js app.
+      # If/When Next.js supports it, we should be able to run `bun --bun run build` here to override any Node binaries.
+      # See: https://bun.sh/docs/cli/bunx#shebangs
```
juliusmarminge, line 130: also added this for future reference about the fact that we're still building with Node, just using Bun as a package manager

.github/workflows/e2e.yml
juliusmarminge: Yea we had the same thought :)

Zeko369: great-minds-think-alike

juliusmarminge (approved):
🚀

juliusmarminge:
Tested on `create-t3-app@7.19.0-alpha-bun.1`

juliusmarminge:
Released in `7.19`

PR #1291, opened 2023-03-22
feat: dynamic OG images for docs #986

iAverages:
Adds dynamic OG images for documentation. 

Closes #986

## ✅ Checklist

- [x] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [x] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] I performed a functional test on my final commit

---

## Changelog

- Setup Astro for server builds
- Setup route for generating dynamic images using Satori

## Screenshots

English:
home
intro
image
image

Large title (Title font gets reduced to avoid overflow and word breaking)
image

Other LFR Languages:
spanish
french
pt
russian

RTL Languages:
image
image
image

💯

PR #1385, opened 2023-04-26
fix: Error with older git version failing to initialize git repo

iAverages:
Closes #<issue>

## ✅ Checklist

---

## Changelog

When creating a project using the CLI with git v2.25.0 you would receive the following error (if you print the error caught [here](https://github.com/iAverages/create-t3-app/blob/37b7653188c0929101afc0e3ff1fc7a0a9c2fd64/cli/src/helpers/git.ts#L131))
```bash
error: refname refs/heads/master not found
fatal: Branch rename failed
```
I believe this happens because no commits have been made after initializing the repo so the ref has not been created.

The command to set the branch was changed to change the HEAD ref using `git symbolic-ref HEAD ${branch}`

There was also a logic with the command selection based on the version. If git updated to v3 then the CLI would use the old logic until the minor version reached 28.

---

## Screenshots

Before:
Broken

After:
Fixed

💯


cli/src/helpers/git.ts
```diff
@@ -112,9 +112,11 @@ export const initializeGit = async (projectDir: string) => {
 
     // --initial-branch flag was added in git v2.28.0
     const { major, minor } = getGitVersion();
-    if (major < 2 || minor < 28) {
+    if (major < 2 || (major == 2 && minor < 28)) {
       await execa("git", ["init"], { cwd: projectDir });
-      await execa("git", ["branch", "-m", branchName], { cwd: projectDir });
+      await execa("git", ["symbolic-ref", "HEAD", `refs/heads/${branchName}`], {
```
c-ehrlich, line 117: maybe add a comment here. idk how commonly used it is, but i had never heard of the symbolic-ref command before

c-ehrlich (commented):
i think this issue snuck in because this code block predates our current git initialization workflow

also good call on editing the major/minor check, won't make a difference for now but might save us a lot of confusion at some point in the future

iAverages:
Is Vercel having issues or did I break deployments with #1291? It's strange how the preview deployments worked before but after it merged deployments broke. It might be that you need to change the build dir to .vercel since the Vercel adapter outputs there instead.

cli/src/helpers/git.ts
iAverages: Added a comment along with a link to the superuser answer I found about it.

juliusmarminge:
> Is Vercel having issues or did I break deployments with #1291? It's strange how the preview deployments worked before but after it merged deployments broke. It might be that you need to change the build dir to .vercel since the Vercel adapter outputs there instead.

Its complaiming about sharp

iAverages:
Updated Astro and sharp, seems to have resolved the deployment issue.

iAverages:
Huh thats so strange. Is it the same error?

juliusmarminge:
> Huh thats so strange. Is it the same error?

Bad cache setup... added .vercel to turbo outputs now. Hoping it's resolved!

PR #1390, opened 2023-04-28
fix: clarify comments regarding prisma use of postgresql

denkristoffer:
Reading https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#string string seems to map to Text already in postgresql, so I think this part of the comment can be removed.

## ✅ Checklist

---

## Changelog

Remove mention of postgresql's need for `@db.Text`  in the Prisma schema


c-ehrlich (approved):
this seems correct, nice catch and thanks for the pr :)

PR #1397, opened 2023-05-03
feat: Added JsonProtocol as the Default Prisma Engine for Serverless …

brieyla1:
…speedups

Closes #<issue>

## ✅ Checklist

- ✅ I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- ✅ The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- ✅ I performed a functional test on my final commit

---

## Changelog

This pull request proposes the enabling of the JsonProtocol by default. The rationale for this change is based on the performance improvements provided by the JsonProtocol without compromising stability, even though it is still a preview feature. Empirical evidence gathered from large-scale projects, with 100+ Prisma calls, demonstrates the significant time reductions enabled by this optimization, particularly in serverless environments such as Vercel.
Another point worth adding is that most projects using nextJS create-t3-app will be deployed on serverless platforms such as vercel, where the real impact is.

Technical Explanation:

Stability: Although the JsonProtocol is currently a preview feature, it has shown no notable stability issues when tested extensively in large projects with numerous Prisma calls. The successful integration and operation of the JsonProtocol in these projects indicate that enabling it by default will not compromise the stability of the t3-create-react package.

Performance Improvements: The redesigned JsonProtocol has eliminated the need for parsing the DMMF (Data Model Meta Format), a time-consuming process that occurred in the previous GraphQL-based wire protocol. The JsonProtocol instead uses plain JSON, which significantly reduces the communication overhead between the Prisma Client and the query engine.

Serverless Environment Benefits: As serverless platforms like Vercel have specific constraints regarding cold start times and resource consumption, the performance improvements offered by the JsonProtocol are particularly advantageous. By default, enabling the JsonProtocol can result in faster response times and a more efficient use of resources in such environments.

Considering these benefits, enabling the JsonProtocol by default in the t3-create-react package will provide users with a more performant experience without sacrificing stability. This change is expected to have a positive impact on the broader community and promote the adoption of the t3-create-react package for both small and large-scale projects.

---

## Screenshots

CleanShot 2023-05-03 at 13 49 26@2x

💯

### DIFF
--- a/.changeset/curvy-dolls-hammer.md
+++ b/.changeset/curvy-dolls-hammer.md
@@ -1,0 +1,13 @@
+---
+"create-t3-app": minor
+---
+
+feat: Added JsonProtocol as the Default Prisma Engine for Serverless speedups
+
+modified:
+
+    - cli/template/extras/prisma/schema/base.prisma
+
+    - cli/template/extras/prisma/schema/with-auth.prisma
+
+Description: This pull request proposes the enabling of the JsonProtocol by default. The rationale for this change is based on the performance improvements provided by the JsonProtocol without compromising stability, even though it is still a preview feature. Empirical evidence gathered from large-scale projects, with 100+ Prisma calls, demonstrates the significant time reductions enabled by this optimization, particularly in serverless environments such as Vercel. Another point worth adding is that most projects using nextJS create-t3-app will be deployed on serverless platforms such as vercel, where the real impact is.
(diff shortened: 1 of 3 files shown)


c-ehrlich (approved):
JsonProtocol already feels prod ready. Let's add it :)

PR #1405, opened 2023-05-05
feat: print error.cause if not a ZodError

J-Zet:
Refactor the errorFormatter function to handle error.cause more cleanly. 

If error.cause is undefined, it will not be included in the output.  If it is an instance of ZodError, it will be included in the zodError property of the output, and if it is not, it will be included in the cause property of the output.

Closes #<issue>

## ✅ Checklist

---

## Changelog
- Adds / does not swallow the `cause` property of the TRPCError.
---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/cli/template/extras/src/server/api/trpc/base.ts
+++ b/cli/template/extras/src/server/api/trpc/base.ts
@@ -51,20 +51,22 @@ export const createTRPCContext = (_opts: CreateNextContextOptions) => {
  * errors on the backend.
  */
 import { initTRPC } from "@trpc/server";
+import { TRPCErrorShape } from "@trpc/server/rpc";
 import superjson from "superjson";
 import { ZodError } from "zod";
 
 const t = initTRPC.context<typeof createTRPCContext>().create({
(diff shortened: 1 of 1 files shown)


cli/template/extras/src/server/api/trpc/base.ts
```diff
@@ -51,20 +51,22 @@ export const createTRPCContext = (_opts: CreateNextContextOptions) => {
...
-        zodError:
-          error.cause instanceof ZodError ? error.cause.flatten() : null,
-      },
-    };
+    const data: TRPCErrorShape["data"] = { ...shape.data };
+    if (error.cause) {
+      if (error.cause instanceof ZodError) {
+        data.zodError = error.cause.flatten();
+      } else {
+        data.cause = error.cause;
```
juliusmarminge, line 66: This is a little scary - I don't think you wanna send every single error down to the client, they may contain information not intended for the public (there is also `onError` in the API handler you can use to send logs of the error, which is only accessible on the server.

PR #1392, opened 2023-04-29
feat: add import order sorting via prettier

FinnDore:
Closes #1332 

## ✅ Checklist

- [x] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [x] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] I performed a functional test on my final commit

---

## Changelog

Added @ianvs/prettier-plugin-sort-import to the cli, docs and generated cta apps

---

## Screenshots

it also seems to format code blocks which is cool 
image

💯

PR #1413, opened 2023-05-08
chore: added prettier import order sorting to the monorepo

FinnDore:
Closes #<issue> 

## ✅ Checklist

---

## Changelog

In #1392 we may be adding prettier import sorting to generated cta apps. this pr adds the same but to the monorepo instead.

looking over the format changes i think they all look fine. i couldn't find any of the problem ones that we don't in #1392 

---

## Screenshots

no lol

💯


FinnDore:
Wondering if i should revert the formats on the template files... probbaly could be handled in #1413. then again that would cause format linting to fail...

cli/template/extras/src/server/api/trpc/base.ts
juliusmarminge, line 1: god damnit... don't want to change these files...

cli/src/utils/renderVersionWarning.ts
```diff
@@ -1,6 +1,14 @@
-import { execSync } from "child_process";
 import { getVersion } from "./getT3Version.js";
 import { logger } from "./logger.js";
+import { execSync } from "child_process";
```
FinnDore, line 3: Looks like its done it again

juliusmarminge: what's wrong here?

FinnDore: https://github.com/t3-oss/create-t3-app/pull/1392#discussion_r1186910626
its the line bellow. the comment and import were further down before.

juliusmarminge: oh right

PR #1426, opened 2023-05-17
fix: Minor fixes

andreifilip123:
* Use octokit instead of fetch
* filter out versions under 5.10.3
* fix mobile dropdowns
* add readme


juliusmarminge:
Target t3-upgrade instead. We'll ship it all once its finished

andreifilip123:
Let me add the readme as well first

juliusmarminge (approved):

PR #1425, opened 2023-05-17
feat: added an option for prettier and a extended eslint config

FinnDore:
Closes #1332

## ✅ Checklist

- [x] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [x] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] I performed a functional test on my final commit

---

## Changelog

_[Short description of what has changed]_

---

## Screenshots

_[Screenshots]_

💯

PR #1427, opened 2023-05-18
feat: added an option for prettier and a extended eslint config (#1425)

juliusmarminge:
cc @FinnDore

### DIFF
--- a/cli/src/cli/index.ts
+++ b/cli/src/cli/index.ts
@@ -120,8 +120,8 @@ export const runCli = async () => {
     )
     /** @experimental - Used for CI E2E tests. Used in conjunction with `--CI` to skip prompting. */
     .option(
-      "-p, --prettier-and-extended-eslint",
-      "Explicitly tell the CLI to use a custom import alias",
+      "--prettier-and-extended-eslint [boolean]",
+      "Opt in to using prettier and an extended eslint config",
       defaultOptions.flags.prettierAndExtendedEslint,
     )
     /** END CI-FLAGS */
--- a/cli/src/cli/index.ts
+++ b/cli/src/cli/index.ts
@@ -2,8 +2,10 @@ import chalk from "chalk";
 import { Command } from "commander";
 import inquirer from "inquirer";
 import { CREATE_T3_APP, DEFAULT_APP_NAME } from "~/consts.js";
-import { type AvailablePackages } from "~/installers/index.js";
-import { availablePackages } from "~/installers/index.js";
+import {
+  availablePackages,
+  type AvailablePackages,
+} from "~/installers/index.js";
 import { getVersion } from "~/utils/getT3Version.js";
 import { getUserPkgManager } from "~/utils/getUserPkgManager.js";
 import { logger } from "~/utils/logger.js";
@@ -15,6 +17,7 @@ interface CliFlags {
   noInstall: boolean;
   default: boolean;
   importAlias: string;
+  prettierAndExtendedEslint: boolean;
 
   /** @internal Used in CI. */
   CI: boolean;
@@ -47,6 +50,7 @@ const defaultOptions: CliResults = {
     prisma: false,
     nextAuth: false,
     importAlias: "~/",
+    prettierAndExtendedEslint: false,
   },
 };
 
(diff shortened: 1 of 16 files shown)


FinnDore:
ah some of the e2e's. ill get on that

FinnDore:
#1428  :)

cli/src/installers/dependencyVersionMap.ts
```diff
@@ -26,5 +23,14 @@ export const dependencyVersionMap = {
...
   superjson: "1.12.2",
+
+  // Prettier
+  prettier: "^2.8.8",
+  "prettier-plugin-tailwindcss": "^0.2.8",
+  "@types/prettier": "^2.7.2",
+  "@ianvs/prettier-plugin-sort-imports": "^3.7.2",
+
+  // Eslint
+  "eslint-config-next": "^13.4.2",
```
juliusmarminge, line 34: so this part doesn't make sense to me. why is `eslint-config-next` not in the default but a more opinionated option, strict ts linting, is?

cli/template/extras/config/_eslintrc.cjs
```diff
@@ -0,0 +1,43 @@
...
+      parserOptions: {
+        project: path.join(__dirname, "tsconfig.json"),
+      },
+    },
+  ],
+  parser: "@typescript-eslint/parser",
+  parserOptions: {
+    project: path.join(__dirname, "tsconfig.json"),
+  },
+  plugins: ["@typescript-eslint"],
```
juliusmarminge, line 21: do we wanna slap some react-hook lint rules in here?

also, eslint-config-prettier?

cli/template/extras/config/_prettier.config.cjs
juliusmarminge, line 1: would it be better if we just build these files? very small diff to have them separate files + very easy to modify them typesafely?

cli/template/extras/config/prettier-with-tailwind.config.cjs
juliusmarminge, line 1: ^^

www/.prettierrc.cjs
```diff
@@ -5,6 +5,7 @@ const config = {
   plugins: [
     ...baseConfig.plugins,
     require.resolve("prettier-plugin-astro"),
+    require.resolve("@ianvs/prettier-plugin-sort-imports"),
```
juliusmarminge, line 8: isn't this already in root?

juliusmarminge (commented):
left a few comments @FinnDore

juliusmarminge (commented):
left a few comments @FinnDore

cli/src/cli/index.ts
```diff
@@ -329,3 +345,16 @@ const promptImportAlias = async (): Promise<string> => {
...
   return importAlias;
 };
+
+const promtPrettierAndExtendedEslint = async (): Promise<boolean> => {
+  const { prettierAndExtendedEslint } = await inquirer.prompt<
+    Pick<CliFlags, "prettierAndExtendedEslint">
+  >({
+    name: "prettierAndExtendedEslint",
+    type: "confirm",
+    message: "Would you like to use prettier and extended an eslint config?",
```
c-ehrlich, line 355:
```suggestion
    message: "Would you like to use prettier and an extended eslint config?",
```

cli/src/installers/dependencyVersionMap.ts
FinnDore: Think i forgot it in both configs. Imo we should have everything that is in the more strict version at the moment in the base version, so i dont really know what we want in the more strict version, those hooks rules you mentioned maybe?

www/.prettierrc.cjs
FinnDore: ye, fixed

FinnDore:
image
A bunch of my comments are pending. not sure what thats about.

FinnDore:
Ive closed #1431 as i think we probably need to work out how this eslint config should look. Im very much for the addtion of optional prettier however im unsure what the extended eslint config should look like as i never really go deep on eslint configs. Julius has potentially mentioned some hook rules might be good for example. Welcome peoples thoughts here :)

Issue #1452, opened 2023-05-26
feat: Make T3 edge ready


FleetAdmiralJakob:
### Is your feature request related to a problem? Please describe.

I think it would be a great addition to the T3 stack if he gets edge ready.

### Describe the solution you'd like to see

Dont rly know. Maybe switch the ORM.

### Describe alternate solutions

-

FleetAdmiralJakob:
Maybe Drizzle (after the last video)

JesseKoldewijn:
I think it would be a good thing to give an ORM select option where you could use either prisma or drizzle instead of fully switching.

Stoffberg:
The T3-Stack team has expressed a preference for avoiding the use of "bleeding edge" technologies. Consequently, while Drizzle represents an interesting potential upgrade, they advise waiting until it matures and reaches its first release (v1) before considering a switch. During this interim period, a viable solution might be to create a community-maintained package or branch dedicated to implementing all the latest features. This approach could allow us to explore and take advantage of emerging technologies on smaller projects, while maintaining stability and reliability in our primary branch.

JesseKoldewijn:
> The T3-Stack team has expressed a preference for avoiding the use of "bleeding edge" technologies. Consequently, while Drizzle represents an interesting potential upgrade, they advise waiting until it matures and reaches its first release (v1) before considering a switch. During this interim period, a viable solution might be to create a community-maintained package or branch dedicated to implementing all the latest features. This approach could allow us to take advantage of newer technologies on smaller projects, while maintaining stability and reliability in the primary branch.

Understandable, I was mainly in the boat of including the choice between prisma and ie drizzle when it's ready 👍

juliusmarminge:
We're a bit blocked by https://github.com/nextauthjs/next-auth/pull/7443 but other than that #1461 & #1446 should address this

JesseKoldewijn:
Awesome! Glad to hear. Haven't gotten around to use drizzle a whole lot myself yet but from what I've experienced in some smaller projects I got bothe the setup and perf is so much better than prisma.

juliusmarminge:
I've mostly been using Kysely lately but there I do something like this:

<img width="326" alt="CleanShot 2023-06-04 at 22 32 48@2x" src="https://github.com/t3-oss/create-t3-app/assets/51714798/e87e4eeb-9176-40bb-ab09-acfb70af77f4">

[...]

PR #1461, opened 2023-06-04
feat: add drizzle

juliusmarminge:
Closes #<issue>

## ✅ Checklist

---

## Changelog

---

## Screenshots

_[Screenshots]_

💯


cli/template/extras/config/drizzle.config.ts
```diff
@@ -0,0 +1,8 @@
+import type { Config } from "drizzle-kit";
```
juliusmarminge, line 1:
```suggestion
import { type Config } from "drizzle-kit";
```

cli/src/installers/dependencyVersionMap.ts
```diff
@@ -11,6 +11,13 @@ export const dependencyVersionMap = {
   prisma: "^4.14.0",
   "@prisma/client": "^4.14.0",
 
+  // Drizzle
+  "drizzle-orm": "^0.26.3",
+  "drizzle-kit": "^0.18.1",
+  dotenv: "^16.1.4",
+  // REVIEW: Is it fine including pscale by default?
```
c-ehrlich, line 18: not sure how i feel about this considering the opinion we've had in the past that the "core stack" should always be free. 

is the reasoning here that this is our best way to be edge ready?

cli/template/extras/src/server/api/routers/example/with-auth-drizzle.ts
```diff
@@ -0,0 +1,24 @@
...
+  hello: publicProcedure
+    .input(z.object({ text: z.string() }))
+    .query(({ input }) => {
+      return {
+        greeting: `Hello ${input.text}`,
+      };
+    }),
+
+  getAll: publicProcedure.query(({ ctx }) => {
+    return ctx.db.query.example.findMany();
```
c-ehrlich, line 18: is this correct?

cli/template/extras/src/server/api/routers/example/with-drizzle.ts
```diff
@@ -0,0 +1,15 @@
...
+export const exampleRouter = createTRPCRouter({
+  hello: publicProcedure
+    .input(z.object({ text: z.string() }))
+    .query(({ input }) => {
+      return {
+        greeting: `Hello ${input.text}`,
+      };
+    }),
+  getAll: publicProcedure.query(({ ctx }) => {
+    return ctx.db.query.example.findMany();
```
c-ehrlich, line 13: is this correct?

c-ehrlich (commented):
will do a functional test tomorrow :)

cli/src/cli/index.ts
```diff
@@ -152,7 +161,16 @@ export const runCli = async () => {
     if (cliResults.flags.trpc) cliResults.packages.push("trpc");
     if (cliResults.flags.tailwind) cliResults.packages.push("tailwind");
     if (cliResults.flags.prisma) cliResults.packages.push("prisma");
+    if (cliResults.flags.drizzle) cliResults.packages.push("drizzle");
     if (cliResults.flags.nextAuth) cliResults.packages.push("nextAuth");
+
+    if (cliResults.flags.prisma && cliResults.flags.drizzle) {
+      console.warn(
+        "Incompatible combination Prisma + Drizzle. Falling back to Prisma only.",
```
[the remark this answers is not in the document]
juliusmarminge: this is just for the CI flags

when you're using the CLI it will keep prompting you if you add invalid stuff (see `validate` in the `promptPackages`:

image

cli/src/helpers/logNextSteps.ts
```diff
@@ -23,10 +23,13 @@ export const logNextSteps = ({
...
   if (packages?.prisma.inUse) {
-    logger.info(
-      `  ${pkgManager === "npm" ? "npx" : pkgManager} prisma db push`,
-    );
+    logger.info(`  ${pkgManager === "npm" ? "npm run" : pkgManager} db:push`);
   }
 
   logger.info(`  ${pkgManager === "npm" ? "npm run" : pkgManager} dev`);
+
+  logger.warn(
```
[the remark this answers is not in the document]
juliusmarminge: yes

cli/src/installers/dependencyVersionMap.ts
juliusmarminge: idk we could take it one step further and prompt for what driver they want - but that's though when it comes to the nextauth adapter as it will differ a bit.

until nextauth ships the adapter and handles all of this for you, maybe we can just go for pscale?

the biggest selling point here is the edge compat and selecting the normal mysql driver wouldn't make sense for that.

idk what's best - hence i left the comment pointing it out - but definetely something to think about

cli/template/extras/src/server/api/routers/example/with-auth-drizzle.ts
juliusmarminge: yep 😎 

CleanShot 2023-06-04 at 21 33 54@2x

cli/template/extras/src/server/api/routers/example/with-drizzle.ts
juliusmarminge: ^^

t3dotgg (commented):
next-auth + drizzle isn't a super happy story at the moment 😓

juliusmarminge:
> next-auth + drizzle isn't a super happy story at the moment 😓 

Why? Worked when I tried it 🙄

.changeset/fast-teachers-explain.md
```diff
@@ -0,0 +1,22 @@
...
+To make the different ORM options as similar as possible, some minor changes has also been made to the Prisma installer:
+
+- a new script `db:push` has been added and is included in both ORM options.
+- the prisma client has been renamed to `db` in the trpc context - you now access your database client like
+  ```ts
+  examples: publicProcedure.query((opts) => {
+     // prisma
+     opts.ctx.db.example.findMany()
+     // drizzle
+     opts.ctx.cb.query.example.findMany()
```
michalbundyra, line 18:
```suggestion
     opts.ctx.db.query.example.findMany()
```

FleetAdmiralJakob:
Fixes #1452

cli/template/extras/config/drizzle.config.ts
```diff
@@ -0,0 +1,9 @@
+import * as dotenv from "dotenv";
+import { type Config } from "drizzle-kit";
+
+dotenv.config();
+
+export default {
+  schema: "./src/server/db/schema.ts",
+  connectionString: process.env.DATABASE_URL,
```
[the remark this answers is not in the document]
juliusmarminge: Drizzle doesnt load the envs themselves (this config is used for drizzle-kit i.e when you db:push)

c-ehrlich (changes requested):
For me this requires one more thing to be ready to release: a dev db for drizzle.

some thoughts/questions related to this:
- drizzle requires much bigger differences in how the schema are defined between different dialects of sql. see their [mysql](https://orm.drizzle.team/docs/column-types/mysql) vs [sqlite](https://orm.drizzle.team/docs/column-types/sqlite) docs.
- this makes our current solution of shipping a sqlite database and then letting people switch to whatever they want less ideal - imo we are being pushed into prescribing a default sql dialect which is the same for dev and prod.
- i believe this dialect should be mysql - it's a database that we all like, and it's the database planetscale uses, which is already very relevant now and might become even more relevant with edge + their adapter.
- we could pivot to something like turso which would enable sqlite everywhere, but although turso seems awesome i'm personally leaning against this
- no matter what our db/orm is, we should always provide a working dev db setup. this setup should work offline, ie "just provision a planetscale instance for dev" is not a valid solution.

where this seems to lead me is a mysql docker container that is started alongside the dev server. i don't love this, as developing completely without docker feels great. but i don't see any other compelling options for drizzle.

thoughts? any alternatives that i'm forgetting? also if we do this, should we also do the same for prisma?

(there's other things that could probably be improved as well, but i'm happy to leave them for future PRs so we can get this out, and merged into the app dir PR)

cli/src/cli/index.ts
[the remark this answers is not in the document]
GermanJablo: Wouldn't a separate question be better? If the default settings are what most users use, it's comfortable to move forward without thinking about the options.

matibox:
nextauthjs/next-auth#8258 https://authjs.dev/reference/adapter/drizzle

Next Auth now has an official drizzle adapter

juliusmarminge:
This comment is buried way up but here's the latest one for y'all:

```sh
pnpm create t3-app@7.18.1-beta.19d5f83
```

cli/src/cli/index.ts
```diff
@@ -1,12 +1,11 @@
...
 import chalk from "chalk";
 import { Command } from "commander";
-import inquirer from "inquirer";
 
 import { CREATE_T3_APP, DEFAULT_APP_NAME } from "~/consts.js";
-import {
-  availablePackages,
-  type AvailablePackages,
-} from "~/installers/index.js";
+// import { getUserPkgManager } from "~/utils/getUserPkgManager.js";
```
c-ehrlich, line 6: two commented imports here

cli/src/installers/nextAuth.ts
```diff
@@ -8,8 +8,12 @@ import { addPackageDependency } from "~/utils/addPackageDependency.js";
 
 export const nextAuthInstaller: Installer = ({ projectDir, packages }) => {
   const usingPrisma = packages?.prisma.inUse;
+  const usingDrizzle = packages?.drizzle.inUse;
+
   const deps: AvailableDependencies[] = ["next-auth"];
   if (usingPrisma) deps.push("@next-auth/prisma-adapter");
+  // This adapter is not yet available on npm so we have our own inhoused
```
c-ehrlich, line 15: isnt it out now?

cli/template/extras/src/server/db/drizzle-schema-auth.ts
```diff
@@ -0,0 +1,105 @@
...
+} from "drizzle-orm/mysql-core";
+import { type AdapterAccount } from "next-auth/adapters";
+
+/**
+ * This is an example of how to use the multi-project schema feature of Drizzle ORM. Use the same
+ * database instance for multiple projects.
+ *
+ * @see https://orm.drizzle.team/docs/goodies#multi-project-schema
+ */
+export const mysqlTable = mysqlTableCreator((name) => `project1_${name}`);
```
c-ehrlich, line 20: could we be slick and replace `project1` with the name of the inited project?

cli/src/cli/index.ts
```diff
@@ -170,38 +190,128 @@ export const runCli = async () => {
...
+            initialValue: "none",
+          });
+        },
+        database: () => {
+          return p.select({
+            message: "What database ORM would you like to use?",
+            options: [
+              { value: "none", label: "None" },
+              { value: "prisma", label: "Prisma" },
+              { value: "drizzle", label: "Drizzle" },
```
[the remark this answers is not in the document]
c-ehrlich: how much additional work would it be? maintaining separate `index-drizzle.ts`, `drizzle-schema-base.ts`, and `drizzle-schema-auth.ts` for each dialect doesn't seem like too big of a deal.

here's roughly the list of possible things that feel like they could possibly be included (doesn't mean we should include all of them)

- dev: planetscale, prod: planetscale (what we already have)
- dev: mysql container, prod: mysql connection url
- dev: mysql container, prod: planetscale (is this even a good idea?)
- dev: postgres container, prod: postgres connection url
- (we could also consider setups with other serverless/edge ready databases)

some of these require docker setups, which means at least a `docker-compose.yml`, and maybe also changes to `package.json` to create a smooth dev experience.

i'm not for or against any of this, not sure what's the right way to go. what do you think?

cli/src/cli/index.ts
```diff
@@ -149,14 +154,29 @@ export const runCli = async () => {
...
+        "Incompatible combination Prisma + Drizzle. Falling back to Prisma only."
+      );
+      cliResults.flags.drizzle = false;
+      cliResults.packages = cliResults.packages.filter(
+        (pkg) => pkg !== "drizzle"
+      );
+    }
+
+    return cliResults;
+  }
```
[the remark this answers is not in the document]
c-ehrlich: i think i'm in favour of exiting with a non-zero code. 2 (Invalid command line usage or invalid input) seems to make sense.

why do you think we shouldn't? does it relate to our own actions that make sure the CLI works in CI?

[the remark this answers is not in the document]
juliusmarminge: > why do you think we shouldn't? does it relate to our own actions that make sure the CLI works in CI?

Yes, it would probably be more work to make the CI matrix work in that case, maybe we can have a check in a catch block or something that just gracefully exits if `process.env.CI` is true?

cli/src/cli/index.ts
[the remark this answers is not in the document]
juliusmarminge: I think either way it can be a follow-up in another release. Dont need to block this one before we have made the final decision?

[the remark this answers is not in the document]
c-ehrlich: agreed

cli/src/cli/index.ts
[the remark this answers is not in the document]
c-ehrlich: yea sounds good

juliusmarminge:
Looks like we can't quite do that @c-ehrlich.

Can we somehow detect this exit in the action and exit the action gracefully from there?

Or maybe we can add an if-statement to the action
```yaml
- name: Check valid config
  if: ${{ matrix.prisma == "true" && matrix.drizzle == "true" }} # unsure of exact syntax
  run: # graceful exit
```

juliusmarminge:
Nice, now the Drizzle + Prisma exits early: https://github.com/t3-oss/create-t3-app/actions/runs/6052746581/job/16426820256?pr=1461

c-ehrlich (approved):
lookin good

www/src/pages/pl/faq.md
```diff
@@ -18,6 +18,7 @@ Jeżeli nie znasz poszczególnych technologi użytych w projekcie, skorzystaj z
 - [Prisma](https://prisma.io)
 - [Tailwind CSS](https://tailwindcss.com)
 - [tRPC](https://trpc.io)
+- [Drizzle](https://drizzle.team/)
```
AndriiSherman, line 21: If create-t3-app is using drizzle-orm itself, I think it would be better to have a link to drizzle-orm docs rather than the drizzle team. I guess this one should be much better: https://orm.drizzle.team/docs/overview. We have also updated an overview page and the documentation itself.

www/src/pages/en/faq.mdx
```diff
@@ -21,6 +21,7 @@ If you are not familiar with the different technologies used in this project, pl
 - [Prisma](https://prisma.io)
 - [Tailwind CSS](https://tailwindcss.com)
 - [tRPC](https://trpc.io)
+- [Drizzle](https://drizzle.team/)
```
AndriiSherman, line 24: If create-t3-app is using drizzle-orm itself, I think it would be better to have a link to drizzle-orm docs rather than the drizzle team. I guess this one should be much better: https://orm.drizzle.team/docs/overview. We have also updated an overview page and the documentation itself.

cli/template/extras/src/server/db/drizzle-schema-base.ts
```diff
@@ -0,0 +1,34 @@
...
+ * database instance for multiple projects.
+ *
+ * @see https://orm.drizzle.team/docs/goodies#multi-project-schema
+ */
+export const mysqlTable = mysqlTableCreator((name) => `project1_${name}`);
+
+export const example = mysqlTable(
+  "example",
+  {
+    id: serial("id").primaryKey(),
```
AndriiSherman, line 24: We will change the documentation examples for MySQL to use `bigint().primaryKey().autoincrement()` instead of serial. Using SERIAL may not explicitly convey the use of a bigint autoincrement field and could lead to issues. We will also make the same adjustment for PostgreSQL, especially since PostgreSQL is beginning to recommend IDENTITY columns over SERIAL column types. While we won't deprecate these options, we don't want to feature them in our examples or promote them as the recommended way to set up a schema

So it would be great, if create-t3-app will also recommend this id columns:
```ts
id: bigint("id", { mode: "number" }).primaryKey().autoincrement()
```
We will update drizzle orm docs with this recommendation as well

cli/template/extras/src/server/db/drizzle-schema-auth.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export const example = mysqlTable(
+  "example",
+  {
+    id: serial("id").primaryKey(),
```
AndriiSherman, line 26: We will change the documentation examples for MySQL to use bigint().primaryKey().autoincrement() instead of serial. Using SERIAL may not explicitly convey the use of a bigint autoincrement field and could lead to issues. We will also make the same adjustment for PostgreSQL, especially since PostgreSQL is beginning to recommend IDENTITY columns over SERIAL column types. While we won't deprecate these options, we don't want to feature them in our examples or promote them as the recommended way to set up a schema

So it would be great, if create-t3-app will also recommend this id columns:
```ts
id: bigint("id", { mode: "number" }).primaryKey().autoincrement()
```
We will update drizzle orm docs with this recommendation as well

AndriiSherman (commented):
I'd like to share some ideas for useful changes to Drizzle in `create-t3-app`

www/src/pages/pl/faq.md
juliusmarminge: Awesome, we'll add proper docs later so we can adress that then I suppose

cli/template/extras/src/server/db/drizzle-schema-base.ts
juliusmarminge: I changed it to use bigint, would love for unsigned support in the near future though 🙏🏼

AndriiSherman: These recommendations are intended for new users, which is why we will update our documentation for now (while there is no unsigned option). Furthermore, the unsigned option should be included in the next release, so in any case, it will be available soon.

I'm curious about why you want to have the unsigned option and why bigint alone is not sufficient. BIGINT has a range of up to 9,223,372,036,854,775,807. I understand that with unsigned, you can extend this range by a factor of two, but even with a signed option, I can imagine that there may be so many rows in a table that reaching 9,223,372,036,854,775,807 would be challenging.

Consider, for example, a messages table receiving 10 billion messages per day. In this scenario, it would take approximately 2,479,401 years to reach the limits. This calculation assumes we are discussing changing from serial to bigint

juliusmarminge: Yea obviously a BIGINT will be enough for most apps, but I just like being explicit what data will be in the column, and since we won't have any negatives it just makes sense that it's unsigned...

AndriiSherman: Agree! 
Will add support for it soon!

juliusmarminge: I think we're good to merge this then!

Exciting, app dir next :)

juliusmarminge:
Ship 🚢

checkerschaf:
Thanks a lot to all of you who have contributed to this 🙏

PR #1466, opened 2023-06-08
feat: remove possible trailing slash from appName prompt

### LINKED ISSUE #1460: feat: CLI add a ./ option for the project name

I want to use ./ to create all the files in my already create directory, but I get this error: `App name must consist of only lowercase alphanumeric characters, '-', and '_'`

If I type in the app name ./ all the files are going into the the directory I'm currently in

No alternative

### Additional information

_No response_

brunoeduardodev:
Closes #1460

## ✅ Checklist

---

## Changelog

_Ignore trailing slashes when prompting the app name_

---

### When prompting `./`

image

### When prompting any path with a trailing slash

image

💯

### DIFF
--- a/.changeset/perfect-geckos-behave.md
+++ b/.changeset/perfect-geckos-behave.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": minor
+---
+
+Ignore trailing slashes when prompting the app name.
--- a/cli/src/cli/index.ts
+++ b/cli/src/cli/index.ts
@@ -231,11 +231,6 @@ const promptAppName = async (): Promise<string> => {
     },
   });
 
-  // Remove potential trailing slash
-  if (appName.endsWith("/")) {
-    return appName.slice(0, -1);
-  }
-
   return appName;
 };
 
--- a/cli/src/cli/index.ts
+++ b/cli/src/cli/index.ts
@@ -231,6 +231,11 @@ const promptAppName = async (): Promise<string> => {
     },
   });
 
+  // Remove potential trailing slash
+  if (appName.endsWith("/")) {
+    return appName.slice(0, -1);
+  }
+
   return appName;
 };
 
--- a/cli/src/utils/parseNameAndPath.ts
+++ b/cli/src/utils/parseNameAndPath.ts
@@ -1,3 +1,4 @@
+import { removeTrailingSlash } from "./removeTrailingSlash.js";
 import pathModule from "path";
 
 /**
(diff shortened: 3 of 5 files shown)


brunoeduardodev:
I don't fully love this solution since there's a duplicated checking, both on the validate function and after getting the prompt result, do y'all have any suggestions on this?

brunoeduardodev:
Currently the PR doesn't affect `appName`s provided by the CLI, the project creation works, but the project name, when passing a trailing slash direct to the CLI results in a empty `appName`, in this case might make sense doing this validation on `parseNameAndPath` but I'm not sure if this would be too much, thoughts?

c-ehrlich:
> Currently the PR doesn't affect `appName`s provided by the CLI, the project creation works, but the project name, when passing a trailing slash direct to the CLI results in a empty `appName`, in this case might make sense doing this validation on `parseNameAndPath` but I'm not sure if this would be too much, thoughts?

not sure about implementation details. but in terms of how it works, it should work the same for:
- typing the app name in response to the prompt
- adding it as an arg to the CLI
- CI (same as above I think?)

brunoeduardodev:
I've just changed the implementation to remove the trailing space after the initial `appName` is get, so potential trailing slashes are removed from the prompt, args or ci,

usecase:

#### `./` by args (appName becomes cwd after parsing)
image


#### `./` by prompt (appName becomes cwd after parsing)
image


#### `./some/path/` by args (appName becomes "path" after parsing)
image

#### `./some/path/` by prompt (appName becomes "path" after parsing)
image

c-ehrlich (changes requested):
this looks really clean :)

one remaining issue is that it doesn't catch a name that is just `/` and thus turned into an empty string, causing the command line to crash (if the user submits an empty string the normal way, their app gets named `my_t3_app` instead).

this behaviour should definitely be taken care of, but not sure what's better:
- doing the same thing as when an empty string is provided at the start
- turning an app name of `/` into `.` as it feels like this might be the user intent (they almost certainly didnt intend to init at the root of their file system)
- an error of some sort

brunoeduardodev:
I see, currently the create-t3-app gives a validation error when a user prompts `/`, and when passing `/` through args it goes to the root dir,

So my suggestion would be keeping this behavior, since there might be some kind of weird scenario where it's wanted to create the app through the root,

this would mean we are going to remove trailing slashes only if there's at least 2 characters, do you have thoughts?

c-ehrlich:
> I see, currently the create-t3-app gives a validation error when a user prompts `/`, and when passing `/` through args it goes to the root dir,
> 
> So my suggestion would be keeping this behavior, since there might be some kind of weird scenario where it's wanted to create the app through the root,
> 
> this would mean we are going to remove trailing slashes only if there's at least 2 characters, do you have thoughts?

Sounds good!

brunoeduardodev:
Just added a commit adding this behavior, I wonder if we should give a hint to the user telling then to use `.` if they want to install the project on the current dir.

I've also moved the verification to an util function, not sure if `utils` folder was the best place for that.

### From prompt

image

### From args

image

c-ehrlich (approved):
This looks good :) Thanks for the PR!

PR #1470, opened 2023-06-15
docs: added steps to deploy to a server

sacsbrainz:
Closes #<issue>

## ✅ Checklist

---

## Changelog

_[added steps to deploy to a server in the documentation]_

---

## Screenshots

_Screenshots_

💯


c-ehrlich (commented):
Thanks for the PR and all the work that went into it. Unfortunately we've decided not to merge it.

The reason is that we don't want to encourage users to deploy a T3 app to bare metal. Anything we publish in the docs ends up being an "endorsed" way of doing things, and we don't see bare metal as that. Our endorsed way to deploy to a long-running server is using a Docker container on your own server, or a serverful PaaS like Fly.io or FlightControl.

However it's a very well written page, and you should definitely publish it on dev.to, or your own blog.

PR #1476, opened 2023-06-19
feat: typescript-eslint@v6

### LINKED ISSUE #1474: feat: Use typescript-eslint@v6's reworked configs

Following up on #1035 -> #1036: typescript-eslint is soon going to release a new v6 major version with reworked configurations. You can read up on it here: https://typescript-eslint.io/blog/announcing-typescript-eslint-v6-beta

Specifically, the [configurations are reworked](https://typescript-eslint.io/blog/announcing-typescript-eslint-v6-beta#reworked-configuration-names) so that the recommended equivalents to what create-t3-app uses now are:

* `'plugin:@typescript-eslint/recommended-type-checked'`
* `'plugin:@typescript-eslint/stylistic-type-checked'`

I'd like to send a PR to try the new configs out here. I like this repo and think you'll all do a good job of letting us know if you hate the changes! ❤️ 

https://github.com/typescript-eslint/typescript-eslint/issues/6760

I suppose create-t3-app could opt to stay on typescript-eslint@v5, or use your own bespoke lint configs... neither option seems very appealing to me. 😉 

### Additional information

Please let me know if I'm being annoying / overly intrusive with this. The v6 has been a long time coming, and I really want to get the configs right.

JoshuaKGoldberg:
Closes #1474 (ref: https://github.com/typescript-eslint/typescript-eslint/issues/6760)

## ✅ Checklist

---

## Changelog

Upgraded typescript-eslint to v6, with reworked ESLint configurations.

You can read https://typescript-eslint.io/blog/announcing-typescript-eslint-v6-beta#user-facing-breaking-changes for the rationale behind the config changes. Essentially, the new recommended starter configs are:

* `"plugin:@typescript-eslint/recommended-type-checked"`
* `"plugin:@typescript-eslint/stylistic-type-checked"`

I've commented any changes I suspect you might disagree with inline, for discussion. Would love to know what you all think! 

💯


.eslintrc.cjs
```diff
@@ -19,7 +19,15 @@ const config = {
...
   },
+  overrides: [
+    // Template files don't have reliable type information
+    {
+      files: ["./cli/template/**/*.{ts,tsx}"],
+      extends: ["plugin:@typescript-eslint/disable-type-checked"],
+    },
+  ],
   rules: {
+    "@typescript-eslint/no-explicit-any": "error",
```
JoshuaKGoldberg, line 30: We right now have `no-explicit-any` moved into the _`strict`_ config for v6. But I understand you were expecting it to be enabled. I'm now wondering if we should keep it at _`recommended`_?

For context, _`strict`_ is something we recommend for more confident TypeScript users. https://v6--typescript-eslint.netlify.app/linting/configs#strict

cli/src/helpers/installDependencies.ts
```diff
@@ -7,9 +7,9 @@ import {
 } from "~/utils/getUserPkgManager.js";
 import { logger } from "~/utils/logger.js";
 
-type Options = {
+interface Options {
```
JoshuaKGoldberg, line 10: https://v6--typescript-eslint.netlify.app/rules/consistent-type-definitions is the rule to enforce which of `interface` or `type` you prefer to use consistently. In v6 right now it's moved from _`strict`_ to _`stylistic`_. How does that make you feel?

For reference, a couple of full text regex searches on the `next` branch:

* `interface .+ \{`: 44 results in 40 files
* `type .+ = \{`: 14 results in 12 files

cli/src/helpers/installDependencies.ts
```diff
@@ -43,8 +43,12 @@ const runInstallCommand = async (
...
           }
         });
-        pnpmSubprocess.on("error", (e) => rej(e));
-        pnpmSubprocess.on("close", () => res());
+        pnpmSubprocess.on("error", (e) => {
+          rej(e);
+        });
+        pnpmSubprocess.on("close", () => {
+          res();
+        });
```
JoshuaKGoldberg, line 51: https://v6--typescript-eslint.netlify.app/rules/no-confusing-void-expression flags cases around `void` such this: a function that returns a value being used in a place that expects it to return `void`. It's in the _`stylistic-type-checked`_ config in v6, after not being in any recommended configs in v5. I worry it's too opinionated and will annoy people. Thoughts?

cli/src/helpers/installPackages.ts
```diff
@@ -6,9 +6,9 @@ import {
 } from "~/installers/index.js";
 import { logger } from "~/utils/logger.js";
 
-type InstallPackagesOptions = {
+type InstallPackagesOptions = InstallerOptions & {
   packages: PkgInstallerMap;
-} & InstallerOptions;
+};
```
JoshuaKGoldberg, line 11: https://v6--typescript-eslint.netlify.app/rules/sort-type-constituents - is newly in the _`stylistic`_ config. While it does have an autofixer, I do worry that it's too nitpicky and will annoy people. Thoughts?

cli/template/base/_eslintrc.cjs
```diff
@@ -1,25 +1,15 @@
...
-      parserOptions: {
-        project: path.join(__dirname, "tsconfig.json"),
-      },
-    },
-  ],
+module.exports = {
   parser: "@typescript-eslint/parser",
   parserOptions: {
-    project: path.join(__dirname, "tsconfig.json"),
+    project: true,
```
JoshuaKGoldberg, line 5: [`project: true` is already in typescript-eslint@v5](https://typescript-eslint.io/linting/typed-linting). Just figured I'd simplify things a bit while I'm here. 😄

package.json
```diff
@@ -65,5 +65,12 @@
     "prettier": "^2.8.8",
     "turbo": "^1.9.6",
     "typescript": "^5.0.4"
+  },
+  "resolutions": {
+    "@typescript-eslint/eslint-plugin": "6.0.0-alpha.158",
+    "@typescript-eslint/parser": "6.0.0-alpha.158",
+    "@typescript-eslint/scope-manager": "6.0.0-alpha.158",
+    "@typescript-eslint/types": "6.0.0-alpha.158",
+    "@typescript-eslint/visitor-keys": "6.0.0-alpha.158"
```
JoshuaKGoldberg, line 74: Re versioning: I'm guessing you won't want to merge this PR until we're out of alpha/beta/rc and in `6.0.0` stable.

upgrade/tsconfig.tsbuildinfo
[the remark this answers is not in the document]
juliusmarminge: no

package.json
juliusmarminge: 👍🏼 

why are the resolutions needed though?

cli/src/helpers/installPackages.ts
juliusmarminge: i think i like it, makes it more like interface where the things you extend comes first

cli/src/helpers/installDependencies.ts
juliusmarminge: hmm i do see some sort of value for it but i'd probably disable it for most of my projects.

definetely on the more opinionated side and i think, as far as i'm concerned, it's on the wrong side of the line at least for me.

.eslintrc.cjs
juliusmarminge: so what's the main difference in strictness between

`strict-type-checked` and the old `recommended-requiring-type-checking`?

upgrade/tsconfig.tsbuildinfo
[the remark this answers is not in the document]
JoshuaKGoldberg: :+1: #1477 -> #1478

.eslintrc.cjs
JoshuaKGoldberg: The _`strict`_ & _`strict-type-checked`_ rulesets are a lot more opinionated. We phrase them as _"Additional strict rules that can also catch bugs but are more opinionated than recommended rules."_. In v6 they're supersets of the respective _`recommended`_ & _`recommended-type-checked`_ configs.

https://v6--typescript-eslint.netlify.app/rules/?supported-rules=xrecommended-strict shows the rules that are in _`strict*`_ and not _`recommended*`_.

cli/src/helpers/installDependencies.ts
JoshuaKGoldberg: Makes sense - I'll disable it in this PR & undo the changes made for it.

cli/src/helpers/installPackages.ts
JoshuaKGoldberg: Great - if we end up removing it from the config, I can manually re-anble it in this PR.

cli/template/base/_eslintrc.cjs
JoshuaKGoldberg: Huh you're right! Great catch, TIL. 

Looking deeper, the `module.exports` property is typed as `NodeJS.Module.exports: any`. So putting a `/** @type */` is ineffective because the property is seen as `any` first. 🤷

juliusmarminge:
Do you also have updates on the new config stuff in eslint? https://eslint.org/docs/latest/use/configure/configuration-files-new

I'm interested in trying that out but last time i tried it was a bit broken

.eslintrc.cjs
c-ehrlich: after reading through the list, i would definitely want about 2/3 of the strict rules and am neutral-but-could-be-pushed-positive on the remaining 1/3. 

what is the motivation for moving `no-explicit-any` to the strict rules?

JoshuaKGoldberg: IIRC it was that a lot of projects end up needing to use `any` (or at least their developers think they do), and we're trying to not be in people's way as much with the recommended rulesets. In a sense this is us capitulating to how new & intermediate TypeScript developers write code (the occasional `any`, not always doing type safe things). Capitulating means less friction in turning on the recommended rulesets, at the cost of being less useful out-of-the-box.

JoshuaKGoldberg:
> Do you also have updates on the new config stuff in eslint? [eslint.org/docs/latest/use/configure/configuration-files-new](https://eslint.org/docs/latest/use/configure/configuration-files-new)

Not yet 😞 but once it's past stage 3 (https://github.com/eslint/eslint/issues/13481) I'd be happy to!

cli/src/helpers/installDependencies.ts
c-ehrlich: this is one of the few that i don't feel strongly about.

my personal preference is "type by default, interface only if necessary (ie need to extend)" but judging by these stats that's not actually what we're doing currently.

what's the eslint team's opinion on this?

JoshuaKGoldberg: We in typescript-eslint[^1] have `interface` over `type` as the default in our recommended rulesets, but I don't believe any of us feel extremely strongly about it. I personally don't. IMO the most important thing for a codebase is to standardize on one to reduce confusion.

I personally have leaned towards `interface`s because they've historically had better error messages in edge cases than `type`. Though over the last couple of years, TypeScript has gotten a lot better at using names for types. So that's not as powerful a motivator as it used to be...

[^1]: Sorry to nitpick - we're a separate team from ESLint core and don't want to accidentally give the impression we're speaking for them 😅

.eslintrc.cjs
nickmccurdy: Do you think educating those users to try to use `unknown` instead could help adoption?

cli/src/helpers/installDependencies.ts
nickmccurdy: Personally, I think `interface` is superior. The official TypeScript wiki recommends preferring it because of performance options and the ability to augment types.

.eslintrc.cjs
JoshuaKGoldberg: Yes, definitely. A lot of these lint rules are better from user education - and can themselves act as user education!

Which reminds me, https://typescript-eslint.io/rules/no-explicit-any could do a bit better in that regard. Filed https://github.com/typescript-eslint/typescript-eslint/issues/7136. Thanks! 😄

cli/src/helpers/installDependencies.ts
JoshuaKGoldberg: Just to be clear - would you like me to disable the rule and revert its changes? I'm happy to, just not sure from the discussion 😅

JoshuaKGoldberg:
Btw, I just updated to typescript-eslint@v6.0.0 stable. 🥳

cli/template/base/package.json
```diff
@@ -20,8 +20,8 @@
     "@types/node": "^18.16.0",
     "@types/react": "^18.2.6",
     "@types/react-dom": "^18.2.4",
-    "@typescript-eslint/eslint-plugin": "^5.59.6",
-    "@typescript-eslint/parser": "^5.59.6",
+    "@typescript-eslint/eslint-plugin": "6.0.0-alpha.158",
```
juliusmarminge, line 23: Missed some

package.json
```diff
@@ -65,5 +65,12 @@
     "prettier": "^2.8.8",
     "turbo": "^1.9.6",
     "typescript": "^5.0.4"
+  },
+  "resolutions": {
+    "@typescript-eslint/eslint-plugin": "6.0.0",
```
juliusmarminge, line 70: Did we need these now with the stable relewse?

juliusmarminge (changes requested):

cli/template/base/package.json
JoshuaKGoldberg: well that explains some issues

package.json
JoshuaKGoldberg: maybe not! https://github.com/t3-oss/create-t3-app/pull/1476#discussion_r1260288775

juliusmarminge (approved):

.eslintrc.cjs
JoshuaKGoldberg: Just following up, I finally got around to documenting this: https://github.com/typescript-eslint/typescript-eslint/pull/8370

PR #1482, opened 2023-06-22
feat: add reminder to commit after creating app

benmarg:
Closes #<issue>

## ✅ Checklist

---

## Changelog

Added a reminder to git commit after creating an application with the CLI

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/.changeset/late-starfishes-kiss.md
+++ b/.changeset/late-starfishes-kiss.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": minor
+---
+
+reminder to git commit after creating an application with the CLI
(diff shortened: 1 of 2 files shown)


nexxeln (approved):

juliusmarminge:
do we really want this?

not commiting in the CLI was a decision we made cause people should have the ability to pick their own message and when to commit, 

adding it as a next step feels like we're enforcing them doing this step which begs the question why does the CLI not do it for them?

benmarg:
IMO this feels like the best middleground. Committing in the CLI is a bit opinionated like you say, but by doing it this way, the user can still opt to choose whatever name they want(they don't have to copy and paste the next step, they can always just plug in their own commit message). 

Including some reminder to commit is important because it's an easy step to forget. How would a first time user of the CLI think to check/know that the CLI stages but doesn't commit for you?

PR #1493, opened 2023-06-27
fix: Disallow trailing commas in cli template code to match the generated prettier config

### LINKED ISSUE #1492: bug: Prettier strips trailing commas in default generated index.tsx NextAuth function

  System:
    OS: macOS 13.3.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 3.40 GB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 16.19.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.3 - /usr/local/bin/npm
    pnpm: 8.6.5 - /usr/local/bin/pnpm
    Watchman: 2023.02.27.00 - /usr/local/bin/watchman

### Describe the bug

Basically the prettier configs are mismatched between the project & the generated app. If you create a project with NextAuth enabled, the below snippet will get generated and prettier will immediately strip the trailing newline. This is happening because the prettier config in the project includes `trailingComma: "all",` but the generated project does not. I think you'd either want to make [this](https://github.com/t3-oss/create-t3-app/blob/next/cli/template/extras/config/prettier.config.cjs) file allow trailing commas in the generated project OR add `trailingComma: "none",` to [this](https://github.com/t3-oss/create-t3-app/blob/next/cli/.prettierrc.cjs)

```
function AuthShowcase() {
  const { data: sessionData } = useSession();

  const { data: secretMessage } = api.example.getSecretMessage.useQuery(
    undefined, // no input
    { enabled: sessionData?.user !== undefined },
  );
```

### Reproduction repo

N/A

### To reproduce

Create any app with NextAuth enabled and run prettier on the generated index.tsx file.

### Additional information

_No response_

dcottr:
Closes #1492

## ✅ Checklist

---

## Changelog

- Disallow trailing commas in cli template code to match the generated prettier config

---

## Screenshots

N/A

💯

### DIFF
--- a/.changeset/gentle-impalas-doubt.md
+++ b/.changeset/gentle-impalas-doubt.md
@@ -2,4 +2,4 @@
 "create-t3-app": patch
 ---
 
-Remove trailing newlines in generated app
+Remove trailing newlines in generated app that aren't allowed by the generated prettier config
--- a/.changeset/gentle-impalas-doubt.md
+++ b/.changeset/gentle-impalas-doubt.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+Remove trailing newlines in generated app
--- a/cli/.prettierrc.cjs
+++ b/cli/.prettierrc.cjs
@@ -5,6 +5,7 @@ const config = {
   ...baseConfig,
   plugins: [...baseConfig.plugins, "prettier-plugin-tailwindcss"],
   tailwindConfig: "./template/extras/config/tailwind.config.ts",
+  trailingComma: "none"
 };
 
 module.exports = config;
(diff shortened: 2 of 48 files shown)


dcottr:
I feel like this got out of hand in terms of diff size, lmk if the preference is to just allow trailing commas on the generated project (that would make the diff smaller). I assumed the preference is to keep the generated app as simple and with as barebone configs as possible

juliusmarminge (commented):
The default should be ES5: https://prettier.io/docs/en/options.html#trailing-commas

dcottr:
> The default should be ES5: https://prettier.io/docs/en/options.html#trailing-commas

Good call thanks, switched it to override to es5 & re-ran prettier.

juliusmarminge (approved):

PR #1562, opened 2023-09-11
chore: ship less files to users with the cli

juliusmarminge:
no need to ship type declarations or src files to users. the CLI already includes a lot of files as is


juliusmarminge:
Before:

CleanShot 2023-09-12 at 00 14 02@2x

After:

CleanShot 2023-09-12 at 00 14 25@2x

PR #1564, opened 2023-09-12
fix(drizzle): yell if default url isn't cahnged

juliusmarminge:
The default example url is a valid url so t3-env wont crash the app, so your app just silently wont work. Adding a refine checking for the default case to prevent this and make sure the app crash before you've changed the URL

### DIFF
--- a/cli/template/extras/src/env/with-auth-db.mjs
+++ b/cli/template/extras/src/env/with-auth-db.mjs
@@ -7,7 +7,13 @@ export const env = createEnv({
    * isn't built with invalid env vars.
    */
   server: {
-    DATABASE_URL: z.string().url(),
+    DATABASE_URL: z
+      .string()
+      .url()
+      .refine(
+        (str) => !str.includes("YOUR_MYSQL_URL_HERE"),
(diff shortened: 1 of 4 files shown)


cli/template/extras/src/env/with-auth-db.mjs
juliusmarminge, line 1: maybe we should split this up into 2 files now (prisma + drizzle) since for prisma the default is sqlite which works OOTB. is that worth doing or just keep it like this?

c-ehrlich: i think it's fine like this for now

c-ehrlich (approved):

cli/template/extras/src/env/with-auth-db.mjs
juliusmarminge: yea no need to over-optimize, people might want to adopt this pattern even if they use prisma

c-ehrlich:
oh it seems like some combination of ci matrix options are failing, probably drizzle because it includes the default mysql url?

PR #1576, opened 2023-09-25
feat(docs): start any example in stackblitz

mqshaikh8:
Closes #<issue>

## ✅ Checklist

- [✅] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [✅] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [✅] I performed a functional test on my final commit

---

## Changelog

I added an example page that lets users pick the different options and see how it would work on stackblitz.

---

## Screenshots

_[Screenshots]_
Screen Shot 2023-09-25 at 8 17 19 AM

💯


www/src/components/docs/exampleOptionForm.astro
```diff
@@ -0,0 +1,91 @@
...
+  class="mb-4 w-full rounded border border-t3-purple-300 p-4 text-lg"
+>
+  {
+    Object.entries(options).map(([code, name]) => (
+      <div>
+        <input
+          id={code}
+          type="checkbox"
+          name={name}
+          checked
```
c-ehrlich, line 78: since drizzle + prisma isn't a valid combination, could you make it so only drizzle is checked? dont care if its a condition on the checked prop or the loop is unwrapped.

c-ehrlich: could maybe even add validation to the form so prisma and drizzle cant be selected together, but doesn't have to be part of this pr

www/src/components/docs/exampleOptionForm.astro
```diff
@@ -0,0 +1,91 @@
...
+          title: "T3 Examples",
+          description: "T3 example apps.",
+          template: "node",
+          files: {
+            "package.json": `{
+                    "name": "T3 Examples",
+                    "version": "1.0.0",
+                    "description": "My Next.js app",
+                    "scripts":{
+                        "dev": " echo 'y' | npx create-t3-app example-app --CI ${checkedInputs} && cd ./my-app  && rm ../package.json " }
```
c-ehrlich, line 27: it's currently scaffolding in `./example-app` and then tries to cd to `./my-app`. i think out of the two i prefer `./example-app`

pnpm-lock.yaml
```diff
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
 
+settings:
```
c-ehrlich, line 3: what are all of these changes to the lockfile? is it just that pnpm changed how some stuff works in a recent version?

c-ehrlich (changes requested):
thanks for the pr. this is a really clever way of implementing the demos :)

www/src/components/docs/exampleOptionForm.astro
mqshaikh8: Hi,
Sure, makes sense. I will keep one in this pr and then maybe make another pr where I add the other one with the conditional logic

pnpm-lock.yaml
mqshaikh8: This happened when I installed the stackblitz sdk using pnpm

www/src/components/docs/exampleOptionForm.astro
mqshaikh8: Yeah, 😅 That was a typo on my part. I'll get that fixed

mqshaikh8:
Hi,
It was a straightforward fix so I just added it in here. I also added a callout just to make sure the prisma and drizzle check switching is understood.

pnpm-lock.yaml
juliusmarminge: Can you make sure you're using pnpm@8.5.1 (as specified in package.json, then run

```
git checkout next pnpm-lock.yaml
pnpm i
```

That should remove any unnecessary diffs

www/src/pages/en/examples.mdx
juliusmarminge, line 1: Some info on this page that they must enter some environment variables in .env when their project is running if they want it to fully work / callout "some features might not work unless you create an env file" etc? Not sure what the best approach is

mqshaikh8:
@juliusmarminge  I decided to add another callout to cover that.  I also am not able to get rod of the pnpm changes. I downgraded to 8.5.

mqshaikh8:
@c-ehrlich @juliusmarminge  Let me know if there are other changes I should do.

juliusmarminge (approved):
This is awesome. It's not starting up the example automatically for me though:

CleanShot 2023-10-06 at 10 48 42@2x

If you wanna look into that as a follow-up PR that would be great, but no pressure as this is perfectly fine!

Issue #1600, opened 2023-10-16
bug:  pnpm dev not working in older versions of Node 18


williamneves:
### Provide environment information

System:
    OS: macOS 13.4.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 201.78 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 19.6.0 - ~/.nvm/versions/node/v19.6.0/bin/node
    npm: 9.4.0 - ~/.nvm/versions/node/v19.6.0/bin/npm
    pnpm: 8.8.0 - ~/.nvm/versions/node/v19.6.0/bin/pnpm

### Describe the bug

Just is not loading at the first installation. I tried ran pnpm create t3-app and use the AppRoute, but after all installation (with works well), when I open the localhost:3000, the error pops:

```
Unhandled Runtime Error
Error: Missing result

Call Stack
TRPCClientError.from
node_modules/.pnpm/@trpc+client@10.40.0_@trpc+server@10.40.0/node_modules/@trpc/client/dist/TRPCClientError-0de4d231.mjs (37:15)
from
node_modules/.pnpm/@trpc+client@10.40.0_@trpc+server@10.40.0/node_modules/@trpc/client/dist/httpBatchLink-204206a5.mjs (200:55)
process.processTicksAndRejections
node:internal/process/task_queues (95:5)
```

### Reproduction repo

don't needed

### To reproduce

Just 
`pnpm create t3-app`

Options Selected:
- Auth: none,
- TS
- TW
- App Route
- ~/

`pnpm dev`

luz33c:
when you run: pnpm build 
![image](https://github.com/t3-oss/create-t3-app/assets/62172559/aad9de57-58e1-4246-a60a-b22a39e3141a)

kwerenta:
Downgrading to Next.js 13.5.4 fixes "Missing result" error

c-ehrlich:
![CleanShot 2023-10-17 at 21 17 00@2x](https://github.com/t3-oss/create-t3-app/assets/8353666/b51b786a-697a-4198-aba3-1adfc2e85bba)

could you give more detailed steps or provide a repo from which this can be reproduced? it's working for me, including with Next 13.5.5

pablasso:
@c-ehrlich hopefully you're able to reproduce with this repo: https://github.com/pablasso/reproduce-no-results, the same thing was happening with me with npm and latest next.

[...]

PR #1608, opened 2023-10-18
feat: check node version in next config

c-ehrlich:
## ✅ Checklist

---

would stop people from running into https://github.com/t3-oss/create-t3-app/issues/1600#issuecomment-1769189236

💯

### DIFF
--- a/cli/template/base/next.config.mjs
+++ b/cli/template/base/next.config.mjs
@@ -4,6 +4,8 @@
  */
 await import("./src/env.mjs");
 
+checkNodeVersion();
+
 /** @type {import("next").NextConfig} */
 const config = {
   reactStrictMode: true,
@@ -20,3 +22,14 @@ const config = {
 };
 
 export default config;
+
+function checkNodeVersion() {
+  const nodeVersion = process.versions.node;
+  const [major, minor] = nodeVersion.split(".").map((n) => parseInt(n));
+  if (!major || !minor) {
+    throw new Error(`Unable to parse Node version: ${nodeVersion}`);
+  }
+  if (major < 18 || (major === 18 && minor < 6) || (major === 19 && minor < 7)) {
+    throw new Error(`You are running an outdated Node version. Please use at least 18.16, 19.7, or 20.0`);
+  }
+}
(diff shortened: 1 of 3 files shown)


juliusmarminge:
I think 18.13 should be fine, if it's the same bug we had for uploadthing with the fetch api

cli/template/base/next.config.mjs
```diff
@@ -20,3 +22,14 @@ const config = {
...
 
 export default config;
+
+function checkNodeVersion() {
+  const nodeVersion = process.versions.node;
+  const [major, minor] = nodeVersion.split(".").map((n) => parseInt(n));
+  if (!major || !minor) {
+    throw new Error(`Unable to parse Node version: ${nodeVersion}`);
+  }
+  if (major < 18 || (major === 18 && minor < 6) || (major === 19 && minor < 7)) {
```
sscotth, line 32: should this be `minor < 16`?

juliusmarminge: 16 is EOL

sscotth: I think you misread `minor`, his check is "<18.6" not "<18.16"

c-ehrlich: yes, thanks :)

c-ehrlich:
> I think 18.13 should be fine, if it's the same bug we had for uploadthing with the fetch api
> 
> could we not just set the engines field in package json instead?

i dont think its possible to specify multiple versions? 20 is becoming LTR in a few days, so we could just require that, but would probably mean we also need to update docker docs etc

juliusmarminge:
> > I think 18.13 should be fine, if it's the same bug we had for uploadthing with the fetch api
> > could we not just set the engines field in package json instead?
> 
> i dont think its possible to specify multiple versions? 20 is becoming LTR in a few days, so we could just require that, but would probably mean we also need to update docker docs etc

```json
"engines": {
  "node": "^18.16 || ^19.7 || >=20"
}
```

Doesnt' that work?

sscotth:
Next just bumped their Node version to `>=18.17.0` in the v14.5.6 canary

https://github.com/vercel/next.js/pull/56896
https://github.com/vercel/next.js/pull/56943

PR #1612, opened 2023-10-20
feat: Add Dropdown for Installation Commands on Homepage

Jacksonmills:
Closes [#1609](https://github.com/t3-oss/create-t3-app/issues/1609)

## ✅ Checklist

- [x] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [x] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] I performed a functional test on my final commit

---

## Changelog

Add dropdown for different package manager copy to clipboard options on homepage

---

## Screenshots

image
image

🫡💖

PR #1621, opened 2023-10-24
fix: replace all incorrect versions

### LINKED ISSUE #1620: bug: create.t3.gg npx link is incorrect

live version of https://create.t3.gg/

### Describe the bug

throughout https://create.t3.gg/ the npx link is incorrect, it shows npx create t3-app@latest instead of npx create t3-app@latest
The - between create and t3 is missing

### Reproduction repo

https://github.com/t3-oss

### To reproduce

go to https://create.t3.gg/ and look at the main CTA and the way it is worded

### Additional information

_No response_

ed-wright:
goes through the repo and replaces all incorrect versions of create t3-app with  create-t3-app This closes 1620

Closes #1620 

## ✅ Checklist

- [✅] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [✅ ] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [✅ ] I performed a functional test on my final commit

---

## Changelog

All incorrect versions of the npx command have been replaced

---

## Screenshots
Old
image
New
image

💯

### DIFF
--- a/www/src/components/landingPage/banner.astro
+++ b/www/src/components/landingPage/banner.astro
@@ -73,7 +73,7 @@ import ClipboardSelect from "./ClipboardSelect";
                   <div
                     class="relative flex items-center rounded-lg border border-t3-purple-200/20 bg-t3-purple-100/10 px-2 py-2 text-sm md:px-3 md:py-3 md:text-lg lg:px-5 lg:py-4 lg:text-xl"
                   >
-                    <code class="mr-12">npx create t3-app@latest</code>
+                    <code class="mr-12">npx create-t3-app@latest</code>
                     <ClipboardSelect client:load />
                   </div>
                 </div>
--- a/.changeset/tricky-suns-float.md
+++ b/.changeset/tricky-suns-float.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+Updated the NPX links to fix them
--- a/.github/workflows/prerelease-comment.yml
+++ b/.github/workflows/prerelease-comment.yml
@@ -48,7 +48,7 @@ jobs:
             A new create-t3-app prerelease is available for testing. You can install this latest build in your project with:
 
             ```sh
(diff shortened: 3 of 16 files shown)


ronanru:
You fixed one wrong install command, but broke all others.
Correct install commands are
- `npm create t3-app@latest`
- `yarn create t3-app@latest`
- `pnpm create t3-app@latest`
- `buun create t3-app@latest`

Also, you don't need a changeset for that kind of change

www/src/components/landingPage/banner.astro
```diff
@@ -73,7 +73,7 @@ import ClipboardSelect from "./ClipboardSelect";
                   <div
                     class="relative flex items-center rounded-lg border border-t3-purple-200/20 bg-t3-purple-100/10 px-2 py-2 text-sm md:px-3 md:py-3 md:text-lg lg:px-5 lg:py-4 lg:text-xl"
                   >
-                    <code class="mr-12">npx create t3-app@latest</code>
+                    <code class="mr-12">npx create-t3-app@latest</code>
```
c-ehrlich, line 76: This one is indeed incorrect - seems to have been missed when #1612 was reviewed. but personally I think it would be better to stay consistent and use `npm create` here, as it's what we're doing everywhere else.

```suggestion
                    <code class="mr-12">npm create t3-app@latest</code>
```

c-ehrlich (changes requested):
Thanks for the pr. The command in `banner.astro` is indeed incorrect. But all the other commands are correct as they are, and the way you changed them (`npm create-t3-app` etc) are not valid commands.

`npm create`, and the equivalent yarn/pnpm/bunx commands) are a real thing, see: https://docs.npmjs.com/cli/v9/commands/npm-init

Could you change the PR to only fix the one on the landing page, and revert the others?

c-ehrlich:
superseded by https://github.com/t3-oss/create-t3-app/pull/1624

PR #1622, opened 2023-10-24
feat: Add more db options

### LINKED ISSUE #1617: feat: Add neon || planetscale to relation database options (postgres || mysql)

I think that for the cases that one has a preference between postgres or mysql we can choose which way to go.

Add an option within the command cli in which we can choose whether to use neon serverless || planetscale serverless

Poder escoger si uno prefiere utilizar quizás postgres en cambio de mysql y de esa manera deducir si por default dejar neon || planetscale

### Additional information

_No response_

### LINKED ISSUE #1646: feat: Prisma Edge Client

Prisma 5.4.0 added support for driver adapters to use serverless adapters over HTTP (and WS, for neon). This completes support for the Edge via PlanetScale using Prisma, previously only possible using Drizzle. Note that this also does NOT require Prisma's "Accelerate" service to use, from what I have read.

This could work really well in concert with https://github.com/t3-oss/create-t3-app/pull/1622, as selecting PlanetScale Serverless as an option would work nicely both on Prisma and on Drizzle.

Add a separate DB client (or just use the edge client) for edge environments:

```ts
// Import required dependencies
import { connect } from '@planetscale/database';
import { PrismaPlanetScale } from '@prisma/adapter-planetscale';
import { PrismaClient } from '@prisma/client/edge';
import { env } from "~/env.mjs";

// Setup global primitives
const globalForPrisma = globalThis as unknown as {
  prisma: PrismaClient | undefined;
};

// Initialize Prisma Client with the PlanetScale serverless database driver
const connection = connect({ url: env.DATABASE_URL });
const adapter = new PrismaPlanetScale(connection);

export const db =
  globalForPrisma.prisma ??
  new PrismaClient({
    log: env.NODE_ENV === "development" ? ["query", "error", "warn"] : ["error"],
    adapter,
  });

if (env.NODE_ENV !== "production") globalForPrisma.prisma = db;
```

Required dependencies:
`prisma` and `@prisma/client`: 5.4.0 or higher (latest as of writing is 5.5.2)
`@prisma/adapter-planetscale`: Same version as Prisma
`@planetscale/database`: Same version as used in current deployments with PlanetScale.

This also requires the `driverAdapters` feature flag to be enabled. Since it is technically not generally availaible, this combination of options might be best marked as "experimental" in the CLI.

I also am not entirely sure about whether the edge client requires the weird `global` manipulation that the normal client does, since it operates differently (and in theory, connection pooling that causes the normal one to be problematic isn't present on the edge variant), but that requires testing that I have not yet performed.

N/A

### Additional information

_No response_

ronanru:
Closes #1617
Closes #1646

## ✅ Checklist

---

## Changelog

Add a database selector, if user chose an ORM.

---

## Screenshots

image

💯


cli/src/installers/drizzle.ts
```diff
@@ -5,20 +5,39 @@ import { type PackageJson } from "type-fest";
...
-    dependencies: ["drizzle-orm", "@planetscale/database"],
+    dependencies: [
+      "drizzle-orm",
+      (
+        {
+          planetscale: "@planetscale/database",
+          mysql: "mysql2",
+          postgres: "pg",
+          sqlite: "better-sqlite3",
+          neon: "@neondatabase/serverless",
```
juliusmarminge, line 37: i don't think we want neon

cli/src/installers/drizzle.ts
```diff
@@ -42,10 +61,85 @@ export const drizzleInstaller: Installer = ({
...
+      schemaContent = schemaContent.replace("  bigint,\n", "");
+      schemaContent = schemaContent.replace(
+        "  int,\n",
+        "  integer as int,\n  serial,\n"
+      );
+      schemaContent = schemaContent.replace(
+        'id: bigint("id", { mode: "number" }).primaryKey().autoincrement(),',
+        'id: serial("post_id").primaryKey(),'
+      );
+      schemaContent = schemaContent.replace("fsp: ", "precision: ");
```
juliusmarminge, line 135: all this looks prone to error at the first change we make. just having them as separate files will be easier to maintain

cli/template/extras/src/server/db/index-drizzle/with-postgres.ts
```diff
@@ -0,0 +1,12 @@
+import { drizzle } from "drizzle-orm/node-postgres";
+import { Client } from "pg";
```
juliusmarminge, line 2: which one is more popular here? i've personally used postgres.js (postgres) more than node-postgres (pg)

juliusmarminge (changes requested):
I like the idea here but there are stuff we need to make sure before merging this.

What's the dev story? Should we include a db container? Recommend something like [DBngin](https://dbngin.com/)? How should the developer get the database running as quickly as possible in dev, without drifting away from the production environment?

---

And as I previously commented, we are not adding neon.

ronanru:
> What's the dev story? Should we include a db container? 
I think adding a script that creates a db container is a great idea. I will add it in a future commit.

cli/template/extras/src/server/db/index-drizzle/with-postgres.ts
ronanru: ps is more popular, but postgres.js is smaller and faster, I will switch to it here.

ronanru:
> Hi @ronanru , super cool idea! Do you need help on this?
> 
> FYI `neon` added a CLI, which should enable programatically creating a project/database/URL.

Thank you. This PR is done and I don't need help. 

I removed neon from this PR, Julius said Planetscale is the only DB provider T3 maintainers recommend, and they can't add stuff they don't recommend to create-t3-app.

juliusmarminge:
Sorry for the delayed review, I'll try looking over this more closely today. We might want to add #1646 as a Prisma + PlanetScale option?

cli/src/installers/envVars.ts
```diff
@@ -55,16 +72,29 @@ const getEnvContent = (
...
 # Change the query params at the end of the URL to "?ssl={"rejectUnauthorized":true}"
-DATABASE_URL='mysql://YOUR_MYSQL_URL_HERE?ssl={"rejectUnauthorized":true}'
-`;
+DATABASE_URL='mysql://YOUR_MYSQL_URL_HERE?ssl={"rejectUnauthorized":true}'`;
+    } else if (databaseProvider === "mysql") {
+      content += `DATABASE_URL="mysql://root:${generateRandomString(
+        12
+      )}@localhost:3306/${projectName}"`;
+    } else if (databaseProvider === "postgres") {
+      content += `DATABASE_URL="postgresql://postgres:${generateRandomString(
```
juliusmarminge, line 89: this seems a bit aggressive to generate a password on their behalf?

juliusmarminge (changes requested):
- Drizzle Schemas should probably have foreign keys for all configurations except PlanetScale
- Unsure about the db container. I know I raised this issue but I don't know what the best solution is. I can only talk from personal experience but I've stopped using docker for dbs nowadays and I think everyone has their own preference. We should make sure to document this in the first step and perhaps list some options 🤔

ronanru:
> Unsure about the db container. I know I raised this issue but I don't know what the best solution is. I can only talk from personal experience but I've stopped using docker for dbs nowadays and I think everyone has their own preference. We should make sure to document this in the first step and perhaps list some options 🤔

Docker is the fastest way to start a dev database I know that works on all major operating systems. Also, it's just one small file, and you can just delete it if you don't need it. I added a paragraph to the docs about it and mentioned DBngin as an alternative.

cli/template/extras/prisma/schema/base-planetscale.prisma
```diff
@@ -0,0 +1,21 @@
+// This is your Prisma schema file,
+// learn more about it in the docs: https://pris.ly/d/prisma-schema
+
+generator client {
+    provider        = "prisma-client-js"
+    previewFeatures = ["driverAdapters"]
+}
+
+datasource db {
+    provider = "sqlite"
```
Wundero, line 10: Should this not be `mysql` since this is a planetscale prisma schema?

Also probably needs `relationMode = "prisma"` in this section too.

cli/template/extras/prisma/schema/with-auth-planetscale.prisma
```diff
@@ -0,0 +1,74 @@
...
+    previewFeatures = ["driverAdapters"]
+}
+
+datasource db {
+    provider = "sqlite"
+    // NOTE: When using mysql or sqlserver, uncomment the @db.Text annotations in model Account below
+    // Further reading:
+    // https://next-auth.js.org/adapters/prisma#create-the-prisma-schema
+    // https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#string
+    url      = env("DATABASE_URL")
```
Wundero, line 15: Likewise.

cli/template/extras/src/server/db/db-prisma-planetscale.ts
```diff
@@ -0,0 +1,21 @@
+import { Client } from "@planetscale/database";
+import { PrismaPlanetScale } from "@prisma/adapter-planetscale";
+import { PrismaClient } from "@prisma/client/edge";
```
Wundero, line 3: Currently, the edge client doesn't support custom adapters (see https://github.com/prisma/prisma/issues/21394), so this would need to be the normal prisma client until that issue is resolved. Once resolved, this can change (and likely also remove some of the global fluff since connection pooling is technically unnecessary when using the edge stuff).

ronanru: Didn't know that, thank you!

juliusmarminge:
Published a beta: 

```sh
bun create t3-app@7.25.0-beta.1
```

juliusmarminge:
> Also, it's just one small file, and you can just delete it if you don't need it

This file is not windows compatible though right?

ronanru:
> > Also, it's just one small file, and you can just delete it if you don't need it
> 
> This file is not windows compatible though right?

It works with WSL. I added the instructions on how to run it on Windows.

juliusmarminge:
So sorry for the delay, but today I tested a lot of this out and I have to say awesome work!

However, using Bun, Drizzle and SQLite doesn't quite work, also Bun has it's own sqlite implementation which supposedly is faster etc. Should we use that? https://orm.drizzle.team/docs/get-started-sqlite#bun-sqlite

CleanShot 2024-01-04 at 16 26 44@2x

cli/template/extras/src/server/auth-app/with-drizzle.ts
```diff
@@ -46,7 +47,7 @@ export const authOptions: NextAuthOptions = {
       },
     }),
   },
-  adapter: DrizzleAdapter(db, createTable),
+  adapter: DrizzleAdapter(db, createTable) as Adapter,
```
juliusmarminge, line 50: ref #1716

juliusmarminge:
the DATABASE_URL for prisma+planetscale also doesn't work, you need to change the ssl query param:

```diff
- DATABASE_URL='mysql://user:pw@aws.connect.psdb.cloud/db?ssl={"rejectUnauthorized":true}'
+ DATABASE_URL='mysql://user:pw@aws.connect.psdb.cloud/db?sslaccept=strict' 
```

juliusmarminge:
Current published version is `7.26.0-beta.4`

juliusmarminge:
> So sorry for the delay, but today I tested a lot of this out and I have to say awesome work!
> 
> However, using Bun, Drizzle and SQLite doesn't quite work, also Bun has it's own sqlite implementation which supposedly is faster etc. Should we use that? https://orm.drizzle.team/docs/get-started-sqlite#bun-sqlite
> 
> EDIT: hmm doesn't look like db push works for that :/
> 
> CleanShot 2024-01-04 at 16 26 44@2x

I'm fine just not suppporting that combo to begin with... Gonna merge this

juliusmarminge (approved):

Issue #1623, opened 2023-10-25
bug (ci): changeset should only be required on changes to cli and outputted app 


c-ehrlich:
### Describe the bug

Currently the CI throws an error on any PR that doesn't include a changeset. However the intended behaviour is that changesets should only be required on changes to the CLI or scaffolded app, ie anything in `/cli` (or maybe even anything in `/cli` except readmes etc)

Jacksonmills:
I tried a fix for this, is there a good way to test the ci workflow locally?

PR #1625, opened 2023-10-25
ci: :zap: Add /cli change validation to CI workflow

Jacksonmills:
Should allow for changes to www and docs to be free from ci errors around changeset, I have little exp working with the ci but thought I would try my attempt at a fix!

Closes [#1623](https://github.com/t3-oss/create-t3-app/issues/1623)

## ✅ Checklist

- [x] I pushed a series of commits specifically designed to trigger various edge cases in the GitHub CI workflow, ensuring comprehensive functional testing of the entire process.

---

## Changelog

Added a new step in the CI workflow to check for changes in the /cli directory. If there are any changes, it will validate the changeset by running `pnpm changeset status --since origin/next`. This ensures that changes made to /cli are properly tracked and validated.

💯⚡

### DIFF
--- a/.changeset/clean-flowers-dress.md
+++ b/.changeset/clean-flowers-dress.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+asd
--- a/.changeset/clean-flowers-dress.md
+++ b/.changeset/clean-flowers-dress.md
@@ -1,5 +1,0 @@
----
-"create-t3-app": patch
----
-
-asd
--- a/.changeset/hip-ligers-yell.md
+++ b/.changeset/hip-ligers-yell.md
@@ -1,5 +1,0 @@
----
-"create-t3-app": patch
----
-
-debug comment
--- a/.changeset/hip-ligers-yell.md
+++ b/.changeset/hip-ligers-yell.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+debug comment
--- a/.changeset/modern-fireants-care.md
+++ b/.changeset/modern-fireants-care.md
@@ -1,5 +1,0 @@
----
-"create-t3-app": patch
----
-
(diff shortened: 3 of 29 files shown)


Jacksonmills:
image

noice

c-ehrlich:
thanks for picking this up! the current version seems to hit false negatives, see https://github.com/t3-oss/create-t3-app/pull/1626

also @juliusmarminge do we also need to include a few files in root such as eslintrc, prettier config, etc? i'm leaning no, because none of them should change the ux of using the cli

Jacksonmills:
> thanks for picking this up! the current version seems to hit false negatives, see https://github.com/t3-oss/create-t3-app/pull/1626
> 
> also @juliusmarminge do we also need to include a few files in root such as eslintrc, prettier config, etc? i'm leaning no, because none of them should change the ux of using the cli

Ah okay thanks for adding the test PR, I'll see what's going on with that

Jacksonmills:
> thanks for picking this up! the current version seems to hit false negatives, see #1626
> 
> also @juliusmarminge do we also need to include a few files in root such as eslintrc, prettier config, etc? i'm leaning no, because none of them should change the ux of using the cli

@c-ehrlich I think I got it working if you would like to pull my changes and test for yourself. I was able to fail and pass with and without a change in cli/src/index.ts

Jacksonmills:
Okay in my testing it fails when no changeset + change in `ci/src/index.ts` and passes when you add changeset or remove the change from `ci/src/index.ts`

Jacksonmills:
Gonna come back to this tomorrow ...

Jacksonmills:
Running all the tests ... but it seems the grep command was the culprit, needed to return true if grep doesn't find any matches, allowing the script to proceed to the else block gracefully.

c-ehrlich (approved):
working great now. thanks so much for adding this!

PR #1668, opened 2023-11-18
feat: automatic env vars for updated drizzle-kit

juliusmarminge:
waiting for updates from drizzle team regarding some weird results from db:push on kit 0.20

PR #1667, opened 2023-11-18
chore: bump some deps

juliusmarminge:
Closes #<issue>

## ✅ Checklist

- [ ] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [ ] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] I performed a functional test on my final commit

---

## Changelog

_[Short description of what has changed]_

---

## Screenshots

_[Screenshots]_

💯

PR #1666, opened 2023-11-18
refactor: use `type: module`

juliusmarminge:
imo every project in 2023 should use `type: module`

Soon we will have an ESM `eslint.config.js` file and then it's just postcss left in cjs land...

PR #1670, opened 2023-11-19
feat: skip http for trpc rsc calls

### LINKED ISSUE #1599: bug:   7.22.0 next build error TRPCClientError: Dynamic server usage

 "ct3aMetadata": {
    "initVersion": "7.22.0"
  },

### Describe the bug

image

### Reproduction repo
.

### To reproduce

when：
- install： `pnpm create t3-app@latest`
- use nextjs  appdir option 
- build：` pnpm build`

### Additional information

_No response_

### LINKED ISSUE #1639: bug: tRPC error boilerplate code deploying to vercel breaks site

  System:
    OS: macOS 14.0
    CPU: (8) arm64 Apple M2
    Memory: 80.75 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.8.0 - /opt/homebrew/bin/node
    npm: 10.2.0 - /opt/homebrew/bin/npm
    pnpm: 8.1.1 - /opt/homebrew/bin/pnpm

### Describe the bug

Used bun create t3-app@latest after upgrading to NextJS 14. I ran it locally and it all worked fine. I then created a repo and tried to deploy to Vercel but I got a build error related to tRPC and an error displayed in the window (Screenshots attached). I then tested this using pnpm and npm and same result. App without tRPC deployed fine. 
Screenshot 2023-10-31 at 4 52 16 PM
image

### Reproduction repo

https://github.com/prestonButton/bbqapp

### To reproduce

This app is just the create-t3-app boilerplate code. I ran bun db:push, bun run dev, and then tried to deploy. and got the errors described above. Just wondering if this is because I am using NextJS 14.

### Additional information

_No response_

### LINKED ISSUE #1650: bug: Invariant: headers() expect to have requestAsyncStorage, none available.

  System:
    OS: macOS 13.6.1
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 217.77 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.17.0 - ~/.nvm/versions/node/v16.17.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.17.0/bin/yarn
    npm: 9.2.0 - ~/.nvm/versions/node/v16.17.0/bin/npm

### Describe the bug

When trying to test posts' hello route via createInnerTRPCContext function, the following error occurs:
````

> test-t3-app-router@0.1.0 test
> jest --ci

 FAIL  src/server/api/routers/post.test.ts
  Test post endpoint
    ✕ Test hello (2 ms)

  ● Test post endpoint › Test hello

    Invariant: headers() expects to have requestAsyncStorage, none available.

      10 |   it("Test hello", async () => {
      11 |     const ctx = await createInnerTRPCContext({
    > 12 |       headers: headers(),
         |                       ^
      13 |     });
      14 |     const caller = appRouter.createCaller(ctx);
      15 |

      at headers (node_modules/next/src/client/components/headers.ts:22:11)
      at Object.<anonymous> (src/server/api/routers/post.test.ts:12:23)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        4.623 s
Ran all test suites.
````

I don't believe the documentation is up to date for [tRPC - Sample Integration Test](https://create.t3.gg/en/usage/trpc#sample-integration-test) as I've tried to use headers() function from "next/headers" similar to trpc/server.ts api function.

### Reproduction repo

https://github.com/stubinnubin/t3-stack-header-bug.git

### To reproduce

Clone repository, then run `npm run test`. 

### Additional information

_No response_

### LINKED ISSUE #1656: bug: Recent Commit Breaks Clerk Based Authentication w/ tRPC

  "initVersion": "7.23.2"
  
  System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (16) x64 Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz
    Memory: 9.02 GB / 15.58 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 18.18.0 - ~/.nvm/versions/node/v18.18.0/bin/node
    Yarn: 1.22.17 - ~/.yarn/bin/yarn
    npm: 10.1.0 - ~/.nvm/versions/node/v18.18.0/bin/npm
    pnpm: 8.9.2 - ~/.local/share/pnpm/pnpm
    bun: 1.0.7 - ~/.bun/bin/bun

### Describe the bug
I've encountered an issue with the recent changes introduced in commit [e6649f18963086b2d016815e60cc9915711a85d8](https://github.com/t3-oss/create-t3-app/commit/e6649f18963086b2d016815e60cc9915711a85d8). These changes appear to have broken the Clerk based authentication, which in turn is affecting the functionality of tRPC in my application.

### Reproduction repo
https://github.com/Jacksonmills/gpts-browser

### To reproduce
To reproduce, revert the changes back to main in this commit [e6649f18963086b2d016815e60cc9915711a85d8](https://github.com/t3-oss/create-t3-app/commit/e6649f18963086b2d016815e60cc9915711a85d8). Once reverted, log in with Clerk and visit a page with a query to get:
```
Unhandled Runtime Error
Error: Cannot read properties of undefined (reading 'resolve')
```
image

### Additional information
I was able to revert the changes in the problematic commit, which fixes all my issues. However, I would love if someone could further explain or suggest how we can go back to using cookies while also making Clerk work. That would be ideal! Thank you. 👏

### LINKED ISSUE #1669: feat: Use tRPC's createCaller API for server-side api requests

Currently, ct3a uses an HTTP proxy layer to make requests on the server, which is a bit more abstraction than should really be necessary. tRPC exposes a `createCaller` function on its routers, which can be used to directly call api functions instead of routing via the proxy. This should be used in place of the HTTP proxy where possible.

The file `trpc/server.ts` should contain code that looks something like the following:

```ts
import { headers, cookies } from "next/headers";

import { appRouter } from "~/server/api/root";
import { createTRPCContext } from "~/server/api/trpc";

export const api = appRouter.createCaller(
  await createTRPCContext({
    req: {
      headers: headers(),
      cookies: cookies(),
    },
  }),
);
```

One caveat here is that I have run into issues trying to get this to properly build or run. If you use `await`, it runs into issues that seem to be causing infinite loops. When not using `await`, it complains that `headers` or `cookies`, doesn't really matter which, does not have an async storage available. I am not sure how to fix either of these issues, 

With `await`, during `collecting page data` phase of build:
image

Without `await`:
image

To fix this in my environment, I use a hack involving Proxy:
```ts
import { headers } from "next/headers";

import { appRouter } from "~/server/api/root";
import { createTRPCContext } from "~/server/api/trpc";

type Caller = ReturnType<typeof appRouter.createCaller>;

let caller: Caller | null = null;

// Cache the caller, since this runs per-request anyways.
const getCaller = () => {
  // This fn should be async if create context is async
  if (!caller) {
    caller = appRouter.createCaller(
      createTRPCContext({
        // Await this if async
        req: {
          headers: headers(),
        },
      }),
    );
  }
  return caller;
};

// @ts-expect-error - this is a hack to get the implementation to work async
const routeOrCallProxy = (path: (string | symbol)[]) => {
  // eslint-disable-next-line @typescript-eslint/ban-types
  return new Proxy<Function>( // We use a fn type because it tricks webpack into letting this actually run
    // eslint-disable-next-line @typescript-eslint/no-empty-function
    getCaller, // Any function will do, since we override apply
    {
      get: (_, prop) => {
        if (caller) {
          // Route to deeper path in the api caller, for example api.edge.user.get().
          //  This is unlikely to be called if caller exists, but just in case.
          let val: unknown = caller;
          for (const p of path) {
            // @ts-expect-error - this is a hack to get the types to work
            val = val[p];
          }
          // @ts-expect-error - this is a hack to get the types to work
          // eslint-disable-next-line @typescript-eslint/no-unsafe-return
          return val[prop];
        }
        return routeOrCallProxy([...path, prop]) as unknown;
      },
      // Note: this is where you would insert async if the context was async
      apply: (_, __, args) => {
        // Function call should retrieve actual result, so we ensure that the caller exists at this point.
        const caller = getCaller(); // Await this if ctx async
        let val: unknown = caller;
        for (const p of path) {
          // @ts-expect-error - this is a hack to get the types to work
          val = val[p];
        }
        // @ts-expect-error - this is a hack to get the types to work
        // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-argument
        return val(...args);
      },
    },
  );
};

export const api = new Proxy<Caller>({} as Caller, {
  get: (_, prop) => {
    if (caller) {
      return caller[prop as keyof Caller];
    }
    return routeOrCallProxy([prop]) as unknown;
  },
});

```
This still has all the type-safety you expect, but lazily instantiates the API so that it doesn't run on import. For the sync version, this has a lot of complexity that in theory could be reduced, but the value of making it work and supporting both sync + async is that users trying to convert to async won't run into as many issues. I haven't optimized the performance of this, and it probably could be made quite a bit faster if I had to guess, but it works. I'll also mention that we don't await the `getCaller` fn in the `api` proxy because then it errors when trying to get to router functions.

N/A, could just keep the current proxy if things don't work out with the caller style.

### Additional information

I'm not a particularly big fan of using the proxy method, so any suggestions are welcome. I plan to make a PR for this at some point, but I'd like to gather some feedback first since this is a bit sketchy.

juliusmarminge:
Closes #1669
Closes #1656
Closes #1650
Closes #1639
Closes #1599

Adds a custom link that just invokes the procedures directly with a cached createContext function, instead of using http even for RSC which is unnecessary


juliusmarminge:
Current dependencies on/for this PR:
* main
  * **PR #1666** <a href="https://app.graphite.dev/github/pr/t3-oss/create-t3-app/1666?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #1667** <a href="https://app.graphite.dev/github/pr/t3-oss/create-t3-app/1667?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #1670** <a href="https://app.graphite.dev/github/pr/t3-oss/create-t3-app/1670?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
      * **PR #1668** <a href="https://app.graphite.dev/github/pr/t3-oss/create-t3-app/1668?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/t3-oss/create-t3-app/1670?utm_source=stack-comment).

cli/template/extras/src/trpc/server.ts
```diff
@@ -1,29 +1,60 @@
 import {
```
juliusmarminge, line 1: should we `import "server-only"` here? I think it's the technically right thing to do so you can't import and attempt to use this in CCs

cli/template/base/_eslintrc.cjs
```diff
@@ -24,6 +24,7 @@ const config = {
       },
     ],
     "@typescript-eslint/no-unused-vars": ["warn", { argsIgnorePattern: "^_" }],
+    "@typescript-eslint/require-await": "off",
```
juliusmarminge, line 27: added this cause this rule is annoying in RSC since for example server actions have to be async, leading to a lot of `disable`'s for this rule spread around

juliusmarminge:
```sh
pnpm create t3-app@7.23.3-beta.64b92d
```

cli/template/extras/src/trpc/server.ts
```diff
@@ -1,29 +1,64 @@
...
+import { appRouter } from "~/server/api/root";
+import { createTRPCContext } from "~/server/api/trpc";
+import { transformer } from "./shared";
 
-export const api = createTRPCProxyClient<AppRouter>({
+/**
+ * This wraps the `createTRPCContext` helper and provides the required context for the tRPC API when
+ * handling a tRPC call from a React Server Component.
+ */
+const createContext = cache(() => {
```
eula01, line 20: doesn't this need to be async, like in `/api/trpc/[trpc]/route.ts` ?

juliusmarminge: createTRPCContext returns a Promise so it is async

cli/template/extras/src/trpc/server.ts
c-ehrlich: yes seems reasonable

cli/template/extras/src/trpc/server.ts
```diff
@@ -1,29 +1,64 @@
...
-        return {
-          cookie: cookies().toString(),
-          "x-trpc-source": "rsc",
-        };
-      },
-    }),
+    /**
+     * Custom RSC link that invokes procedures directly in the server component Don't be too afraid
+     * about the complexity here, it's just wrapping `callProcedure` with an observable to make it a
+     * valid ending link for tRPC.
```
c-ehrlich, line 40: i dont think that explanation will help anyone who is already intimidated by the code. maybe something that omits some technical details but just explains reason, like
```suggestion
     * Custom RSC link that lets us invoke procedures without using http requests.
     * Since Server Components always run on the server, we can just call the procedure as a function.
```

cli/template/extras/src/trpc/server.ts
c-ehrlich, line 20: i dont love that we have functions with the same `createContext` name in `server.ts` and `route.ts`. but not sure`createRSCContext` is any better.

wanted to bring it up but dont let naming bikeshedding stop us from merging :)

c-ehrlich (approved):
few nits but happy to merge as is

jakobevangelista:
this may fix the issue of clerk not working on a fresh install of the latest create t3 app, a known issue, but in my tests, it does not allow the clerk data/cookies/headers to be passed from the nextjs app to the trpc server instance in order for clerk auth functions like getAuth(opts.req) and auth() to work in trpc middleware

PR #1687, opened 2023-12-05
chore: reuse existing AppRouter type

sinasab:
Minor update to reuse existing `AppRouter` type via import.


c-ehrlich (changes requested):
Thanks for the PR. I agree with the change. Could you add a proper changeset?

sinasab:
> Thanks for the PR. I agree with the change. Could you add a proper changeset? (delete the existing changeset file, run `pnpm changeset` and follow the cli instructions)

I'm less familiar with changeset so feel free to correct me if I did anything wrong, but it should be good to go now!

Thanks for the review 😁

PR #1766, opened 2024-02-20
fix: fix too many clients error on Postgres due to HMR with Drizzle

fuegoio:
## ✅ Checklist

---

## Changelog

* Add a singleton for Postgres connection with Drizzle to avoid creating a new connection on every Hot Reload

---

With Drizzle and a Postgres connection, we always had the issue of having `too many clients already` because of the HMR (see https://www.answeroverflow.com/m/1146224610002600067).

Having a singleton to fix that seems to be the best way.

Happy to implement this logic on other databases if needed, or write some tests, but I did not know if the approach would be accepted!

Thanks for your work :)


juliusmarminge:
Should we use a similar pattern that we do for prisma to keep code consistency?

https://github.com/t3-oss/create-t3-app/blob/main/cli/template/extras/src/server/db/db-prisma.ts

juliusmarminge (changes requested):
Should we use a similar pattern that we do for prisma to keep code consistency?

https://github.com/t3-oss/create-t3-app/blob/main/cli/template/extras/src/server/db/db-prisma.ts

fuegoio:
> Should we use a similar pattern that we do for prisma to keep code consistency?
> 
> https://github.com/t3-oss/create-t3-app/blob/main/cli/template/extras/src/server/db/db-prisma.ts

You're right I did not see it sorry, changed it on all drizzle databases as well for consistency.

cli/template/extras/src/server/db/index-drizzle/with-planetscale.ts
juliusmarminge, line 1: planetscale doesn't need this as their serverless driver handles connections for you

fuegoio: Thanks removed it ;)

juliusmarminge (approved):
LGTM - asked the Drizzle team for their opinion on the change too

cli/template/extras/src/server/db/index-drizzle/with-mysql.ts
juliusmarminge, line 1: playing with the thought of putting the singleton pattern on the connection instead of the drizzle client, thoughts?

```ts
import { drizzle } from "drizzle-orm/mysql2";
import { createPool, type Pool } from "mysql2/promise";

import { env } from "~/env.js";
import * as schema from "./schema";

const globalForMySQL = globalThis as unknown as {
  pool: Pool | undefined;
};

const pool = globalForMySQL.pool ?? createPool({ uri: env.DATABASE_URL });
export const db = drizzle(pool, { schema, mode: "default" });

if (env.NODE_ENV !== "production") globalForMySQL.pool = pool;
```

fuegoio: Yes I think that could be done, if the drizzle is just a wrapper anyway that should not impact anything. But for me that adds complexity and we could also just cache the drizzle object instantiation as well.

juliusmarminge: > if the drizzle is just a wrapper anyway that should not impact anything.

yes the response i got from drizzle was that this had nothing to do with drizzle but the underlying db connection

>  But for me that adds complexity and we could also just cache the drizzle object instantiation as well.

how so? my suggestion is less LOC ;) - i just figured that for new users it'd be good to know what we're intending to cache and that is the db connection

fuegoio:
Changed it to only cache the connection instead of the whole drizzle client instance, clearer you are right ;)

juliusmarminge (approved):

juliusmarminge (approved):

PR #1802, opened 2024-03-15
fix: non planetscale env.js refers to YOUR_MYSQL_URL_HERE

### LINKED ISSUE #1795: bug: `.env.js` always validates `DATABASE_URL` for `YOUR_MYSQL_URL_HERE` even when using a different database provider

System:
OS: Windows 10 10.0.19045
CPU: (16) x64 AMD Ryzen 7 3700X 8-Core Processor
Memory: 18.68 GB / 31.94 GB
Binaries:
Node: 20.10.0 - C:\Program Files\nodejs\node.EXE
npm: 10.2.5 - C:\Program Files\nodejs\npm.CMD
pnpm: 8.14.1 - ~\AppData\Local\pnpm\pnpm.EXE

### Describe the bug

`.env.js` always validates `DATABASE_URL` for `YOUR_MYSQL_URL_HERE` even when using a different database.

`YOUR_MYSQL_URL_HERE` is used as a default value in `.env` and `.env.example` when choosing the planetscale option.

### Reproduction repo

N/A

### To reproduce

1. Initialize new t3 app
2. Choose any database ORM
3. Choose database provider
4. default `DATABASE_URL` in `.env` & `.env.example` does not match what `.env.js` validates

### Additional information

_No response_

c-ehrlich:
Closes #1795 

## ✅ Checklist

💯


cli/src/installers/envVars.ts
```diff
@@ -25,9 +26,15 @@ export const envVariablesInstaller: Installer = ({
...
-  else if (usingDb) envFile = "with-db.js";
+  if (usingDb) {
+    if (usingPlanetScale) {
+      if (usingAuth) envFile = "with-auth-db-planetscale.js";
+      else envFile = "with-db-planetscale.js";
+    } else {
+      if (usingAuth) envFile = "with-auth-db.js";
+      else envFile = "with-db.js";
+    }
+  }
```
c-ehrlich, line 37: dont love this, feel free to suggest something easier to read

cli/template/extras/src/env/with-auth-db-planetscale.js
```diff
@@ -0,0 +1,60 @@
+import { createEnv } from "@t3-oss/env-nextjs";
+import { z } from "zod";
+
+export const env = createEnv({
+  /**
+   * Specify your server-side environment variables schema here. This way you can ensure the app
+   * isn't built with invalid env vars.
+   */
+  server: {
+    DATABASE_URL: z.string().url(),
```
juliusmarminge, line 10: what's your thought on swapping out the url for host, username and password? I tend to prefer that as it's easier to switch between different users but idk if it's just me

juliusmarminge, line 10: this looks like a planetscale file, shouldn't that have the same schema as the `with-db-planetscale`, i.e. the check for default URL?

c-ehrlich: yea

c-ehrlich: just for planetscale or also for other DBs where this can be possible (i think everything except sqlite?)

c-ehrlich: personally if im swapping locally i usually just keep two connection strings in my env file and comment one out. i don't think i would find it more convenient, but not less convenient either. maybe worse for DB providers that only give connection strings so users would have to extract host/user/pw?

dont feel strongly either way, feel free to make the change if you think it's better :)

juliusmarminge: Mainly planetscale maybe? Their docs suggests it: image image

No need for the ssl query param note then too

juliusmarminge (approved):

PR #1893, opened 2024-05-19
feat: add `db:generate` and `db:migrate` scripts

brandon93s:
Closes #<issue>

## ✅ Checklist

---

## Changelog

- Add `db:generate` and `db:migrate` scripts

---

## Screenshots

N/A

💯

### DIFF
--- a/.changeset/hot-chairs-fail.md
+++ b/.changeset/hot-chairs-fail.md
@@ -2,4 +2,4 @@
 "create-t3-app": minor
 ---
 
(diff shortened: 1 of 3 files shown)


juliusmarminge (changes requested):
Add for prisma too

brandon93s:
Prisma added.  They're functionally a bit different in that prisma generate will also run in dev with the `db:generate` command, but I believe this to be the expected default.  Generate only could be opted-in with `npm run db:generate -- --create-only`.

PR #1925, opened 2024-07-02
fix: promise delaying database calls

ashermyers:
Closes #<issue>

## ✅ Checklist

- [ X ] I have followed every step in the [contributing guide](https://github.com/t3-oss/create-t3-app/blob/main/CONTRIBUTING.md) (updated 2022-10-06).
- [ X ] The PR title follows the convention we established [conventional-commit](https://www.conventionalcommits.org/en/v1.0.0/)
- [ X ] I performed a functional test on my final commit

---

## Changelog

Removes purposely slow database queries in the cli generated app.
This functionality is unneeded and unwanted. It almost made me not use this project because of the performance issues of a simple database write.

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/cli/template/extras/src/server/api/routers/post/base.ts
+++ b/cli/template/extras/src/server/api/routers/post/base.ts
@@ -19,9 +19,6 @@ export const postRouter = createTRPCRouter({
   create: publicProcedure
     .input(z.object({ name: z.string().min(1) }))
     .mutation(async ({ input }) => {
-      // simulate a slow db call
-      await new Promise((resolve) => setTimeout(resolve, 1000));
-
       post = { id: post.id + 1, name: input.name };
       return post;
     }),
--- a/cli/template/extras/src/server/api/routers/post/with-auth-drizzle.ts
+++ b/cli/template/extras/src/server/api/routers/post/with-auth-drizzle.ts
@@ -19,9 +19,6 @@ export const postRouter = createTRPCRouter({
   create: protectedProcedure
     .input(z.object({ name: z.string().min(1) }))
     .mutation(async ({ ctx, input }) => {
-      // simulate a slow db call
-      await new Promise((resolve) => setTimeout(resolve, 1000));
-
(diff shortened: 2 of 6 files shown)


juliusmarminge:
You can also have this in a middleware that only runs in dev

https://github.com/juliusmarminge/trellix-trpc/blob/main/src/trpc.ts#L32-L36

ashermyers:
> You can also have this in a middleware that only runs in dev
> 
> https://github.com/juliusmarminge/trellix-trpc/blob/main/src/trpc.ts#L32-L36

Why would anyone purposefully add a delay that’s completely unneeded. This shouldn’t be the behavior in production or development in my opinion.

juliusmarminge:
> Why would anyone purposefully add a delay that’s completely unneeded.

Cause when developing locally you're usually much closer to your database than your users are in prod, with very little network issues. Maybe your db is even local on your machine making the delays close to 0.

By having an artificial delay in dev you are more likely to discover waterfalls leading to bad UX in prod.

> This shouldn’t be the behavior in production

Correct. That should be fixed

> or development in my opinion.

open for a discussion but I do think it provides value. A whole second is too much though 😅

ashermyers:
> > Why would anyone purposefully add a delay that’s completely unneeded.
> 
> Cause when developing locally you're usually much closer to your database than your users are in prod, with very little network issues. Maybe your db is even local on your machine making the delays close to 0.
> 
> By having an artificial delay in dev you are more likely to discover waterfalls leading to bad UX in prod.
> 
> > This shouldn’t be the behavior in production
> 
> Correct. That should be fixed
> 
> > or development in my opinion.
> 
> open for a discussion but I do think it provides value. A whole second is too much though 😅

I definitely agree with one second being to much and I also believe it’s best to be removed entirely. If the developer wants it they can add it themselves but the first impression for a boilerplate or template should not be a purposefully slow experience.

PR #1970, opened 2024-08-22
fix: github stars count

dipanshurdev:
Closes #<issue>

## ✅ Checklist

---

## Changelog

The GitHub star count was not displaying correctly. Previously, the "githubStars" in "githubicon.astro" file variable was not being updated with the fetched star count, causing the display to remain at the fallback value of "23K."

Updated the code to correctly display GitHub star. The "githubStars" variable now correctly reflects the fetched star count instead of using a fixed fallback value, ensuring accurate formatting and display of large numbers.

---

### Before:

Screenshot 2024-08-22 205959

### After:

Screenshot 2024-08-22 210005

💯

### DIFF
--- a/www/src/components/navigation/githubIcon.astro
+++ b/www/src/components/navigation/githubIcon.astro
@@ -5,10 +5,14 @@ import { fetchGithub } from "../../utils/fetchGithub";
 let githubStars = 23000;
 
 try {
-  (await fetchGithub("https://api.github.com/repos/t3-oss/create-t3-app", {
-    throwIfNoAuth: false,
-    fetchType: "repo",
-  }).then((data) => data?.stargazers_count)) ?? 23000;
+  const fetchedStars = await fetchGithub(
+    "https://api.github.com/repos/t3-oss/create-t3-app",
+    {
+      throwIfNoAuth: false,
+      fetchType: "repo",
+    },
+  ).then((data) => data?.stargazers_count);
+  githubStars = fetchedStars ?? 23000;
 } catch (e) {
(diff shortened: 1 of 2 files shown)


www/src/components/navigation/githubIcon.astro
```diff
@@ -5,15 +5,21 @@ import { fetchGithub } from "../../utils/fetchGithub";
 let githubStars = 23000;
```
juliusmarminge, line 5:
```suggestion
let githubStars = 23000;
```

www/src/components/navigation/githubIcon.astro
```diff
@@ -5,15 +5,21 @@ import { fetchGithub } from "../../utils/fetchGithub";
...
-  }).then((data) => data?.stargazers_count)) ?? 23000;
+  const fetchedStars = await fetchGithub(
+    "https://api.github.com/repos/t3-oss/create-t3-app",
+    {
+      throwIfNoAuth: false,
+      fetchType: "repo",
+    },
+  ).then((data) => data?.stargazers_count);
+
+  githubStars = fetchedStars ?? 23000;
```
juliusmarminge, line 16:
```suggestion
  githubStars = fetchedStars ?? 25000;
```

juliusmarminge (commented):
Might as well update the fallback to be more in line with reality...

should really fix the root cause of this, we got rate limited by the github api even if it should only make a request every hour, just... aint got time for it

juliusmarminge (approved):

Issue #1983, opened 2024-09-19
bug: ./startDatabase CLI in says the database is alive when it's not in WSL.


Sapython:
### Provide environment information

It's not but here is the version

`"initVersion": "7.37.0"`

### Describe the bug

If you run the `./startDatabase` command inside WSLand the docker desktop on windows is running, it works fine. But if you pause the docker desktop on Windows side the WSL cli breaks. As the condition to check if docker is present is met, but is docker running is not. This can be checked by running `docker info` which fails to output a valid status runtime code that triggers the condition that the docker desktop is not running but is present in the env `PATH`.

### Reproduction repo

https://github.com/Sapython/interior-design-project

### To reproduce

1. Just create a new project inside WSL
2. Stop docker desktop from running on windows.
3. Then run `start-database.sh` and it will show a success message.

### Additional information

![image](https://github.com/user-attachments/assets/6aa6ea2d-1d87-481c-80d7-850635cc1c0f)

The First line is with Docker desktop running and the second one is without but in both cases it shows that the container is operational.

Sapython:
Pull request for this fix is [https://github.com/t3-oss/create-t3-app/pull/1984](https://github.com/t3-oss/create-t3-app/pull/1984)

juliusmarminge:
Closed by https://github.com/t3-oss/create-t3-app/pull/1975

PR #1984, opened 2024-09-19
fix: Added a condition, if docker is alive or not.  #1983

Sapython:
This caused problems when docker was not running which resulted in the CLI misinforming me that the database was already running. But it wasn't.

Closes #<issue>

## ✅ Checklist

---

## Changelog

Added a simple condition to run `docker info` to check if the docker daemon is actually running or not.
---

## Screenshots

_[Screenshots]_
image

In this image you can see that the both of the execution of `start-database.sh` are successful. But the second one should fail because I killed the docker daemon but the CLI wasn't able to pick it up.
💯

### DIFF
--- a/cli/template/extras/start-database/mysql.sh
+++ b/cli/template/extras/start-database/mysql.sh
@@ -16,6 +16,11 @@ if ! [ -x "$(command -v docker)" ]; then
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


juliusmarminge:
Closed by https://github.com/t3-oss/create-t3-app/pull/1975

PR #2021, opened 2024-11-10
feat: Support biome.js as a linter / formatter option in the cli

### LINKED ISSUE #1973: feat: Add Biome.js Implementation

No but yes. If you want a better developer experience you should use Biome instead of Prettier and ESLint. It is better in performance and is together and compativle

An option in the CLI where you could choose between Eslint and biome and prettier and biome.

Use biome

### Additional information

_No response_

aidansunbury:
Closes #1973 

## ✅ Checklist

## Changelog
The CLI now prompts the user to select either eslint/prettier or biome as a formatting and linting option. If eslint/prettier is selected, the app is scaffolded the same way it always is. If biome is selected, files (prettier.config.js, .eslintrc.cjs) and all eslint/prettier related dependencies are no longer included and instead a biome.jsonc file is generated and @biomejs/biome is installed.

Prettier and eslint are enabled by default through cli flags, but a linter selection can now be specified with either an --eslint or --biome flag.

Additionally, the installation of prettier used to be tied to using tailwind css, meaning selecting *no* for "Will you be using Tailwind CSS for styling?" meant that prettier was not installed. A choice of formatter and a choice of using tailwind css should not be coupled, so now prettier will be installed so long as biome is not selected as a formatter. The 'prettier-plugin-tailwindcss' will only be configured in the prettier config file if tailwind css is selected.

For the biome.jsonc file, I just used the [default configuration](https://biomejs.dev/guides/getting-started/#configuration) and added the [use-sorted-classes](https://biomejs.dev/linter/rules/use-sorted-classes/) nursery rule. I would love to hear thoughts on what should be in the default config, but making changes to the template file will not impact the changes to the functionality of this feature.

## Screenshots
Screenshot 2024-11-09 at 5 37 16 PM

💯


Ikkevoid:
I checked it out locally, and from what I can tell, things are working. However, the format:check and format:write commands aren't working for me. Unfortunately, there's no biome format --check, and it's also impossible to specify file extensions to go through, as you can with Prettier.

Another thing to consider: do you plan to do anything with the next lint commands inside the package.json file?

aidansunbury:
> I checked it out locally, and from what I can tell, things are working. However, the format:check and format:write commands aren't working for me. Unfortunately, there's no biome format --check, and it's also impossible to specify file extensions to go through, as you can with Prettier.
> 
> Another thing to consider: do you plan to do anything with the next lint commands inside the package.json file?

Thanks for catching that! There are now three scripts added when you use biome. Check linting and formatting without applying fixes, apply only safe fixes, and apply unsafe fixes. They should all be working now, and biome only checks js/ts files anyways, so it is not needed to specify the file types.

Screenshot 2024-11-14 at 9 18 12 PM

I removed the `next lint` scripts when using biome, as the whole points of using biome is to then stop using eslint. 

Also, as a minor refactoring note, I created an `addPackageScript` helper and refactored a few of the installers to use it.

aidansunbury:
@TheCukitoDev Biome has three main commands: format, lint, and check (which runs the formatter and linter). I personally pretty much always just use the check command, but would you from a dx perspective also prefer to have lint and format commands already included in the package.json?

TheCukitoDev:
Hello, i created other PR Request because i dont know how could I add code to this one. I added another option with your Biome implementation and a mixed one that uses ESLint and Biome as I like to use. The PR is the #2025

juliusmarminge (changes requested):
This almost looks mergeable, but I think we might want to run `biome check --write --unsafe` in the CLI before staging the files, because currently the repo is not in a "tidy" state after init

we also need CI that prevents future regressinos, ensuring the generated app is always formatted, passes linting etc

aidansunbury:
@juliusmarminge I added a `format.ts` file which runs either `format:write` or `check:unsafe` once the rest of the project is scaffolded. 

As far as adding the checks to CI, we are running into the issue that a [matrix can only generate a max of 256 variations](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow), which adding the additional variable of selecting a formatter puts us over. You can see this limit being reached in 08173ced6eb8b8a1fa51e791be4cec17bcc9093b.

In the long run, we need to come up with a new testing strategy which will keep us under the 256 maximum while still allowing more options to be added to the cli and tested. Once this new strategy is implemented, we can add the formatter checks.

cli/src/index.ts
```diff
@@ -89,6 +90,8 @@ const main = async () => {
     await initializeGit(projectDir);
   }
 
+  await formatProject({ pkgManager, projectDir, eslint, biome });
```
juliusmarminge, line 93: This should run before initing git so that the all files are staged

cli/template/extras/config/biome.jsonc
```diff
@@ -0,0 +1,25 @@
+{
+  "$schema": "https://biomejs.dev/schemas/1.9.4/schema.json",
+  "vcs": {
+    "enabled": false,
```
juliusmarminge, line 4: Don't you want these on?

cli/template/extras/config/biome.jsonc
```diff
@@ -0,0 +1,25 @@
...
+    "clientKind": "git",
+    "useIgnoreFile": false
+  },
+  "files": { "ignoreUnknown": false, "ignore": [] },
+  "formatter": { "enabled": true, "indentStyle": "space", "indentWidth": 2 },
```
juliusmarminge, line 9: Let's maybe just use defaults - no need to have opinions here ?

cli/template/extras/config/biome.jsonc
```diff
@@ -0,0 +1,25 @@
...
+  "organizeImports": { "enabled": true },
+  "linter": {
+    "enabled": true,
+    "rules": { 
+      "nursery": {
+        "useSortedClasses": {
```
juliusmarminge, line 15: 👍

cli/template/extras/config/biome.jsonc
```diff
@@ -0,0 +1,25 @@
...
+          "level": "error",
+          "options": {
+            "functions": ["clsx", "cva", "cn"]
+          }
+        }
+      },  
+    "recommended": true }
+  },
+  "javascript": { "formatter": { "quoteStyle": "double" } }
```
juliusmarminge, line 24: While I do agree with your choice, again let's just use defaults

juliusmarminge (changes requested):

cli/src/helpers/format.ts
```diff
@@ -0,0 +1,33 @@
...
+}) => {
+  logger.info("Formatting project...");
+  const spinner = ora("Running format command\n").start();
+
+  if (eslint) {
+    await execa(pkgManager, ["format:write"], {
+      cwd: projectDir,
+    });
+  } else if (biome) {
+    await execa(pkgManager, ["check:unsafe"], {
```
juliusmarminge, line 28: CleanShot 2025-01-25 at 15 33 33@2x

hmm still getting a project that doesn't pass lint

juliusmarminge (changes requested):

cli/src/helpers/format.ts
juliusmarminge: i think even if we can't have a matrix that includes biome we should atleast have 1 app scaffold (perhaps the variant with as much selected as possible? biome+auth+trpc+tailwind) to try and surface this as best as possible

aidansunbury:
The issue was that I was only checking CI flags to see what formatting should run, and not the packages itself. This caused the formatting to work when using the flags but not the interactive interface.

In addition to fixing that, I added a `check-well-formatted` test which builds the app with biome and runs `biome check` and builds the app with eslint and checks the linting and formatting.

aidansunbury:
Also I just modified the bun test to check for a `bun.lock` instead of `bun.lockb` since the [latest version of bun](https://bun.sh/blog/bun-v1.2#introducing-bun-lock) now defaults to the text based `bun.lock`. I think everything should be good to go now.

aidansunbury:
@juliusmarminge Anything else before this can get merged?

aidansunbury:
> @juliusmarminge Anything else before this can get merged?

Bumping this

juliusmarminge (approved):

Issue #1927, opened 2024-07-02
feat: Use docker compose instead of a single docker container


TheDevilOnLine:
### Is your feature request related to a problem? Please describe.

A lot of our developers work on Windows and running `start-database.sh` requires a bunch of extra steps

### Describe the solution you'd like to see

`start-database.sh` suggest running itself in WSL, but I think it would be way easier to supply a `docker-compose.yml`. This also allows for random password generation to be used in the `docker-compose.yml` and in the `.env` file.

### Describe alternate solutions

N/A

### Additional information

If people are interested in this solution, I can make a pull request with the suggested functionality.

datasalaryman:
Hi, to clarify, do you see the server and client parts of the nextjs to be separate docker containers? or something else?

TheDevilOnLine:
@datasalaryman just the database (so the stuff that's currently being started into docker when you run the start-database.sh command)

datasalaryman:
@TheDevilOnLine so the docker-compose file will be structured like this?

```
services:
  web:
    ...
  postgres:
    ...
```

okay, I'll see if I can take that on

PR #2049, opened 2025-01-25
feat: Add docker compose

NathanaelRea:
Closes #<issue>

## ✅ Checklist

---

## Changelog

Add `docker-compose.yml` file for postgres and mysql database options. `start-database.sh` now exports the needed environment variables and calls `docker compose up -d`. This serves as a more observable example, and is more extensible if the user wants to add services.

Moves forward #1927 because the user can now run `docker compose up -d`, but not complete because would probably want to:

1. Unify `./start-databse.sh` into an entrypoint for a `Dockerfile` (one line difference - setting container name)
2. Move `generate db password` into cli after choosing db provider to simplify `start-database.yml`

---

## Screenshots

_[Screenshots]_

💯

### DIFF
--- a/.changeset/fair-ducks-grow.md
+++ b/.changeset/fair-ducks-grow.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+create docker-compose.yml for postgres/mysql database options


NathanaelRea:
Also I noticed that currently the mysql `start-database.sh` file uses `postgres` for the container name. Should I fix it here or create a separate PR?

PR #2169, opened 2025-12-03
fix(ui): update default value for GitHub stars count component

Mist-wu:
## Changelog

Updated the default value for the GitHub stars count component from `25000` to `28000` . 

---

## Screenshots
before:
before

after:
after


juliusmarminge (approved):
maybe one day we'll figure out why astro isn't caching this fetch correctly so we can show the actual number without hitting rate limits ☺️

for now, thank you!

PR #2218, opened 2026-03-15
fix(cli): handle boolean importAlias when --import-alias has no value

### LINKED ISSUE #2215: bug: --import-alias flag crashes with TypeError when used in CI mode

  System:
    OS: macOS 26.2
    CPU: Unknown
    Memory: 122.91 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.4.1 - /opt/homebrew/bin/node
    npm: 11.4.2 - /opt/homebrew/bin/npm
    pnpm: 10.8.0 - /opt/homebrew/bin/pnpm

### Describe the bug

Using the `--import-alias` (or `-i`) flag in CI mode causes a `TypeError: i.replace is not a function` crash after the boilerplate is successfully installed.

The issue is in `cli/src/helpers/setImportAlias.ts` (line 24-25), where `.replace()` is called on the `importAlias` value. Commander parses the `--import-alias` option with an optional value (`[alias]`), which can result in the value being `true` (boolean) instead of a string, causing the crash.

Expected behavior: passing `--import-alias "@/"` should set the import alias to `@/` without errors.

### Reproduction repo

No repo needed — this is reproducible with any fresh scaffold using the published CLI.

### To reproduce

1. Run the following command:

```bash
pnpm dlx create-t3-app@latest my-app --CI --tailwind --appRouter --eslint --import-alias "@/"
```

2. The CLI scaffolds successfully and installs boilerplate, but then crashes:

```
Using: pnpm

✔ my-app scaffolded successfully!

Adding boilerplate...
✔ Successfully setup boilerplate for tailwind
✔ Successfully setup boilerplate for envVariables
✔ Successfully setup boilerplate for eslint

Aborting installation...
TypeError: i.replace is not a function
```

3. The same happens with the short flag `-i "@/"`.

4. Without `--import-alias`, the command works fine:

```bash
pnpm dlx create-t3-app@latest my-app --CI --tailwind --appRouter --eslint
```

### Additional information

[this is AI]

The root cause is in `cli/src/cli/index.ts` line 147, where the Commander option is defined as:

```typescript
.option(
  "-i, --import-alias [alias]",
  "Explicitly tell the CLI to use a custom import alias",
  defaultOptions.flags.importAlias
)
```

The `[alias]` (optional value) syntax allows Commander to return `true` instead of the actual string value in some cases. Then in `cli/src/helpers/setImportAlias.ts`:

```typescript
const normalizedImportAlias = importAlias
    .replace(/\*/g, "")
    .replace(/[^/]$/, "$&/");
```

`true.replace()` throws `TypeError`.

Possible fixes:
- Change `[alias]` to `<alias>` (required value) in the Commander option definition
- Add defensive type checking before calling `.replace()`

mahmoodhamdi:
Fixes #2215

When `--import-alias` (or `-i`) is used in CI mode without providing a value, Commander parses the option as `true` (boolean) instead of a string. This causes a `TypeError: i.replace is not a function` crash in `setImportAlias.ts` because `.replace()` is called on a boolean.

The fix adds a type check after parsing CLI options - if `importAlias` is not a string, it falls back to the default value (`"~/"`).

**Reproduction:**
```bash
npx create-t3-app@latest my-app --CI --import-alias
# TypeError: i.replace is not a function
```

**After fix:**
```bash
npx create-t3-app@latest my-app --CI --import-alias
# Uses default "~/" alias
```

### DIFF
--- a/.changeset/fix-import-alias-crash.md
+++ b/.changeset/fix-import-alias-crash.md
@@ -1,0 +1,5 @@
+---
+"create-t3-app": patch
+---
+
+fix(cli): handle boolean importAlias when --import-alias is used without a value in CI mode
--- a/cli/src/cli/index.ts
+++ b/cli/src/cli/index.ts
@@ -200,6 +200,12 @@ export const runCli = async (): Promise<CliResults> => {
 
   cliResults.flags = program.opts();
 
+  // When --import-alias is passed without a value, Commander sets it to true (boolean).
+  // Fall back to the default alias in that case.
+  if (typeof cliResults.flags.importAlias !== "string") {
+    cliResults.flags.importAlias = defaultOptions.flags.importAlias;
+  }
+
   /** @internal Used for CI E2E tests. */
   if (cliResults.flags.CI) {
     cliResults.packages = [];


mahmoodhamdi:
Closing — apologies for the noise.

